PR #58, opened 2016-06-26
[WHITELIST] images.twee-app.com

Tazer:
We are having problems when proxying/serving images. Cause thetvdb isn't
allowing hotlinking , so we need to have the images on our side. So
cause the tvdb using /banners , we get caught. So I have whitelisted our
domain exactly like thetvdb.com domain.

Here is the logs also
log

if you want to visit the site it's https://web.twee-app.com , so you can verify it's not ads. An example image under the banners are: https://images.twee-app.com/banners/posters/80224-1.jpg


Tazer:
I did rename our folder. so guess this isn't needed.

PR #115, opened 2016-08-31
Neon lab fix

mdesnoyer:
Make the filters for Neon Labs accurate.

We are an image optimization company. For some of our customers, we do A/B testing on imagery and collect telemetry on how well images are doing. It makes sense for our tracking domain to be on the privacy list so that people can opt out, however, the existing filter is too broad and catches our entire domain (including corp site www.neon-lab.com, our app app.neon-lab.com etc).

These changes make sure that our tracking endpoints can be blocked, but that the rest of our business is not.

Mark Desnoyer,
VP Engineering, Neon Labs Inc.


ryanbr:
https://github.com/easylist/easylist/commit/e6f7476ee82713bc3e4d4f422918aea2ea98b1b7

PR #119, opened 2016-09-03
Add Google Analytics by MonsterInsights to whitelist

chriscct7:
We've got a lot of uBlock users who when they try to view our plugin, the block of anything with `google-analytics` in the slug breaks the backend WordPress dashboard for the whole plugin


gorhill:
You have an example URL where this occurs?

gorhill:
If I understand correctly, the resource `/wp-content/plugins/google-analytics-for-wordpress` is pulled only by the Wordpress dashboard, never when a user browses a site using your plugin?

chriscct7:
Hi there,
You would need to install the plugin to see the issue. The issue is our
plugin has a backend settings panel area, that's only visible to the site
owner, but right now, the asset files we use to make and style that panel
are blocked.

ryanbr:
Without being tested this wouldn't whitelisted.

PR #147, opened 2016-09-18
https://www.angieslist.com

cmorse:
Stops annoying popup that comes up on every view of website


monzta:
Could you post a screenshot of that overlay? I'm only getting the zip code one.

cmorse:
angies list popup

cmorse:
It doesn't seem to happen on every page on the site. But on the pages where it does happen, it pops up on every view.

PR #165, opened 2016-10-04
Removed adf.ly from third-party popups. This rule is NOT blocking the ad

nigeon:
Removed adf.ly from third-party popups. 
This rule is NOT blocking the ad. Instead, this is blocking the user to reach destination URL as it's opened in a new window.

### DIFF
--- a/easylist/easylist_specific_block_popup.txt
+++ b/easylist/easylist_specific_block_popup.txt
@@ -1,7 +1,7 @@
 .link/$popup,domain=bigfile.to
 /sendspace-pop.$popup,domain=sendspace.com
 ^utm_source=$popup,domain=exashare.com|sex.com|thepiratebay.am|thepiratebay.gd|thepiratebay.la|thepiratebay.mn|thepiratebay.vg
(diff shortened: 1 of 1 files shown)


Khrin:
It's working correctly for me on Firefox.

nigeon:
No, is not working **properly**. And I mean, it's blocking the ads, but it's also breaking the website, as I can't reach the destination URL. I'll try to explain it in detail.

I've created an example link http://adf.ly/1eZhRZ that points to github.com homepage. 
In case you are not familiar with adf.ly, it's a url shortener and the destination should reached after clicking "Skip Ad" button.

The problem of the current rules is that is making adf.ly links impossible to use.
Following examples happen every try

This describes what currently happens In Firefox with ABP: 
1. Adfly loads with a disclaimer message, when you click "Continue" a popup opens and is automatically closed. Good.
2. Click on "Continue" again, the disclaimer disappears, the countdown restarts.
3. The "Skip Ad" button appears, you click on it and the destination opens in a new tab but is automatically closed! Can't reach the final url.

If we look into the HTML of the page, we can see how "Skip ad" button, is a link with target="_blank" with the destination url as href.

It's true that with the current rules, no ad is appearing, but you can't either reach the destination URL :S.

In Chrome with ABP is even worse:
1. Adfly loads with a disclaimer message, when you click "Continue" a popup opens and is **not closed**. Ok.. manually close...
2. Click on "Continue" again, the disclaimer disappears, the countdown restarts.
3. The "Skip Ad" button appears, you click on it and the destination opens in a new tab but is automatically closed! Can't reach the final url.
4. Active window becomes a full page advert.


Thanks.

PR #182, opened 2016-10-26
Remove iframe block on diablofans.com, which intereferes with comment…

cpaul747:
… and forum post editing


Khrin:
I guess this has been fixed with https://github.com/easylist/easylist/commit/cb682bedaffa778318569967b81bd92b5c585a86

PR #246, opened 2017-01-17
Replace https://siftscience.com with https://hexagon-analytics.com

KaoruDev:
I work at Sift Science and we provide fraud prevention services for online businesses. Most of our clients use a JavaScript snippet that is hosted on our domain to more accurately predict fraud on their sites.

We have noticed that our domain [siftscience.com](https://siftscience.com) is on the blocked list, which breaks our website. Please unblock this domain, as we no longer send user information to this domain. We have changed the endpoint to be hosted on [hexagon-analytics.com](http://hexagon-analytics.com) so that we may get our main domain unblocked.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -806,6 +806,7 @@
 ||heatmap.it^$third-party
 ||hellosherpa.com^$third-party
 ||hentaicounter.com^$third-party
+||hexagon-analytics.com^$third-party
 ||heystaks.com^$third-party
 ||hiconversion.com^$third-party
 ||higherengine.com^$third-party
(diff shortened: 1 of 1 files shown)


gotitbro:
@monzta I am seeing third party requests and resources being loaded from _siftscience.com_ even now. Such as here: `https://www.truthfinder.com/`

A JS is loaded from siftscience.com: `https://cdn.siftscience.com/s.js`

KaoruDev:
@gotitbro sorry could clarify a bit more? 

> I am seeing third party requests and resources being loaded from siftscience.com even now. Such as here: https://www.truthfinder.com/

Are you saying that `www.truthfinder.com` is loading resources from siftscience.com? Is it `s.js` file you mentioned previously?

In the original comment I discussed how some of our customers will still download a javascript file from our main domain but that the javascript itself won't send data to our main domain, instead it sends the information to hexagon-analytics.com. Hopefully this clears things up a bit.

gotitbro:
@kaorudev If a JS from your domain triggers further analytical requests its just better to block it than the further connections made. Also your domain was blocked by `siftscience.com^$third-party` which didn't affect the main website. So I see no point in having it removed.

PR #252, opened 2017-01-23
Additions to easylist_adservers.txt

diaux:
Both zerozo.work and rolinda.work redirect to publited.com, which shows to be ad network business.

If you go to http://www.mejortorrent.com/ and click on any movie, for example see this one http://www.mejortorrent.com/peli-descargar-torrent-14552-Stereo.html , then click "Descargar" (spanish for "Download") and then finally click where it says "Pincha aquí para descargar el torrent." (spanish for "Click here to download torrent."). Popups will appear. I got one from URL rolinda.work. If you go to rolinda.work, it redirects to publited.com.

The other URL in the commit is zerozo.work, but I was not able to reproduce this one (this was originally reported in EasyList Spanish Forum a couple months ago, that's why it may have changed). But if you go to zerozo.work, it redirects to publited.com.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -950,6 +950,7 @@
 ||betpartners.it^$third-party
 ||betrad.com^$third-party
 ||bettingpartners.com^$third-party
+||bezoya.work^$third-party
 ||bf-ad.net^$third-party
 ||bfast.com^$third-party
 ||bh3.net^$third-party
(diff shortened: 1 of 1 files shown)


diaux:
Added another commit, for divxatope1.com. Another domain that redirects to publited.com: bezoya.work. Unfortunately, I was not able to reproduce the issue on the reported website, but I am including it into this Pull Request either way.

PR #253, opened 2017-01-23
A: http://www.forospyware.com/

diaux:
There's an ad in http://www.forospyware.com/, top-right of the page. If you Inspect Element in Firefox, you will see img src="https://max-infospyware.netdna-ssl.com/images/adsMB/Spanish_MB_468X60v2.jpg"

"adsMB" apparently refers to "ads **M**alware**b**ytes"

### DIFF
--- a/easylist/easylist_general_block.txt
+++ b/easylist/easylist_general_block.txt
@@ -4381,6 +4381,7 @@
 /images/ads.
 /images/ads/*
 /images/ads_
+/images/adsMB/*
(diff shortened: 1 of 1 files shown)


Khrin:
I think that this can be refined even more: https://github.com/easylist/easylist/commit/014463b62653682562f41d7cd2587273157ea229

diaux:
Yes, your commit makes more sense than mine. Thank you!

PR #255, opened 2017-01-23
Addition to easylist_general_hide.txt

diaux:
Found div class=rightAds in http://www.tandildiario.com/ that suits EasyList. Adding general rule to hide this class.


Khrin:
There's much more that can be added: https://github.com/easylist/easylist/commit/b0b723812b6786e506ff1f27e51c5de3f30b4486 Thank you.

diaux:
Good catch! Thank you very much! Your commit looks way better than mine yep!

PR #257, opened 2017-01-24
Add Google Analytics by MonsterInsights to whitelist

chriscct7:
Google Analytics For MonsterInsights is the most popular Google Analytics integration for WordPress, allowing website owners to view their Google Analytics statistics without having to log into Google Analytics (backend dashboard)

Right now, EasyList's EasyPrivacy is blocking the CSS, JS, and images we load to make our WordPress backend panel when uBlock Origin (which uses this whitelist) is used by our users. It's basically the same issue as the Google Analytics Dashboard for WP plugin had which was whitelisted into EasyList EasyPrivacy.

Without uBlock:
http://screencloud.net/v/90o8

With uBlock Origin (which includes EasyList EasyPrivacy on by default):
http://screencloud.net/v/oe30 (it blocks all JS/CSS/Images)

Disabling the EasyPrivacy filterset in uBlock solves the issue.

It also breaks images in unrelated sites, for example, sites like WordPress.org where it blocks our logo, banner, and screenshots: https://wordpress.org/plugins-wp/google-analytics-for-wordpress/
as well as on any site that takes screenshots of our plugin and names the images using our slug, such as on plugin review websites.

This commit adds whitelist for our plugin's asset files based on the way the whitelist rules the other WP plugins are whitelisted in the filterset.

### DIFF
--- a/easyprivacy/easyprivacy_whitelist.txt
+++ b/easyprivacy/easyprivacy_whitelist.txt
@@ -12,6 +12,8 @@
 @@/cdn-cgi/pe/bag2?*skimlinks.js$domain=droid-life.com
 @@/cdn-cgi/pe/bag2?*static.getclicky.com%2Fjs$domain=amazingpics.net
 @@/cdn-cgi/pe/bag2?*yieldbot.intent.js$domain=droid-life.com
+@@/wp-content/mu-plugins/google-analytics-for-wordpress/*
+@@/wp-content/plugins/google-analytics-for-wordpress/*
 @@/wp-content/plugins/google-analytics-dashboard-for-wp/*$~third-party
 @@||1dmp.io/pixel.gif$image,domain=overclockers.co.uk
 @@||247sports.com/site/minify.js?*/scripts/stats/$script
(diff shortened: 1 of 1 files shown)


chriscct7:
TravisCI fail is unrelated to PR. Seems to be an authentication error in the build script being used `fatal: Authentication failed for 'https://@github.com/easylist/easylist/'`

Khrin:
This should fix all the necessary: https://github.com/easylist/easylist/commit/8363bf4b92e550a12c96fb66517e5c0520e24326

chriscct7:
@Khrin we have a second slug that's `google-analytics-premium` (so we would need those 3 rules also). Additionally unless it's whitelisted somewhere I didn't see `ps.w.org/google-analytics-analytics-for-wordpress/assets/` would also need to be whitelisted as WP.org uses that url for the current plugin directory. Did you want me to open a new pr for those?

Khrin:
We already have `@@||ps.w.org^*/assets/$image,domain=wordpress.org`.
Please tell me if this fixes all the remaining: https://github.com/easylist/easylist/commit/5345495298fd741333365d09ed13eb90dd7a5aa6

chriscct7:
Yep that works!

PR #262, opened 2017-01-30
Additions to easylist_adservers.txt

diaux:
This should be continuation of #252. In the past few days, I have been encountering several ads coming from several domains ending in .work, which was suspicious to me. Today, I looked at URL www.mejortorrent.com and I found another one (fluxybe.work). I decided I would do a whois on that one, and so I found Registrant Name, then I looked at a reverse whois for that Registrant Name (see this link: http://www.viewdns.info/reversewhois/?q=AFOVEL+SA). I then checked all domains in the list, and found several domains (most of them ending in .work) that are not included into easylist_adservers.txt. All of them redirect to publited.com. Using mejortorrent.com as a example, they have used zerozo.work in the past to launch ads (https://forums.lanik.us/viewtopic.php?f=103&t=27560), then rolinda.work (see #252), and now today they are using fluxybe.work. Looks like a pool of domains, and they rotate between them. Other websites probably do the same. I believe all of these should be included into EasyList. Thanks.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -719,6 +719,7 @@
 ||affplanet.com^$third-party
 ||afftrack.com^$third-party
 ||aflrm.com^$third-party
+||afovelsa.com^$third-party
 ||africawin.com^$third-party
 ||afterdownload.com^$third-party
 ||afterdownloads.com^$third-party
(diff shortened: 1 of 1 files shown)


Khrin:
Done.

Issue #267, opened 2017-02-02
WorldSSL.net - Not an advertiser


igobyterry:
Hello,

WorldSSL.net is not an advertising company. It is the domain name used for the OnApp CDN federation. This has broken the CDN service for thousands of websites. You can see this in the help docs at https://help.cdn.net/hc/en-us/articles/213170485-Custom-SSL-Setup

biomancer:
Confirming, this domain is used by CDN providers and is not an advertiser, please accept the PR asap, multiple sites cannot load their assets currently.

biomancer:
Source commit with addition is https://github.com/easylist/easylist/commit/d7626c0c16be9ceecb461f , @Khrin, take a look please

Khrin:
I apologize for the misinterpretation, now it's removed. https://github.com/easylist/easylist/commit/74e0401057cbfb2f99bb538a076c7a580659364e

PR #268, opened 2017-02-02
Remove worldssl.com from easylist_adservers.txt

IlyaPomaskin:
https://github.com/easylist/easylist/issues/267


mapx-:
see also
https://forums.lanik.us/viewtopic.php?f=64&t=35293

PR #274, opened 2017-02-06
Addition to easylist_whitelist.txt

diaux:
Anti adb filter for veja.abril.com.br. This one came from an Issue Report to EasyList Spanish, but the website is in portuguese, and not in spanish. Therefore this filter belongs to EasyList and not EasyList Spanish.


Khrin:
I've just put it in a different section ("Non-english") where we already have a bunch of Brazilian filters. Thank you!

diaux:
OK! In the future, I will put filters for non English websites into that
section when doing the commits, so you don't have to do it yourself. Thanks!

EasyList Spanish

PR #291, opened 2017-02-21
No third-party cookie on a.mobify.com

sketchytechky:
Hi there,

(following on a thread of email with @monzta )

The promises of Mobify is to use our javascript to overlay the Mobile website.  It is party of the functionality of the website (which we require first party cookie) - but we do not use 3rd party cookie to track users.

Blocking the script breaks the customer experience on Mobile devices.  Would like to remove ourselves from the list and to be whitelisted. 

Example site: extrastores.com

List of cookies
image

Many thanks!

Our legal privacy policy: http://downloads.mobify.com/legal/mobify-privacy.pdf


Khrin:
There's no need to double post and moreover there's no need to also whitelist your domain.

PR #307, opened 2017-03-11
Block popup on Openload's alternative domain

darkmorpher:
This has better effect however: `about:blank$popup,domain=oload.tv`


darkmorpher:
This has better effect however: `about:blank$popup,domain=oload.tv`

PR #362, opened 2017-04-09
allows Disqus loading

ptcong:


monzta:
Already fixed in EasyList.

PR #380, opened 2017-04-18
Addition to easylist_whitelist_international.txt

juliosgarbi:
Create exception to load googletagmanager script in www.saraiva.com.br

### DIFF
--- a/easyprivacy/easyprivacy_whitelist_international.txt
+++ b/easyprivacy/easyprivacy_whitelist_international.txt
@@ -242,7 +242,7 @@
 @@||nostv.pt^*/angulartics-google-analytics.js?
 @@||siteapps.com^$script,domain=netcombo.com.br
 @@||stats.g.doubleclick.net/dc.js$domain=netcombo.com.br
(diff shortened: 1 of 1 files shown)


monzta:
Why is it necessary?

juliosgarbi:
Saraiva uses googletagmanager to inject the script of our product into your site. When we access Saraiva's website through Firefox with the ublock extension installed googletagmanager is blocked, thus blocking the injection of our script as well.

monzta:
Could you post the URL of the script of your product please?

juliosgarbi:
If you access this url http://saraiva.com.br/travessuras-da-menina-ma-1464513.html you can see our product here: 

our-product

Our product is a social module within Saraiva ecommerce.

The script that is called by GTM is the https://saraiva.more.social/expressions/dist/install.js but as Saraiva uses GTM to manage our script it stops being called when we go through firefox with ublock installed

monzta:
Thanks! But then I would recommend using `saraiva.com.br` instead of `www.saraiva.com.br` in the filter.

juliosgarbi:
Done!

PR #384, opened 2017-04-19
Remove 3 rules for marca.com (website in Spanish already covered in EasyList Spanish)

diaux:
Removed the following rules for marca.com (website in Spanish already covered in EasyList Spanish)

marca.com###publi_sup -> obsolete, no longer needed + website in Spanish either way
@@||marca.com/deporte/css/*/publicidad.css -> obsolete, no longer needed + website in Spanish either way
marca.com#@#.content-ad -> no longer needed, made obsolete by @@||marca.com^generichide in EasyList Spanish


Khrin:
marca.com is indeed a website mostly in Spanish language, but it's also present a sizeable section with content in English, which is why a few filters are present in EasyList. I can't verify right now, but possibly the generichide filter may be required also for the English pages.

diaux:
Yes, you are correct. I did not notice website was also available in English. I have included another commit, with the Anti-Adblock filters for this website, that are also available in EasyList Spanish. See below.

@@||marca.com^$generichide
@@||el-mundo.net/js/advertisement.js$script,third-party,domain=marca.com
||el-mundo.net/js/fm*.js$script

I believe it should be safe to remove the 3 original filters, after this new commit.

PR #392, opened 2017-05-04
Remove non-intrusive wix ads from blacklist

dimarwix:


monzta:
What exactly is the reasoning behind this?

dimarwix:
Hi there @monzta , apologies, should have provided the reason in the PR.
In fact, there are two reasons:

* First and foremost:
**Once the listed urls are blocked, sites may entirely fail to load.**
* Secondly:
They are non-intrusive, and only appear on websites created on the Wix platform for free, or by users that explicitly allow these ads on their site, without any revenue. They are non-intrusive, aren't based on tracking cookies and aren't generating profit to any advertisement agency.

These are intended to expose the Wix brand to users that may potentially want to promote their online presence with Wix.

Thank you for your attention,
Dima

dimarwix:
A request has also been [posted on the forum](https://forums.lanik.us/viewtopic.php?f=64&t=36643&p=117771&hilit=wixads.min.js&sid=5c4b18f2b62c9700ae83dd71d8e0482e#p117771), by our own @technotronic12 

Thanks

monzta:
Could you post some sites where these issues occur?

dimarwix:
We obviously created a fix as soon as we became aware of such a critical issue. 
However, you can still reproduce it using the older version of the code here:
http://dimarysk.wixsite.com/ecom2?ReactSource=1.2224.9

We would like to eliminate the possibility of this happening again.

monzta:
> We would like to eliminate the possibility of this happening again. 

I'm afraind we cannot guarantee this since there could always be some generic blocking rule blocking such a resource in the future. 

Closing this since the initial issue doesn't exist anymore.

PR #396, opened 2017-05-07
Update easylist_specific_block.txt - imgking.co

zero77:


zero77:
What do you mean by Example URL.

monzta:
The exact URL of the website you want to add this filter for.

PR #398, opened 2017-05-07
Update easylist_specific_block.txt - imgprime.com

zero77:


ryanbr:
Closing, the script doesn't do alot anyways, and we mitigate most of the popups/ads on their site.

PR #399, opened 2017-05-07
Update easylist_specific_hide.txt

zero77:


zero77:
Example URL: porn.com and any sub pages.

zero77:
moved to #410

PR #400, opened 2017-05-07
Update easylist_specific_block.txt

zero77:


zero77:
This filter is not needed because, pull request #399 covers it.

PR #401, opened 2017-05-08
Update easylist_whitelist_popup.txt

Narcwis:
added my websites to the popup


monzta:
Why exactly is it necessary?

Narcwis:
I am serving the documents to my clients via blob URLs so that they can be
displayed using the built in PDF viewer (Firefox, Opera and Chrome since
IE11 or Edge don't support this feature yet) obviously in order to have
access to any document you need to be a client of our law firm (Harding
Evans) and you will need your Microsoft account to be enabled on our
service.

If you could add our website to the list it would ease our work since we
would be able to remove the annoying  Adblock alerts currently in place.

I understand the security concerns of it and if you for decide that without
a test site cannot implement this change in the list all I can say is that
is completely reasonable but I will definitely be in contact in the future
when I will be able to provide you with said test site.

PR #419, opened 2017-06-01
hide youtube video cards

verhovsky:
video thumbnails overlaid on the video before you're done watching


ryanbr:
people will complain either way if added or not, lets leave it.

PR #427, opened 2017-06-08
Addition to easylist_adservers.txt

diaux:
**graffica.info**
||axiomaticalley.com^$third-party

Blocks Admiral anti adb script, but it really looks like an adserver to me, given that navigating to that URL displays nothing, and also given the weird/funny hostname, similar to others. I'd like to add this filter into EasyList if you agree. Thanks.


ryanbr:
Blocking Admiral will cause issues on other sites.

Issue #465, opened 2017-07-07
Rule blocking Websocket connections


dekdevy:
All of sudden my web app client side javascript socket.io connection can't be established to the server.
I checked as to why this would be happening, and this filter:

`
/\:\/\/([0-9]{1,3}\.){3}[0-9]{1,3}/$third-party,xmlhttprequest
`

Is blocking this XHR connection:
`
http://xx.xxx.xx.xx/socket.io/?EIO=3&transport=polling&t=LUTudJd
`
Redacted the ip by xxx.xxx.xxx.xx, and it is the reason why things are blocked.
Its been working fine over the last 6 month and suddenly apparently you can not make any connections to a raw ip part? The ip part is what is triggering the filter. Any help / solution?

Issue #464, opened 2017-07-07
This rule blocking too many connections


shuax:
https://github.com/easylist/easylist/commit/4264e8434a2d17d5fdd5c5790368a3491843cbbc

7affer:
Agreed. It's simply blocking all ajax calls by IP address.

PR #473, opened 2017-07-09
Update easylist general_block.txt - PR for issue #464

7affer:
PR for issue #464 and #465 
Rule is blocking all requests by all possible IP address. It's too wide and it makes work with dev servers impossible.


Khrin:
I don't see the point in this change, if you look more carefully, you can see that the filter is already disabled.

7affer:
My bad :) I had issue with that 2 days ago but i didn't noticed it got fixed.

PR #478, opened 2017-07-12
Fix Content & Layout Issues for NBCUniversal Sites

richrhee:
This change removes the rule added (commit 679058f) for csnbayarea.com that is blocking services that deliver page layout and content across all NBCUniversal websites.  It causes issues with content not appearing correctly and broken page functionality.  With this rule removed, I have confirmed that ads on csnbayarea.com still get blocked, as it is already covered under the blocking rules for Google (Doubleclick) ads.


ryanbr:
Was removed a while ago.

Issue #463, opened 2017-07-07
Block Sovrn Anti-Adblock Scripts  [Pull Request #483]


PowerAccess:
Please add the following filters for Sovrn's anti-adblock scripts & popups:

They seem fairly popular commercial anti-adblock solutions and those are just the script hosting domains (as opposed to the sovrn.com corporate site), so I would suggest adding without a domain anchor:

```
||lijit.com^$third-party
||sovrnlabs.net^$third-party

###sovrn_container
###sovrn_beacon
```

You can see this anti-adblock script in action at:
http://westeros.org

xeonx1:
Please approve my Pull Request #483 to add block and hide rules for Sovrn's commercial anti-adblock and ad scripts, with commits adding what I had described above.

PowerAccess:
Can Pull Request #483 be approved? 
It's for a commercial anti-adblock script that prevents adblock usage on many sites.

PR #483, opened 2017-07-18
A: http://westeros.org/ (for #463 - Sovrn Anti-Adblock scripts)

PowerAccess:
Added Sovrn Anti-Adblock scripts ad-server block and general element hiding rules, as I had reported in issue #463 
Sovrn Sovrn.com provides commercial script Anti-adblock and ad scripts used on various sites, which they host at sovrnlabs.net and lijit.com


PowerAccess:
Can this Pull Request be approved?

ryanbr:
Won't be added to Easylist, subscribe to Enhanced trackers list instead.

PR #491, opened 2017-07-29
P: http://item.gmarket.co.kr

devunt:
Fix `gmarket.co.kr` for div popups to work correctly.


monzta:
Which "div popups" exactly?

devunt:
image

Go to http://item.gmarket.co.kr/Item?goodscode=919630134, and click the above white button.

PR #501, opened 2017-08-11
Create easylist_adservers_dmca_removals.txt

a-raccoon:
New DMCA removals list. Just covering our butts!

### DIFF
--- a/easylist/easylist_adservers_dmca_removals.txt
+++ b/easylist/easylist_adservers_dmca_removals.txt
@@ -1,0 +1,3 @@
+! Adservers permanently removed from EasyList due to DMCA legal claim. (Just covering our butts!)
+! Please do not add the following domains back onto the list.
+||functionalclam.com^$third-party


Redfoxmoon3:
Yes because complying with false DMCA is the correct thing to do. No.

zapzapzap1:
@Redfoxmoon3 I, uh, don't think that's their intention...

patricknelson:
Perfectly valid to keep a running list of domains removed due to DMCA. Being open source, it is in our best interest to have full disclosure. Also: No ad blocker should be allowed to automatically include this list, as it is against the spirit of the DMCA removals.

hook54321a:
@patricknelson I'm not sure that anything could be done to "forbid" ad blockers from automatically including this list since it is open source.

patricknelson:
Sure. To ensure the posterity of it, I wanted us to start out with a culture of encouraging it's existence for reference purposes only (how you use it is up to you and we cannot control that).

alphapapa:
> Sure. To ensure the posterity of it, I wanted us to start out with a culture of encouraging it's existence for reference purposes only (how you use it is up to you and we cannot control that).

*wink wink*?  If we're talking legally, I don't think that flies in court, e.g. torrent tracker sites who "don't host infringing content" but still get ruled against because everyone knows their intended purpose.

a-raccoon:
Without this file, it would be functionally impossible to prevent site(s) from accidentally being re-added to EasyList in a year from now.  Assuming we live in a DMCA world now, this is our only real solution.

Mind you that Google keeps a similarly open public record of URIs removed from search results due to DMCA takedown. You can find link to that list of URIs at the bottom of any Google search results page that would have contained those same URIs.

Redfoxmoon3:
It's not a VALID dmca, if this were to become valid; you could use the exact same logic if you were a malware author; antivirus companies catch your shit, you DMCA the companies to have them remove you from their database. It's _INSANITY_

a-raccoon:
I'm not going to disagree with that opinion.  However, this specific edit is going on the assumption that the DMCA assertion is going to be enforced by GitHub for the foreseeable future.  We can talk about removing this file in 2 years when the DMCA finally gets challenged in court.

ryanbr:
You're free to browse the list history (we made no secret of it), and keeping a list for the sake of keeping a list isn't productive.

PR #503, opened 2017-08-12
Local list creation, README, forbidden domains

bbb31:
Created a tool `genplates.py` to generate the lists locally.

I added a README; it indicates how to use `genplates.py`.

I created a `forbidden` list that contains DMCA take-down domains.

I added potentially linked "forbidden" domains and block them outright.

I've added a few element hiding rules as well.


ryanbr:
Blocking Admiral can be problematic, and will often disabled the site completely.

ryanbr:
Also, if you have specific issues, create seperate reports.

PR #512, opened 2017-08-22
Also whitelist adex.png, to avoid adex logo being blacklisted

Ivshti:
I am from adex.network, the site that we whitelisted on easylist a few weeks ago

Now we have a different issue: our logo is getting blocked on this web page: https://coinmarketcap.com/assets/adex/

I think that adding such an exception makes sense and fixes the issue

### DIFF
--- a/easylist/easylist_general_block.txt
+++ b/easylist/easylist_general_block.txt
@@ -1605,7 +1605,7 @@
 /adError/*
 /adevent.$domain=~adevent.com
 /adevents.$domain=~adevents.com.au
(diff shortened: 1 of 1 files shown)


ryanbr:
looks resolved, whitelist not needed

Ivshti:
Unfortunately it was resolved with a workaround. 

I still no reason why adex.png wouldn't be whitelisted - this blacklist should cover domains, not files IMO

ryanbr:
Will happen, if a user decides to create a file or folder with a "ad" like names. We can't avoid it.

PR #673, opened 2017-10-18
Remove : datadome.co

vasyl-vaskul:
As @Khrin  [said](https://forums.lanik.us/viewtopic.php?p=119756#p119756), we are not a tracking company. Our business is to protect websites from scraping bots. In order to do so, we have an Javascript tag that will get elements from the end-user browser not to have a global fingerprint of a device but to see if the elements returned by a device are coherent (and not an emulated browser like Selenium, Chrome Headless and so on).

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -530,7 +530,6 @@
 ||data-analytics.jp^$third-party
 ||databrain.com^$third-party
 ||datacaciques.com^$third-party
-||datadome.co^$third-party
 ||datafeedfile.com^$third-party
 ||datam.com^$third-party
 ||datamind.ru^$third-party


monsieurjplus:
I can confirm the activity of DataDome as my company is a customer.
They are not tracking the users, just filtering the traffic to protect the website from scraping or spam bots.
And their presence in this list is a problem for us because their Captcha page is not rendered on our website, so the false positives can't pass their captcha!
Please remove it from the list.
Thank you

Khrin:
Perfect, so you can give us an example link, right?

monsieurjplus:
Hello @Khrin 

Yes, there is 1 of my websites equiped with DataDome: https://subwaystats.com/

Thank you

bfabre:
I'm Benjamin, DataDome CTO. 
As Vasyl explained, we are a security solution that detect and block Bot activity. 
Because of this, EasyList users will most likely be considered as robot (because JS won't be triggered), and users may received a Captcha on our customer's websites (some example can be seen on our website).

PR #683, opened 2017-10-25
Whitelisting VMware Clarity libraries

mathisscott:
I tried to be as explicit as possible and placed this next to the other whitelisted entry using the blocked “clarity-*” string.

Please see:

https://forums.lanik.us/viewtopic.php?f=64&t=35937&p=117983&hilit=clarity#p117983

…and…

https://github.com/vmware/clarity/issues/850

…for reference.

I avoided using RegExp because it appeared that the library in general avoided them (likely for performance reasons as listed in the AdBlock dox).

Signed-off-by: Scott Mathis <smathis@vmware.com>

### DIFF
--- a/easyprivacy/easyprivacy_whitelist.txt
+++ b/easyprivacy/easyprivacy_whitelist.txt
@@ -1,3 +1,7 @@
+@@/clarity-angular.js$script,~third-party
+@@/clarity-angular.min.js$script,~third-party
+@@/clarity-icons.js$script,~third-party
+@@/clarity-icons.min.js$script,~third-party
 @@/cdn-cgi/pe/bag2?*geoiplookup$xmlhttprequest,domain=orain.org
 @@/cdn-cgi/pe/bag2?*google-analytics.com%2Fanalytics.js$domain=amypink.de|biznews.com|cryptospot.me|dailycaller.com|droid-life.com|forwardprogressives.com|fpif.org|fullpotentialma.com|geekzone.co.nz|goldsday.com|hoyentec.com|imageupload.co.uk|is-arquitectura.es|lazygamer.net|lingholic.com|mmanews.com|nehandaradio.com|nmac.to|orain.org|tvunblock.com|unilad.co.uk|xrussianteens.com|youngcons.com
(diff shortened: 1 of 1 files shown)


SMed79:
First your filters are wrong & sure will not be merged.

Thank you for noise.

mathisscott:
@SMed79 
Thanks for the feedback.

I'd like to try to get these changes in. Any tips as to how I can improve the filters to do so?

mathisscott:
@SMed79 
I think I got it. It's a little broad. But I'm forcing it to only load from the same host. Please let me know if there is something else I can do to improve the PR.

LanikSJ:
> I'm actually surprised that there haven't been ad devs that haven't figured out that they can name their files "bootstrap[something]" or "jquery[something]" to bypass this ad blocker. That seems like a low bar to me.

As you've said above your PR will allow anyone to use any of the files you've whitelisted to bypass any script blocking.  That's not a workable solution.  But nice try. 🙂

mathisscott:
@LanikSJ
So what can we do? This seems very targeted at our project. Is the only recourse to change our project name over and over and over?

The overly general restriction on anything named clarity is a little over the top.

SMed79:
> So what can we do?

Fix it from your side (I speak for myself as i'm not an EL authors).

LanikSJ:
@mathisscott 
I've said it before if you like I'll repeat it again here.
> Ask VMware to use a different file name. 
You need to rename your JS file names.  I get that they are part of the NPM package you're installing, however if you want this problem to go away that's your only viable option.

(I speak for myself also not a EL author)

Seeing as no author has replied with a conflicting point of view, take it for what's it worth.

Khrin:
This request is completely unacceptable for me.

xrobau:
Well, the only option left is to ensure everyone saves their project with the filename jquery.min.js - that will ensure it's never blocked by EasyList.

LanikSJ:
Yeap that's totally going to work 👍  not like EL doesn't have 100s of entries for jquery blocking it.

/sarcasm off

PR #690, opened 2017-10-31
Brazilian Site Specific Hiding Rules and more

lfmnovaes:
. Created a place to add these rules
. Removed www.dostavanadom.com, site doesn't exist anymore
. Added globo.com paywall annoyance block (ref: https://forums.lanik.us/viewtopic.php?f=62&t=38837 )


Khrin:
Sorry, but we don't block paywalls.

PR #698, opened 2017-11-03
Update easylist_general_block.txt

Firminator:
Added corrected HomepageAdvertis**e**mentBottom. as it had a typo.


ryanbr:
Keep in mind these aren't always typos. Sites will mispell elements to avoid detection.

PR #699, opened 2017-11-06
whitelist of adpage.io

jordivantuil:


Khrin:
I haven't understood why this change should be necessary.

jordivantuil:
I can imagine that. 
AdPage.io is a platform where you can build landingpages.
Because our domain contains "ad", our application doesn't load the required assets.

I've created a demo account for you so you can see where the application is about.
You can login on https://app.adpage.io using demo/demo.

Thanks for your time.

ryanbr:
Without working examples, I don't trust this request.

PR #749, opened 2017-12-04
Fix mail.yahoo.com selector.

kaewka:
Fix mail.yahoo.com selector.


kaewka:
@SMed79 Could you merge this?  This rule removed the first item in the list.  For paid user (ad-free user) who doesn't have ads in Yahoo Mail, this rule remove the first message in the message list.  In this PR, I only target the ad node and not the first item in message list.  Thanks

SMed79:
ONLY the EasyList [Authors](https://easylist.to/pages/about.html) can modify filters & i am not one of them.

@ryanbr @monzta FYI

kaewka:
Thanks @SMed79 

@ryanbr @monzta Can you look into this?  I am an engineer from Yahoo Mail.  We got a lot of complains from our paid users (we do not display ads for paid users) that the first message is missing.  We simplified the ad node, so that it is now very easy to block it and doesn't break user's experience.

Thanks

PR #759, opened 2017-12-12
A: https://www.amazon.com/

Coder-256:
uBlockOrigin/uAssets#793


ryanbr:
This would be considered selfpromo's, would be different if it was advertising 3rd-party sites. Otherwise, its fine, add it to your personal list.

PR #784, opened 2017-12-20
A: http://www.sfgate.com/

SMed79:


Khrin:
Isn't a commenting system? (I haven't checked yet, just from what I can recall...)

PR #894, opened 2018-02-07
changing `netlify.com` to `inspiring-dijkstra-4f30b2`

depadiernos:
`netlify.com` is a web hosting company that hosts a great number of sites. The offending script is hosted on the `inspiring-dijkstra-4f30b2` subdomain, which has already been disabled by Netlify. This commit would allow all the legitimate users of Netlify to continue using adblockers that include easylist.

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -2284,7 +2284,7 @@
 ||zylom.com^*/tracking_spotlight.js
 ||zytpirwai.net/track/
 ! Mining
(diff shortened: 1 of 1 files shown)


Khrin:
Merge request closed because would invalid our filter. Do you have any tangible example of this false positive?

depadiernos:
so the script was hosted on inspiring-dijkstra-4f30b2.netlify.com, we've blocked that site as soon as we were made aware of it. It's no longer being served so a tangible example would be difficult. What would we need to provide to have netlify.com removed from the list?

depadiernos:
I made another commit that would just remove the subdomain that I replaced `netlify.com` with.

PR #900, opened 2018-02-10
Allow OpenStreetMap.org events self-promotion

Firefishy:
PR to remove OpenStreetMap.org which was added in b5c76ea17e12b95c711ce512c2389795fc853bd0

OpenStreetMap.org's banner for announcing our own community conferences and events. This isn't a banner for advertising 3rd party things.

Please don't block our ability to advertise our conferences and our community events on our own site.

PR #906, opened 2018-02-12
A: unlist https://www.openstreetmap.org/ self-promotion

Firefishy:
Rebase #900


Khrin:
Since the previous request was refused just a few hours ago, why do you really expect that opening a new perfectly identical request, even without any sort explanation, could receive a different answer? Really, I want to understand the logic behind this move.

Firefishy:
Because I think is absurd you would block us from self-promoting our annual Italian & Germany community conferences. Specifically targeting them would seem to be against your listed policy.

I revised the commit to cleanly merge in-case you decided to have a change of heart.

Firefishy:
Reference workaround we declined https://github.com/openstreetmap/openstreetmap-website/pull/1743

PedaB:
Reading [your definition of "advert"](https://easylist.to/2011/07/11/the-definition-of-advert-and-link-exchange-policy.html) that @ryanbr is referring to, our event notes are clearly no adverts. The definition states "... in return for goods or services" but our local conferences (in this case the Italian and the German one) are even free to participate in and are merely a factual information about a conference of a sub-community of ours.

pnorman:
Quoting from the policy,

> "Promotion"

Yes, it is promoting.

> "Third-party"
> The aim of a website is to entice people to read more of its content, which visitors presumably desire to do, as indicated by their presence on the domain; they should therefore not be hindered in doing so by a filter subscription. Whether websites are third-party or first-party (what happens if one person owns two domains?) is a question for the EasyList authors to consider.

The audience for osm.org is the OSM community. The goal is to entice them to view other sites within the OSM community - that of conferences the community runs. 

With the OSM community, events are done under different legal organizations, but it's all part of the same community.

> "... in return for goods or services"
> An important question that has been raised is whether an advert requires the exchange of money, to which I would respond that it is not an essential element. Money may be one of the goods supplied, but there may also be services provided in return. Therefore a link to, say, a web host would be removed if this is a sponsored link required as a provision of service, but not if the description is merely factual, as this would not be promotion.

Money is not exchanged. To the extent that services are exchanged, it is that both the OSMF and the body organizing the conference are part of the OSM community.

I haven't discussed the appearance of the image, if it is intrusive, it's size, or anything like that. The definition of advert in the policy does depend on these.

ryanbr:
**Easy fix;** Make the image into a text link instead. Text ahref linking to an external site is fine obviously.

Again, this issue report will also be closed because nothing new will be added. 

Quick Example of [ self-promo.](https://i.imgur.com/RtHaYKA.png) I found. Assuming they're not considered intrusive, they'll be left alone.

Its pretty obvious openstreetmap ads arent self-promos (see reference image above).  **Openstreetmap is advertising services for third-party sites via an image banner. IT IS AN ADVERT, therefore its blocked.**

ryanbr:
By the way, I'm not trying to be the bad guy here, we (as a adblock community) have seen many iterations of adverts, self-promo, native advertising etc. What Openstreetmap is doing isn't really anything new to be honest.

I understand Openstreetmap is trying push the details of another site out, but how its done is considered an advert. Sure its not the worst type of adverts (going through ad/tracking companys) but its still promoting another site in a Banner Ad.

Just fix it on the server side, just use a text link instead so we can all move on.

PR #915, opened 2018-02-16
Fix ##.uk-alert.

pniaps:
Commit 2947f27f6cfa52fd38980e7a81a59fc503285e2f blocks things on
websites done with uikit, including official uikit documentation on
getuikit.com

### DIFF
--- a/fanboy-addon/fanboy_annoyance_general_hide.txt
+++ b/fanboy-addon/fanboy_annoyance_general_hide.txt
@@ -8611,7 +8611,6 @@
 ##.ui-cookie-warning
 ##.ui-cookieAgree
 ##.ui-dialog-cookies
-##.uk-alert
 ##.uk-cookie
 ##.uk-cookie-banner
 ##.uk-cookies
--- a/fanboy-addon/fanboy_annoyance_specific_hide.txt
+++ b/fanboy-addon/fanboy_annoyance_specific_hide.txt
@@ -3988,6 +3988,8 @@ tbs.com,tntdrama.com##.toast
 twitch.tv##.toast-manager__container
 hankooktire.com##.top_ep_wrap
 mabknives.co.uk##.topbarBox
+superiorwallpapers.com##.uk-alert
+wallpapermania.eu##.uk-alert
 vodafone-wifi.com##.uparrow
 anisearch.com##.urgent-msg
 cnn.com##.user-msg


ryanbr:
Browsed; `https://getuikit.com/docs/avoiding-conflicts`  and other options on the left, seems fine?

pniaps:
That document has nothing to do with this. It only explains how to scope or change the prefix.

You dont want to block all ".uk-alert" elements on all sites (i think), you only want to block the cookies message on wallpapermania.eu and superiorwallpapers.com.

The only solution at the moment is stop using this list.

Check next example from https://getuikit.com/v2/docs/alert.html

With list disabled:
image

With list enabled
image

Do you really think this is correct?

ryanbr:
Okay, just whitelisted the site.

https://github.com/easylist/easylist/commit/8f6a30ea85f14ca1d0a2389d0fc1c31842fc0f99

pniaps:
Not only that site, instead all sites done with uikit. You are blocking things in many many sites done with uikit, in fact, wallpapermania.eu and superiorwallpapers.com are done with uikit.

It's like bootstrap, a web front-end framework, so there can be many many sites using this class.

My pull request reverts your change when this 'was broken'.

Thanks in advance.

ryanbr:
From what I've seen so far, except the this site. its used for cookie messages (in both wallpapermania.eu and superiorwallpapers.com)

pniaps:
Thats is only because they have used the class "uk-alert" in the div of the cookies.

Why don't you block the 'div' tag instead? because, that two sites has the cookies messages in a div.

How many sites do you know using uikit to be sure that class is used for cookies messages?

All themes from https://yootheme.com/themes (the creators of uikit) uses uikit and people using it can use class "uk-alert" anywhere, for example to highlite something in an article.

More examples of templates done with uikit:

Altair Admin Material Design UIkit Template
http://altair_html.tzdthemes.com/components_common.html

Responsive Knowledge Base & FAQ HTML Template
http://demo3.pressapps.co/html/knowledge/components.html

Courage - A Clean WordPress Blogging Theme
https://demo.bahemian.com/courage/uikit-elements-2/

IT Delight - Gantry 5, eCommerce/J2Store Joomla Template
http://demo.inspiretheme.com/templates/delight/index.php/features/uikit-framework

pniaps:
maybe @saschadube can explain it better than me

ryanbr:
I had the same issue with previously with a site sharing the [##.notification](https://forums.lanik.us/viewtopic.php?f=64&t=28174&p=89399) for cookies and other notifications.

The only fix is on the server side. Change the cookie div's. Because all I see is sites using ##.uk-alert for cookie divs so far.

pniaps:
I Think you should use more restrictive selectors to block only that you need, not all.

The link of tweakers.net, you are breaking the functionality of the site, but here you are breaking the functionality on many sites because you are blocking class "uk-alert" in whole internet.

If you cannot block only the cookies message in wallpapermania.eu, at lest block "uk-alert" only in wallpapermania.eu

The fix is not in the server.
Do you want all people using "uk-alert" change it, to fix your mistake?
If wallpapermania.eu had a class "fb_content" in the cookies message, will you block ##.fb_content and ask Facebook to fix in their side?

ryanbr:
https://github.com/easylist/easylist/commit/0a4925787badcc38052194b5264d4c202db5202b

i've reverted it, but if sites continue to use this specific element for cookie classes. Then they expect to be blocked.

pniaps:
Ok, thanks.

PR #923, opened 2018-02-22
M:(NSFW) https://hitomi.la/

forgodssake:
Adjusting the filter to only affect the specific site

Updates comit 2af3db94fd2d711591c100900a83b2c453f4240e

### DIFF
--- a/easylist/easylist_general_block.txt
+++ b/easylist/easylist_general_block.txt
@@ -4177,7 +4177,6 @@
 /exo120x60.
 /exobanner.
 /exoclick.$domain=~exoclick.kayako.com
-/exoclick/*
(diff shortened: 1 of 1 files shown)


Khrin:
Why this change should be necessary?

forgodssake:
ExoClick's twitter timeline is broken.
ExoClick's corporate website has broken parts because some .js have exoclick in it's path
and I guess many other public websites that use the username in it's path will have broken parts because this filter is too generic

SMed79:
https://github.com/easylist/easylist/commit/3fe27dd94d9bb4dd20c01a263a10cb8032ec0a3d

PR #931, opened 2018-02-25
A: https://t.17track.net/en#nums=123456789

SMed79:


Khrin:
I don't feel myself enough confident to add them as generic filters... Sorry

PR #935, opened 2018-02-27
Added "Battle for the net" pop ups

Julian2000nl:
This will block out the annoying "battle for the net" pop ups. It completely prevents them from loading so it's impossible for them to store cookies/track the user. 

Tested and works flawless.


ryanbr:
Rather than this commit, you need an issue report showing the origin the popups.

PR #939, opened 2018-03-01
Simplified and corrected Zazzle entries

rlamoni:
The entries that allowed Zazzle's CDN to serve images on Zazzle's websites were not broad enough.  Also, Zazzle has 4 English websites 4 different English-speaking countries.

### DIFF
--- a/easylist/easylist_whitelist.txt
+++ b/easylist/easylist_whitelist.txt
@@ -2023,8 +2023,10 @@
 @@||zap2it.com/ads/newsletter/$image,~third-party
 @@||zattoo.com/?advideo/*;vidAS=PRE_ROLL;$object-subrequest
 @@||zattoo.com/advertising/channelswitch/$subdocument
-@@||zcache.com*/advertising$image,domain=zazzle.com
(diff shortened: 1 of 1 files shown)


ryanbr:
Too generic, wholesale whitelisting of the site like this is over the top.

rlamoni:
We must have made a mistake.  Our intention was to whitelist only images being viewed on the zazzle sites from the zazzle image CDN (known as zcache).  We had not intended to whitelist the site for everyone (although, since zcache does not host ads, whitelisting it globally doesn't seem too unreasonable).

Can you help us craft a set of rules that whitelist images hosted from our CDN domains on our sites? 

If you still think we are lying about being a retailer (and not a ad-network) our legal team is more than happy to work up a contract with you that includes assurances about our business model and commitment to keeping our stores ad-free.

ryanbr:
Okay, looked into it further, they seem sane enough.

https://github.com/easylist/easylist/commit/ee60342882f33bf80eebf57bb04e731053c34d3a

SMed79:
Link for reference https://forums.lanik.us/viewtopic.php?f=64&t=39984

PR #955, opened 2018-03-11
P: (tumblr.com) https://forums.lanik.us/viewtopic.php?f=64&t=40185

SMed79:


SMed79:
@ryanbr fyi.

nightpool:
an exception for http(s)://new-xkit.github.io would also solve the problem for us. We've had hundreds of users in our chat rooms and ask box over the past two days due to this change, and likely 10x more who haven't reached out to us or haven't reached the 24 hour window where the extension checks for updates yet.

nightpool:
any update on this? we're on day two of triaging support requests.

Khrin:
Our filter has been modified. 
https://github.com/easylist/easylist/commit/536b5ece1fa1cced0ea2144c9a22a5b124062e2f

PR #981, opened 2018-03-27
A: https://go.marybaldwin.edu/forums/topic/pacific-gas-and-electric-pgande/

SMed79:
```
https://essayerudite.com/images/banner/500x500.jpg
```
Comming from a spam but still safe to block it.

e.g. at `http://mountblade.pl/showthread.php?tid=195715`


Khrin:
No problem for me to block it, but if you don't mind I prefer to consider it as "third party" element.

PR #984, opened 2018-03-28
https://temp-mail.org/    - corrected a typo & added some rules

OpenBSD-Neurotik:
# # 976


Khrin:
1) Not a typo, filter is correct.
2) Not something that would belong to EasyList.

Issue #979, opened 2018-03-26
buonpertutti.it/buoni-sconto/ cannot access coupons any more


lutzhoeger:
### List the website(s) you're having issues:

`http://buonpertutti.corriere.it/buoni-sconto/`

### What happens?

Coupons are not loaded any more. The page stays empty. Access to valassis.eu is being blocked.

### List Subscriptions you're using:

Easylist

### Your settings

- OS/version: macOS 10.13.4
- Browser/version: Firefoy 59.0.1
- Adblock Extension/version: uBlock Origin 1.15.18

### Other details:

Issue caused by easylist/easylist_adservers.txt, blocking all valassis.eu

valassis.eu has been added to the list by user @Khrin (commit 25d1f7d)

Please note, that valassis.eu is NOT an adserver; it is being used to merely serve coupons, which are embedded upon site owner's requests. At the moment, this issue blocks users across Europe from accessing & printing coupons.

see two screens for coupons blocked and coupons not blocked

![bpt-couponsblocked](https://user-images.githubusercontent.com/2639902/37890522-23246c2e-30d1-11e8-8506-93bd3bca4b61.png)
![bpt-couponsnotblocked](https://user-images.githubusercontent.com/2639902/37890528-25fe37e0-30d1-11e8-80da-b70627533ab7.png)

Thank You!

Khrin:
The filter has been removed.

PR #987, opened 2018-03-28
Added tracking domains found on several sites

OpenBSD-Neurotik:
#979 

Related to:
http://buonpertutti.corriere.it/buoni-sconto/
https://www.hogarmania.com/
http://www.valesycupones.es/
http://www.yosoycuponero.es/
https://www.ahorradoras.com/


Khrin:
Some of them are already present, other aren't even trackers.

Issue #969, opened 2018-03-20
theoutline.com


prodan1234:
**List the website(s) you're having issues:**

```
theoutline.com
```

**What happens?**

Advertising carousel in the middle of the article view not blocked. 

**List Subscriptions you're using:**

easylist

[Screenshot](https://i.imgur.com/y4NLt14.png)

- OS/version: macOS 10.13.3
- Browser/version:  Chrome 66.0.3359.33 
- Adblock Extension/version: uBlock 1.15.2

SMed79:
https://forums.lanik.us/viewtopic.php?f=62&t=40263

PR #989, opened 2018-03-28
Added hide advertisements for theoutline.com

OpenBSD-Neurotik:
#969


Khrin:
Could broke scrolling between articles.

PR #992, opened 2018-03-28
Added anti-track for Printhero.com

OpenBSD-Neurotik:


Khrin:
First filter is already present, second and third filters would likely breakup the entire site.

PR #993, opened 2018-03-28
block popular social tracker scripts

OpenBSD-Neurotik:


Khrin:
Example url?

OpenBSD-Neurotik:
https://trends.builtwith.com/websitelist/ShareThis

So on Liveleak.com for example the script url is:
```
https://platform-api.sharethis.com/js/sharethis.js#property=58f878f5caaba30012e7e309&product=inline-share-buttons

snippet:
protocol+"://count-server.sharethis.com/v2.0/get_counts?"+e.qs({cb:"window.__sharethis__."+s,url:t.url,refDomain:e.hostname(document.referrer),sop:!0})))},e.loadFacebookShareCount=function(t,o){var n;return n="cb"+e.uid(),e[n]=function(e){var t;return o(JSON.parse((null!=e&&null!=(t=e.share)?t.share_count:void 0)||0))},e.js(e.protocol+"://graph.facebook.com?"+e.qs({id:t,callback:"window.__sharethis__."+n}))},e.logGoogleAnalyticsEvent=function(e,t,o){var n,r;return r=window.ga,n=window._gaq,r?r("send","event",e,t,o):n?n.push(["_trackEvent",e,t,o]):void 0},e.newElement=function(t){var o,n;return void 0===t&&

```

They have even written toolbar malware:
https://malwaretips.com/blogs/ads-by-share-this-removal/

Khrin:
As far as I know, `sharethis.com` is a sharing service, where known trackers are already blocked by EasyPrivacy and the service itself is blocked by the Fanboy's Social List.

PR #1008, opened 2018-04-01
P: https://www.htmlgames.com/game/Daily+Bridges

SMed79:


monzta:
No issue for me. Any particular config that is necessary to test with?

SMed79:
The problem was that the game does not play but it seems fine now.

PR #1020, opened 2018-04-05
A: http://www.digg.com/

gtdeng:
Add new `www.digg.com` filters


monzta:
Where do these filters remove what exactly?

monzta:
I'm not seeing any sponsored stuff. Do I need to log in?

PR #1047, opened 2018-04-17
Remove filter: `/newrelic.*`

detro:

`http(s)://newrelic.*`

### What happens?

New Relic internal services (i.e. `newrelic.SERVICE_HERE.com`) don't work properly. All sort of different resources get blocked.

EasyPrivacy

### Your settings

- OS/version: any
- Browser/version: any that supports extensions
- Adblock Extension/version: AdBlock, AdBlockPlus, Fanboy

### Other details:

It's understandable for this project to block beacon and tracking, but blocking an entire domain is extreme (and doesn't seem to be the norm for any other domain).

New Relic is a legitimate business and should be treated as such.

The rule I'm removing doesn't help any user, but just disrupts anybody using `newrelic.*` domains, i.e. New Relic employees.

In fact, the rule just means: `block any 'newrelic.*' EXCEPT on '*.newrelic.com'`.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -2080,7 +2080,6 @@
 /neustar.beacon.
 /new.cnt.aspx?
 /newlog.php?
-/newrelic.$domain=~newrelic.com
 /newscount/*
 /newSophus/*
 /newsstat/*


wulczer:
Yeah, this rule seems to be a bit too broad. It's blocking access to pages like https://newrelic.zendesk.com/ and others that follow that pattern of giving customers their own subdomains.

Full disclosure, both @detro and myself work for New Relic, but this change does indeed feel misguided.

ryanbr:
note we block a lot of "legitimate businesses", tracking and likewise. I've tweaked the filter. Not sure why newrelic employees would use a tracking list in the first place.

detro:
Thanks for the tweak, just blocking `/newrelic.js` makes much more sense.

About your "perplexity": why wouldn't we use an AdBlock extension in our browser? You guys maintain amazingly good lists that make browsing lighter, faster and reduces the burden of intrusive ads: it's totally in our interest to use those lists (plus, they come pre-packaged and auto-updated with the extension).

This is why it was so annoying: it was written in a way that blocked WAY more than it clearly intended to block.

PR #1108, opened 2018-04-29
Remove nut.cc from the list

marmistrz:
nut.cc is a free domain provider. Websites just taking advantage of a free domain should not be blocked.

Let me know if there's a more specific reason for the whole DNS subtree being blocked.


marmistrz:
@ryanbr with 41604123e76f0ab454438b341f135816053912b9 having hit the master should we close this PR?

PR #1123, opened 2018-05-05
M: Minor change (regex popup)

smed79:


smed79:
https://github.com/easylist/easylist/commit/dbd670b9#r28945689

PR #1145, opened 2018-05-11
A: http://community.tollwood-festival.info/ (ryanbr/fanboy-adblock#375)

smed79:


ryanbr:
Should be taken care of already

PR #1166, opened 2018-05-18
M: Update Propellerads filter

smed79:
propellerads
* `http://xijgedjgg5f55.com/1526682/?`
* https://forums.lanik.us/viewtopic.php?p=136783#p136783


smed79:
@ryanbr fyi.

PR #1217, opened 2018-05-26
A: http://www.nemzetisport.hu/minden_mas_foci/szombati-sportmusor-0526-2636793

smed79:


ryanbr:
Should be resolved now. Update subs.

PR #1338, opened 2018-06-12
A: (opened email tracker) https://madmimi.com/

smed79:
sbl


ntoskrnl-x86:
IIRC, they use multiple subdomains for this.

PR #1341, opened 2018-06-13
remove gravityrd-services.com

torzsmokus:
This domain is used for getting in-site recommendations and search results from Yusp by Gravity R&D on many sites around the web. I consider it an error to block the whole domain altogether, as it degrades the normal operation (e.g. predictive search) of certain sites.

### DIFF
--- a/fanboy-addon/fanboy_annoyance_thirdparty.txt
+++ b/fanboy-addon/fanboy_annoyance_thirdparty.txt
@@ -132,7 +132,6 @@
 ||google.com/uds/?file=orkut&
 ||goroost.com^$third-party
 ||gravity.com^$third-party
-||gravityrd-services.com^$third-party
 ||grazeit.com^$third-party
 ||groupon.com^*/affiliate_widget/$third-party
 ||grvmedia.com^$third-party


ryanbr:
Its in the Annoyances list for a reason, limiting many 3rd party scripts that try to "enhance the user experience".

torzsmokus:
@ryanbr 
Let me disagree. I would not consider e.g. the predictive search box on `www.tutti.ch` an _annoyance_ or _trying to "enhance the user experience"_. It is rather part of the core functionality of the site, implemented by an external vendor (Yusp by Gravity R&D).

If you find our recommendations on a specific site annoying, please block those ones specifically instead of crippling several hundreds of sites.

ryanbr:
Annoyances isn't a required list, its an optional, user-selectable list. It'll block unnecessary third-party scripts from injecting into a site. Thus this site was blocked in this list. Lets not to hype up your services, its not "crippling" these sites.
Just don't use the Annoyances list if you don't agree.

torzsmokus:
> block unnecessary third-party scripts 
but our script is _necessary_ for the _normal_ operation of many sites, as I already explained above :confused:

PR #1356, opened 2018-06-14
A: https://www.fifa.com/

smed79:
```
https://www.fifa.com/worldcup/
https://www.fifa.com/worldcup/destination/index.html
```

https://forums.lanik.us/viewtopic.php?f=62&t=41028


ryanbr:
Why did you remove those filters?

smed79:
Sorry, it was not intentional. 
I was editing the same file in two windows and I did not pay attention when I pushed the PR.

PR #1379, opened 2018-06-18
exclude affiliprint.com/advertiser/

maldn:


smed79:
Why this exception? What is the issue? How to reproduce?

Khrin:
Why this change is required? The url is inaccessible for me.

PR #1400, opened 2018-06-21
M: adjusted bebi.com script filter

smed79:
Changed `...wp-includes))(.*)/...` to `...wp-includes)\/)(.*)/...`


smed79:
@ryanbr FYI.

PR #1437, opened 2018-06-28
A: https://tusfiles.net/q4za8cgkvghg

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=33062
https://forums.lanik.us/viewtopic.php?f=62&t=40707


hendricha:
Hi! I would like to ask what was the reason for blocking djtflbt20bdde.cloudfront.net ? It is as CDN used by helpscout.net . It is indeed used on tusfiles, but also used on several sites for user input. (Including our own.) And while I see why it would be a privacy issue (third party url visited, therefore third party tracking), and thus it would be okay to be on the easy privacy list, I don't think it directly serves ads.

PR #1601, opened 2018-07-26
A: http://igg-games.com/

smed79:


smed79:
@ryanbr fyi

peace2000:
Conflicts now.

peace2000:
What do these rules block?

PR #1621, opened 2018-07-30
block natero

benwaffle:
The analytics script comes from https://event.natero.com/scripts/natero_analytics.min.js and the events are posted to `events.natero.com`


Khrin:
Have you an example link where this script it is used?

smed79:
@Khrin https://forums.lanik.us/viewtopic.php?f=62&t=41304

benwaffle:
What else needs to be done for this to be merged?

smed79:
@ryanbr fyi.

benwaffle:
cool, thanks

PR #1637, opened 2018-07-31
Whitelisted .fa-rss on hearthis.at

adranger:
Fixed sidebar for logged in users on `https://hearthis.at/`. This is the built-in site "feed" showing recent episodes, not a link to an RSS feed or an external page.

heart


adranger:
@ryanbr Why don't you want this to be fixed?

You've ignored this too, even though @okiehsch has confirmed it: https://github.com/ryanbr/fanboy-adblock/issues/510

PR #1707, opened 2018-08-16
https://forums.lanik.us/viewtopic.php?p=139673#p139673

mapx-:


mapx-:
@Khrin  @ryanbr

smed79:
@mapx- 

Please, when you push a commit following the [correct syntax](https://forums.lanik.us/viewtopic.php?p=131958#p131958), this will much help for maintenance.

> **A:** (as Addition) : means that a filter to block an advertisement has been added...
> **P:** (as Problem) : means that a problem has been fixed...
> **M:** (as Moved or Modified) : means that a filter has been moved, added or deleted...

This will much help for maintenance

Thank you.

PR #1717, opened 2018-08-17
M: Update upload.so domains (added uplod.io & uplod.ws / removed uplod.it)

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=41442


smed79:
@ryanbr

PR #1718, opened 2018-08-18
A: http://us.chantelle.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=34261


smed79:
@ryanbr Please also check this old report https://forums.lanik.us/viewtopic.php?f=62&t=37133

PR #1721, opened 2018-08-19
A: https://download.cnet.com/

smed79:


Khrin:
We had some problem blocking this domain...

PR #1836, opened 2018-09-15
A: https://www.mitele.es/directo/cuatro

smed79:


peace2000:
`##.opacity_layer_consent` Doesn't seem to block anything.

smed79:
@Khrin @ryanbr `##.opacity_layer_consent` still a good generic.

PR #1843, opened 2018-09-16
M: Update bebi script

smed79:
Excluded the folder "ajax" (e.g. `https://www1.9anime.to/watch/mo-dao-zu-shi.ploj/qpo345`)


smed79:
@ryanbr

PR #1855, opened 2018-09-19
A: simpleanalytics.io

costpermille:
```
https://cdn.simpleanalytics.io/hello.js
https://cdn.simpleanalytics.io/hello.gif
https://api.simpleanalytics.io
https://api.simpleanalytics.io/post
https://simpleanalytics.github.io
```

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1830,6 +1830,7 @@
 ||silvergamed.com^$third-party
 ||silverpop.com^$third-party
 ||silverpush.co^$third-party
+||simpleanalytics.io^$third-party
 ||simpleheatmaps.com^$third-party
 ||simplehitcounter.com^$third-party
 ||simplereach.com^$third-party


ryanbr:
Denied, need an example site.

costpermille:
`https://simpleanalytics.io` contains it at the end:

```
  <script async src="https://cdn.simpleanalytics.io/hello.js"></script>
  <noscript><img src="https://api.simpleanalytics.io/hello.gif" alt=""></noscript>
```

Khrin:
We need an example site (a real one, no demos) where this service is actually used.

costpermille:
Thanks for the clarification - here are some hits from ublock:
```
excuseme.wtf
ronaldlangeveld.com
```

ryanbr:
Why should it be removed (considering its being blocked by an anti-tracking list)

costpermille:
I am confused, this was an addition to the list (and was not blocked at submission time by EP), not a removal?

ryanbr:
oh my mistake.

adriaanvanrossum:
Creator of Simple Analytics here. I have a question for the maintainers of Easylist. I know it's needed to have a list like this to disable all the trackers and companies that build profiles on their visitors. Also with ads, it's really scary how companies tread privacy and I'm very concerned about how they take care of customer data. So first of all, I'm happy this list exists.

But what happens if we block alternatives even if those alternatives are taking the privacy of the user very serious. I care about the privacy of the individual. I don't collect any personal information (I don't even store IP's). Even if you have your *Do Not Track*-setting turned on in the browser I do **not** collect **any** information (see [our script](https://github.com/simpleanalytics/cdn.simpleanalytics.io/blob/61c6f5e5c8204f65d6afaf58b4bc1f6ce2faa862/hello.js#L36-L37)).

So if we keep blocking the companies that do good, and respect the privacy of the users, what kind of sign is it to just block those companies? I think it's wrong and we shouldn't put every company on the list just because they are sending a request.

I'd love to talk about this further and see if I can address any concerns the maintainers or users of Easylist might have.

adriaanvanrossum:
> If you actually care, you would make it self hosted ...

I don't have the resources to built that at the moment. I'm very early stage and I need to make drastic changes in my code all the time (this will change in the future). I can't keep my code up to date on self hosted servers at the moment. I can reconsider this when I know my feature set is defined.

> ..., but I don't want to see my requests sent to a US company

That's a great point. I heart Iceland (https://www.ivpn.net/internet-privacy-laws-in-iceland) could be a good option. Do you have any advise on where to place the server?

PR #1864, opened 2018-09-20
A: (NSFW) https://rule34hentai.net/post/view/335140

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=41736


peace2000:
Still relevant?

DandelionSprout:
I am unable to reproduce the exact problems that would necessitate these entries. However, there are ***enormous*** problems with popups and popunders when having AdGuard Base and uBlock Filters turned off.

PR #1874, opened 2018-09-21
A: http://www.logitheque.com/articles/la_home_hub_2534.htm

smed79:
https://i.imgur.com/xweYJO3.png


peace2000:
That does not appear to me, maybe it's not in use anymore.

Issue #1966, opened 2018-10-01
mlb.com


smed79:
**website :**
* `https://www.mlb.com/`

**desciption :** (1)
* Sponsores banners

**filters :**
```
||prod-gameday.mlbstatic.com^*/images/schedule/postseason-sponsorship/
```

**desciption :** (2)
* Social icons

**filters :**

```
##.p-follow--sm
```
or
```
##.p-follow__logo--email
##.p-follow__logo--youtube
##.p-follow__logo--snapchat
##.p-follow__logo--instagram
##.p-follow__logo--twitter
##.p-follow__logo--facebook
```
or
```
##a[title="Follow on facebook"]
##a[title="Follow on twitter"]
##a[title="Follow on instagram"]
##a[title="Follow on snapchat"]
##a[title="Follow on youtube"]
##a[title="Follow on email"]
```

PR #1967, opened 2018-10-01
A: https://www.mlb.com/video/embed/c-2512690883?

smed79:


smed79:
See also [#1966](https://git.io/fxJne)

PR #1973, opened 2018-10-02
A: https://sp-today.com/

smed79:
anti adblock, see https://forums.lanik.us/viewtopic.php?f=62&t=41827


smed79:
for the anti adblock, see https://forums.lanik.us/viewtopic.php?f=62&t=41827

PR #2042, opened 2018-10-11
Adding Akamai browser fingerprinting paths

hugoearnest97:
Akamai deploy obfuscated scripts from these paths. One path has common scripts that can be accessed on any Akamai-served domain with fingerprinting enabled. The other path serves coded scripts that change per session. The two groups of scripts are structurally different, but use similar attacks.

The scripts are highly invasive and attack a broad surface of the web-browser. They assemble a large fingerprint and send it back to Akamai. They are served from the same domain as the visited site, because the site is deployed through Akamai's CDN.

Attacks used by the scripts include canvas fingerprinting, ActiveX, performance profiling, and font enumeration, plus many more.

PS: It's strange that a privacy list is managed on a forum that uses Google ReCaptcha and requires a personally-identifiable email address. I think this is a bad policy. It prevents privacy-interested people contributing. I tried joining the forum to discuss this first and could not.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -297,6 +297,7 @@
 /__utm.js
 /__varnish_geoip
 /__wsm.gif
+/_bm/*.js
 /_dts.gif?
 /_lib/ga.js
 /_owa.gif?
@@ -445,6 +446,7 @@
 /ajaxtracker.
 /ajx/ptrack/*
 /akam/*/pixel_
+/akam/10/*
 /akamai/pixy.gif
 /akamai_analytics_
 /AkamaiAnalytics.


Khrin:
Do you have an example site where to locate such resources?

(About the forum: @LanikSJ)

smed79:
@Khrin 

> Do you have an example site where to locate such resources?

for the folder `/_bm/`, all what i have find is
```
https://www.zara.com/_bm/bd-1-30
https://www.mobile.de/_bm/bd-1-30
https://www.fedex.com/_bm/bd-1-30
https://www.united.com/_bm/bd-1-30
https://www.fidelity.com/_bm/bd-1-30
https://www.hotels.com/_bm/bd-1-30
https://www.jcrew.com/_bm/bd-1-30
https://www.delta.com/_bm/bd-1-30
https://www.blibli.com/_bm/bd-1-30
https://www.capitalone.com/_bm/bd-1-30
https://www.groupon.com/_bm/bd-1-30
```

hugoearnest97:
@Khrin @smed79 - The paths above to those scripts, `bd-1-30` in each case. Those are all one of the fingerprinting scripts used by Akamai. They are served through Akamai, but they look as they come from the visited domain. Because Akamai is a CDN.

There are other scripts served from `/akam/10/` (or maybe `/akam/\d+/` that are different in structure. But they fingerprint the same techniques. Those other scripts are not the same filename for each server, or for different sessions. But they are very similar structure to one another.

I would like to reopen this, I did not get notifications from this PR and I do not log in every day.

smed79:
@hugoearnest97 Easylit authors never add filters without proof and until now, you did not provide an example site where we can see `/akam/10/`.

hugoearnest97:
@smed79 - Whether you receive an attacking script from `/akam/10/` or from `/_bm/` seems random. Or, maybe based on which CDN server you hit. Or maybe what value Akamai see in attacking you. I have seen scripts from both paths, from the same site, on different visits. I cannot give a guaranteed way to reproduce except to say visit the sites above a few times. Maybe with different browsers, or different IPs. Or with different cookiejars.

Maybe if I know more I will be able to guarantee a reproducable case. But right now it seems random.

hugoearnest97:
Also, the scripts from `/akam/10/` seem session-linked. When I save a URL for a script from that path, and visit later, it 404s. The scripts from `/_bm/` are not session-linked. Their hashes are the same when visited on many domains. So I have URLs for the `/akam/10/` scripts I have seen. But they no longer work.

hugoearnest97:
BTW other filenames I have seen in `/_bm/` are `/_bm/async.js` and `/_bm/bd-1-31`. These are pretty similar scripts to one another. Maybe they are different development versions of the same script. They might be served from different CDN servers or maybe at random.

smed79:
https://github.com/easylist/easylist/commit/aae705fbbfe1e043640ddcf7d1d55e4cf146fd31#commitcomment-31034716

PR #2083, opened 2018-10-16
Fix: Rule breaking survey embeds (topbox.ca)

injust:
Fixes ryanbr/fanboy-adblock#573.
easylist/easylist@0bd4898 added `~subdocument` to the wrong rule; the domain should have been `surveygizmo.com` instead of `surveymonkey.com`.


ryanbr:
https://github.com/easylist/easylist/commit/554e9e67682cb66028767e2864a2b49c827d28e3

PR #2131, opened 2018-10-24
block ads from Xiaomi

CrescendoFang:
block all ads from Xiaomi server.


smed79:
https://github.com/easylist/easylist/issues/2113

PR #2144, opened 2018-10-26
A: Update #2135

smed79:


smed79:
@ryanbr If you cannot reproduce, this is because the site serve a [different document](https://github.com/uBlockOrigin/uAssets/issues/3826).

PR #2251, opened 2018-11-08
A: http://btstors.com/ (Coin Mining)

smed79:
for details see https://git.io/fpkZJ


smed79:
Sites suspended after being reported.

PR #2279, opened 2018-11-12
M: translate.google.com

krystian3w:
Fix https://github.com/ryanbr/fanboy-adblock/issues/631

> When i click on the "Google Apps" button (see screenshot)
> obraz
> 
> it disappears (see screenshot)
> obraz
> 
> Check it out? uh?


krystian3w:
> Try a new browser with automatic translation feature.

PR #2280, opened 2018-11-12
 A: pull_request_template.md

krystian3w:
Add comment to all feature pull request with instruction what the letters mean.

```html
<!--
https://forums.lanik.us/viewtopic.php?p=131958#p131958

A: (as Addition) : means that a filter to block an advertisement has been added
P: (as Problem) : means that a problem has been fixed, for example by adding an exception in the case of a video that does not start, or in the case of an anti-adblock.
M: (as Moved or Modified) : means that a filter has been moved, added or deleted without being linked to a particular URL. This initial does not need to be followed by a URL.
--> 
```

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,6 @@
+<!--
+https://forums.lanik.us/viewtopic.php?p=131958#p131958
+A: (as Addition) : means that a filter to block an advertisement has been added
+P: (as Problem) : means that a problem has been fixed, for example by adding an exception in the case of a video that does not start, or in the case of an anti-adblock.
+M: (as Moved or Modified) : means that a filter has been moved, added or deleted without being linked to a particular URL. This initial does not need to be followed by a URL.
+--> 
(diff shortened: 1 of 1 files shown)


peace2000:
Good idea!

krystian3w:
But ignored

spirillen:
Hey @krystian3w 

May I suggest you add a short description on what ech file covers within the `easylist*/` folders, could be a super "remeberroll" when you are not a fulltime easylist reader/constributer

krystian3w:
Honestly, they won't want such PR at all then.

Besides, I don't always find myself in which file to paste something because of collecting some filter groups into a small "regional" file.

Przechwytywanie |
--- |

spirillen:
Yeah it's a shame they have turned down the suggestion of a `README.md` file in the past :cry: but let's see if they will like it this time, so everybody who think this is a good idea, give a feedback, comment or emojis....

peace2000:
@Khrin Why not to add these explanations? New users users who are not aware of these would definetely benefit.

PR #2408, opened 2018-12-07
A:u-torrent.eu

mediumkreation:
Found in URL: http://www.u-torrent.eu/movies/baahubali-2
u-torrent eu_18-12-07 12 44 27

Found in URL: http://www.u-torrent.eu/search.php?query=venom
u-torrent eu_18-12-07 12 48 21


peace2000:
Conflicting now.

peace2000:
Relevant, these rules hide bogus download buttons.

peace2000:
Lol the whole url itself seems to be just a fake site.

http://www.u-torrent.eu/movies/baahubali-2

kuva

No movie can be 190MB in size... Also the site uses the name of a known Torrenz-service.

PR #2412, opened 2018-12-07
A:electronicshub.org

mediumkreation:
Found in URL: https://www.electronicshub.org/
electronicshub org_18-12-07 15 08 23

Found in URL: https://www.electronicshub.org/best-capacitor-kits/
electronicshub org_18-12-07 15 08 54
electronicshub org_18-12-07 15 09 02


peace2000:
This should be closed. Some of those rules don't block/hide anything anymore and one rule is meant to block those capacitor kit Amazon links but they should not be blocked because the point of that subpage is to list them. I don't consider it an ad.

peace2000:
New, already accepted version has been accepted.

PR #2433, opened 2018-12-09
A:artstation.com

mediumkreation:
Found in URL: https://www.artstation.com/
artstation com_18-12-09 13 51 36


peace2000:
There is no element `##support-artstation-block` anymore.

@Khrin

peace2000:
There is no element `##support-artstation-block` anymore, this should be closed.

PR #2435, opened 2018-12-09
A:winprizesonline.in

mediumkreation:
Found in URL: http://winprizesonline.in/
winprizesonline in_18-12-09 14 33 28


peace2000:
Problem still relevant. This does not fix it because even though it blocks popups it also makes it impossible to access different sections on the site.

peace2000:
Relevant. Just click "Submit" and ad popup will open.

PR #2444, opened 2018-12-11
P:linkpizza.com

arjanpronk-kizitos:
On a logged-in admin page, users can filter by advertiser name, the exclusion added for ?advertiser= allows ajax requests to go through again.


ryanbr:
A whitelist would be better, submit testable examples in the Easylist forums.

PR #2447, opened 2018-12-11
A:idiva.com

mediumkreation:
Found in URl: https://www.idiva.com/news-entertainment/after-sacred-games-netflix-announces-three-more-indian-original-shows/17074716
idiva com_18-12-11 19 43 31
idiva com_18-12-11 19 43 40


peace2000:
Conflicts. Cannot access from Europe, not willing to mess with proxies.

peace2000:
Did not see ads. (Had to use a proxy for that site, hope it didn't affect the result).

PR #2461, opened 2018-12-13
A:mrunal.org

mediumkreation:
Found in URL: https://mrunal.org/
mrunal org_18-12-13 18 48 19


peace2000:
Conflickting now. Also rule is old now. I would use:

`||mrunal.org/*ad`

peace2000:
`||mrunal.org^*/ad-$image` - Doesn't work.

I would use:

`||mrunal.org/*ad`

It blocks many ad images.

kuva

(Not all the images fit to one picture.)

PR #2462, opened 2018-12-13
A:slideplayer.com

mediumkreation:
Found in URL: http://slideplayer.com/slide/2790499/
**Issue Description**: the Ads change position based on width of the browser (responsive)

slideplayer com_18-12-13 19 17 54


Khrin:
I don't see any unblocked any adverts there and "Addthis" isn't an Adserver.

PR #2481, opened 2018-12-16
A: (radiokorea.com) easylist/KoreanList#1

smed79:


Khrin:
Sorry for having missed it before... And I believe that I could make it as generic filter.

PR #2559, opened 2018-12-30
A: http://url4btc.com/82XQ4r

smed79:


Khrin:
Why removing that filter?

PR #2575, opened 2019-01-01
A: (mi.com / xiaomi.com) MIUI OS tracking domains

smed79:


Khrin:
If you don't mind, I've modified a little bit your proposed filters...

smed79:
Thank you.

PR #2635, opened 2019-01-10
A: gamecopyworld.com & gamecopyworld.eu

smed79:


smed79:
@Khrin 

**bad timing**, I did not touch yavli filter. 
new pach here #2648

PR #2649, opened 2019-01-12
A: http://www.psx-place.com/forums/

smed79:
```
http://www.psx-place.com/js/ipbforo/js/jquery.totop.js
http://www.psx-place.com/js/ipbforo/js/jquery.min.totop.js
```

### DIFF
--- a/fanboy-addon/fanboy_annoyance_general_block.txt
+++ b/fanboy-addon/fanboy_annoyance_general_block.txt
@@ -103,6 +103,7 @@
 /iphoneapp-icon.
 /itpsocialbuttons/*
 /jquery.androidbanner.js
+/jquery.totop.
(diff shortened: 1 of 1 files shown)


peace2000:
Still relevant, `/jquery.totop.` is being used on many webpages: https://publicwww.com/websites/%22%2Fjquery.totop.min.js%22/

peace2000:
Still relevant, `/jquery.totop.` is being used on many webpages: https://publicwww.com/websites/%22%2Fjquery.totop.%22/ so a general block rule is justifiable.

Khrin:
A similar filter was removed due to excessive false positives and honestly I believe such generic filter could have the same behaviour.

peace2000:
Oh ok. But site specific filter for this site would do?

Khrin:
I'm not sure if it's worth the effort.

PR #2680, opened 2019-01-20
A: (nsfw) xxxbunker.com

smed79:


smed79:
[#4683](https://github.com/uBlockOrigin/uAssets/issues/4683)

PR #2699, opened 2019-01-24
A: https://www.satoshinakamotoblog.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42504


peace2000:
_1-6-0x6-0-0. doesn't bring any hits in my logger.

smed79:
@Khrin Still a good generic dimension.

PR #2709, opened 2019-01-24
A: https://imgur.com/

smed79:


smed79:
[uAssets#4704](https://github.com/uBlockOrigin/uAssets/issues/4704)

PR #2727, opened 2019-01-26
A: http://crabcut.net/WeRsSt35

smed79:


smed79:
https://github.com/uBlockOrigin/uAssets/issues/4717

PR #2736, opened 2019-01-27
A: #2735

smed79:


Khrin:
These aren't tracking services. https://en.wikipedia.org/wiki/WURFL

PR #2797, opened 2019-02-02
A: https://vip7stream.pw/bc0069a07bf78daa34378fca6f0f3f59-live.html

smed79:
https://i.imgur.com/br8gKr4.png
https://forums.lanik.us/viewtopic.php?p=145348#p145348

### DIFF
--- a/easylist/easylist_adservers_popup.txt
+++ b/easylist/easylist_adservers_popup.txt
@@ -59,6 +59,7 @@
 ||ad4game.com^$popup,third-party
 ||ad6media.fr^$popup,third-party
 ||adbetclickin.pink^$popup
+||adbetnet.com^$popup,third-party
 ||adbma.com^$popup,third-party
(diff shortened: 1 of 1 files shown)


peace2000:
https://vip7stream.pw/ cannot be accessed.

peace2000:
`adbetnet.com` is blocked already.

smed79:
> adbetnet.com is blocked already.

no the popup !!

peace2000:
||adbetnet.com^$third-party

Doesn't that block popups as well?

Khrin:
@peace2000 Not for ABP.

peace2000:
Oh, well that´s a flaw then.

Khrin:
I don't think that should be intended as a flaw, but as a way to differentiate newly opened pages (`$popup`) and regular pages (`$document`), so a different approach as that used by Ublock Origin.

PR #2799, opened 2019-02-03
P: https://xdrive.cc/embed/aM0jvuQXGw5ookg

smed79:
https://forums.lanik.us/viewtopic.php?p=145361#p145361


smed79:
@Khrin

peace2000:
Cannot connect to https://xdrive.cc , if this status quo persists longer time, probably https://xdrive.cc related entries should be removed.

PR #2804, opened 2019-02-03
A: https://www.mylink.zone/71JG

smed79:


peace2000:
Only this one `||max-adserv.com^$third-party` was shown on the hits in my logger.

PR #2805, opened 2019-02-03
A: #2804

smed79:


peace2000:
Related: https://github.com/easylist/easylist/pull/2804

PR #2814, opened 2019-02-05
A: https://www.lesechos.fr/

smed79:


peace2000:
Where `assets.lesechos.com/common/js/xtcore.js` is shown in the logger? That link is available but cannot see it in my logger.

This link however https://investir.lesechos.fr/index.php shows this in my logger:
https://investir.lesechos.fr/xtcore.js

those are a bit different though.

PR #2815, opened 2019-02-05
A: http://streamcloud.eu/tu6hqbygtf44

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42573


peace2000:
`streamcloud.eu` is not available, so this req is not relevant anymore.

PR #2816, opened 2019-02-05
A: https://www.telechargerunevideo.com/finish?id=u39pqd26zr

smed79:


peace2000:
`www.telechargerunevideo.com` inaccessible. However, `eclkmpbn.com` is being used elsewhere: http://www.senimovie.co/

Hit in my logger:
`http://eclkmpbn.com/adServe/banners?tid=51737_78995_5`

PR #2821, opened 2019-02-06
A: https://www.btc-echo.de/

smed79:


peace2000:
Didn't see this element `##.adrotate-banner-top` in element picker anymore on that domain.

smed79:
Still a good generic hidding filter.

peace2000:
These two search engines: https://publicwww.com/websites/%22adrotate-banner-top%22+depth%3Aall/

https://nerdydata.com/reports/new?search=%7B%22all%22%3A%5B%7B%22type%22%3A%22code%22,%22value%22%3A%22.adrotate-banner-top%22%7D%5D,%22any%22%3A%5B%5D,%22none%22%3A%5B%5D%7D

Produce no hits for "adrotate-banner-top", that's why I wouldn't consider it a good generic.

PR #2826, opened 2019-02-08
A: https://streamp1ay.me/e7vqpgl5gy0j

smed79:
https://forums.lanik.us/viewtopic.php?p=145496#p145496


peace2000:
https://streamp1ay.me can't be connected.

PR #2828, opened 2019-02-08
P: (solarmoviez.ru) https://forums.lanik.us/viewtopic.php?f=64&t=42583

smed79:


peace2000:
`solarmoviez.ru` itself is not accessible, it redirects to `www1.solarmovie.to`, which has no issues at the moment.

PR #2831, opened 2019-02-09
A: https://mega-dvdrip.info/

smed79:
https://github.com/uBlockOrigin/uAssets/issues/2198#issuecomment-460095409


peace2000:
sisney.net is dead domain now.

PR #2832, opened 2019-02-09
A: http://mozdream.live/w9No1hx

smed79:
https://github.com/uBlockOrigin/uAssets/issues/4847


peace2000:
`||holexknw.loan^$third-party` already added.

PR #2835, opened 2019-02-10
A: https://www.file-up.org/y2lhcmug0bjl

smed79:


peace2000:
Third-party ad scripts that are being used on that domain are already blocked by Easylist. also third-party fonts.

I would also add blocking for popups:

`||www.file-up.org^$popup` as when I click the white space on that link, A popup opens where I'm being taken to a page where the logo of site is being displayed.

smed79:
https://git.io/Jv2Uo

PR #2840, opened 2019-02-11
P: https://arsenal.army/shop/individualnaya-zashchita

smed79:
https://forums.lanik.us/viewtopic.php?f=64&t=42607


peace2000:
Fanboy Annoyance list blocks product images. Smed's fix works.

Old problem but still relevant.

@ryanbr

smed79:
@Khrin

PR #2843, opened 2019-02-11
A: http://agar.pro/play/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42608


peace2000:
`agar.pro/play/assets/js/rightclick.js` this file is not there anymore.

peace2000:
The site cannot be accessed also now.

PR #2845, opened 2019-02-12
A: https://pctnew.com/

smed79:


peace2000:
`pctnew.com` is now `dontorrent.org`.

`iskkrmwivxfmdjwxme.bid` is alive:
kuva

But there are no connections to that domain on `dontorrent.org`. But anyway, it should still be blocked in my opinion as it clearly announces it's purpose.

PR #2855, opened 2019-02-13
A: http://gmarket.co.kr/

smed79:


peace2000:
||script.gmarket.co.kr/js/common/uuid.js
||script.gmarket.co.kr/js/header/StatSNA.js

Both still relevant, needs rebasing.

PR #2861, opened 2019-02-16
A: https://services.iamhansen.xyzpng

smed79:


peace2000:
Still relevant, needs rebasing. Also the title has typo, png should be removed from the end.

PR #2864, opened 2019-02-16
A: https://www.dailymail.co.uk/

smed79:


Khrin:
In the past, blocking these scripts have caused some false positives...

smed79:
Ok, thank you.

PR #2868, opened 2019-02-18
A: https://downloadhub.wiki/

smed79:


peace2000:
Still relevant request but it's `downloadhub.blog` now.

Here is a sample link where that ad can be found:
https://downloadhub.blog/it-happened-in-calcutta-2020-hindi-season-01-complete-720p-hdrip-x264/

PR #2871, opened 2019-02-18
A: https://www.t-online.de/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42634


peace2000:
`tagm.tchibo.de` already in the Easyprvicay list.

However, that domain is not in use anymore, it redirects to:

https://www.exactag.com/

Khrin:
I don't see it used elsewhere.

spirillen:
> `tagm.tchibo.de` already in the Easyprvicay list.
> 
> However, that domain is not in use anymore, it redirects to:
> 
> https://www.exactag.com/ (which is not blocked)

It is now https://www.mypdns.org/T627 :smiley: :grin:

PR #2875, opened 2019-02-18
A: https://www.heise.de/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42638


peace2000:
`||responder.wt.heise.de^` is blocked already by now.

PR #2881, opened 2019-02-19
Mixer.com follow button

mtcdood:
The channel follow button on streaming site Mixer.com was being hidden by this filter.


Khrin:
I haven't approved this pull request because your proposed change is invalid, anyway now is fixed.

PR #2893, opened 2019-02-22
A: http://worldlingnews.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42652


peace2000:
` ||idevaffiliate.com^$third-party`
and
` ||trafficroots.com^$third-party`

The first one redirects to` idevdirect.com` now.

The second one is the same.

PR #2921, opened 2019-02-27
Add api.markitondemand.com to whitelist

Far3:


Khrin:
Why such a filter should be added? Is this domain incorrectly blocked somewhere?

Far3:
Yes, I have documentation at the following forum post. It is blocking the serving of APIs on that domain. Further documentation below.

https://forums.lanik.us/viewtopic.php?f=64&t=42679

Khrin:
I have removed the filter, and added a more specific one, to block just the tracking service.

Far3:
Thank you very much @Khrin. When does this list get updated to https://easylist.to/easylist/easyprivacy.txt? On Saturday, March 2nd?

Khrin:
Should be already updated. Users should automatically get latest version in a few days, depends by the software.

PR #2946, opened 2019-03-04
A: https://www.fidor.de/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42694


peace2000:
`collect.igodigital.com` already there and it's 404.

PR #2955, opened 2019-03-04
P: (NSFW) xbooru.com

smed79:
https://forums.lanik.us/viewtopic.php?f=64&t=42676


peace2000:
Still relevant, needs rebasing.

Khrin:
Why should be needed?

peace2000:
I meant that this address: `xbooru.com/script/application.js` still works.

PR #2963, opened 2019-03-05
A: discordapp.com

ealgase:
This blocks the science endpoint of the Discord API which is used exclusively for tracking users.


Khrin:
(IIRC) We had a filter blocking this element, but has been removed due to some false positives.

PR #2967, opened 2019-03-06
Remove airswap.io from this list

rhyeal:
The URL for airswap.io is a widget for using the AirSwap platform. It must be explicitly added by a developer per https://developers.airswap.io/

It is not a popup or banner ad.

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -1521,7 +1521,6 @@
 ||prnewswire.com/rit.gif?
 ||prnewswire.com/rt.gif?
 ||proac.nationwide.com^
-||production.airswap.io^
 ||projop.dnsalias.com/intranet-crm-tracking/
 ||prontohome.com/permuto.do
 ||propertyfinder.ae/js/ga.js


smed79:
> It is not a popup or banner ad.

FYI, the issue here is about privacy.

rhyeal:
Can you help me understand what the issue is with privacy? I believe the only tracking is Google Analytics and Mixpanel, which are fairly common (and already blocked).

rhyeal:
I wanted to add some additional comments on this PR since it seems that it has not been addressed yet.

Taking a look at the [EasyPrivacy policy](https://easylist.to/pages/policy.html) and [acceptable first party tracking](https://easylist.to/2011/08/31/what-is-acceptable-first-party-tracking.html), I'm not able to understand why this line was added.

- "Evidence should be provided that filters need to be added to the subscription."
  - I do not see any evidence as to why this line was added in the first place
- "Filters for first-party trackers should only be added if the items blocked collect a significant amount of personal data."
  - **IP address and user agent**: This is not collected beyond what is inherently sent in each HTTP request
  - **Screen resolution, colour depth and browser plugins**: This information is already blocked on this domain via the blocking of Mixpanel and Google Analytics, and is not collected by other scripts.
  - **Time zone**: This information is already blocked on this domain via the blocking of Mixpanel and Google Analytics, and is not collected by other scripts.
  - **Language**: This information is recorded in the browser for rendering purposes only
  - **JavaScript status**: This information is used to determine which message to render, a JavaScript required message or the React app
  - **Advert blocking status**: This information is not recorded
  - **Page status**: This information is acceptable to record
  - **"Random" numbers**: This is not used
  - **Object tracking (e.g. using Flash or Java to track users)**: Neither Flash nor Java is used.
- "Consistently problematic filters should be removed from the subscription."
  - There is documented evidence that this filter impedes third party websites from running their business: https://forums.lanik.us/viewtopic.php?f=64&t=42718&p=146190#p146190

peace2000:
`production.airswap.io` is 404. So now at least it should be removed.

rhyeal:
@peace2000 Any chance this will be pulled in to the master branch then?

PR #2969, opened 2019-03-07
A: https://www.mediafire.com/error.php?errno=386

smed79:


smed79:
https://github.com/uBlockOrigin/uAssets/issues/4826

PR #2971, opened 2019-03-07
A: https://streamango.tv/episode/genius-season-1-episode-1/

smed79:


peace2000:
Not sure if that site is in proper use anymore, but the rule suggested `###ad_buts` would still do as a general rule, that element is being used on 47 websites:

https://publicwww.com/websites/%22%23%23%23ad_buts%22/

PR #2972, opened 2019-03-07
P: https://trancentral.tv/2018/06/trancentral-weekly-new-psytrance-releases-24-06-2018/

smed79:
https://forums.lanik.us/viewtopic.php?f=64&t=42724


peace2000:
Problem still relevant. Smed's fix works.

PR #2997, opened 2019-03-09
A: http://hulblog.com/

smed79:


peace2000:
I don't see `##.toplink[href="#top"]` anymore, looks redundant.

PR #3000, opened 2019-03-10
A: http://www.torcidatricolor.tv/bbb.php

smed79:
https://github.com/abp-filters/abp-filters-anti-cv/issues/93


peace2000:
www.torcidatricolor.tv is 404.

PR #3001, opened 2019-03-10
A: banners.itunes.apple.com

wjthieme:


peace2000:
Why is this rule needed?

PR #3006, opened 2019-03-10
A: http://torcidatricolor.tv/jogo8.html

smed79:
https://github.com/abp-filters/abp-filters-anti-cv/issues/93


peace2000:
http://torcidatricolor.tv/ does not respond at the moment.

PR #3015, opened 2019-03-11
A: (NSFW) sexu.com

smed79:


smed79:
https://github.com/uBlockOrigin/uAssets/issues/1161

peace2000:
premium.sexu.com dead domain
xxxity.com seems to be expired

Other domains alive.

PR #3037, opened 2019-03-13
P: https://www.washingtonpost.com/transportation/2019/03/12/where-plane-model-that-crashed-ethiopia-flies

smed79:
https://i.imgur.com/P30ggYg.png
https://forums.lanik.us/viewtopic.php?f=64&t=42745


smed79:
@ryanbr @monzta

PR #3057, opened 2019-03-14
A: https://thevid.tv/v/y3jw4dezcko4t

smed79:
https://github.com/abp-filters/abp-filters-anti-cv/issues/110
https://forums.lanik.us/viewtopic.php?f=62&t=36462


peace2000:
I don't see any requests to /style/medio/ so this is redundant now. I'm not sure if one needs to logged in. Link in the title is 404 now, domain is not.

peace2000:
thevid.tv is 403 Forbidden

PR #3061, opened 2019-03-14
A: https://www.narcity.com/

smed79:


peace2000:
brealtime.com is already in the list.
emxdgt.com already in the list on this form: hb.emxdgt.com

emxdgt.com redirects to emxdigital.com (not in the list)

emxdigital.com is clearly an ad server, judged by their front page.

PR #3065, opened 2019-03-15
A: http://short.pe/DfAiqE

smed79:


DandelionSprout:
I approve of this simplification of generic entries, despite me being unable to reproduce the entries on the exact example page.

PR #3067, opened 2019-03-15
A: https://www.sol.de/

smed79:


smed79:
https://github.com/uBlockOrigin/uAssets/issues/5129

peace2000:
Blocks `/dsgvo-consent.` , general blocking rule. Easylist has hiding rule for it already (`##.dsgvo-container`) but no blocking rule.

PR #3068, opened 2019-03-15
A: #3067

smed79:


peace2000:
`###adn_layerad` not found anymore on `www.sol.de`.

smed79:
@Khrin Not found on `sol.de` does not mean anything and does not justify the close of PR.
The filter still a good generic https://publicwww.com/websites/adn_layerad/

PR #3069, opened 2019-03-15
Update easylist_thirdparty.txt

PentagramTr:
`Bugun` and `Zaman `newspapers were closed.Also `e-kolay.net` was closed.

This rule already exists in the Turkish filter.
`||ad.reklamport.com^$third-party `

https://github.com/AdguardTeam/AdguardFilters/blob/ad31a58f7a5fe741a6fdb99a775f636125ae88ca/TurkishFilter/sections/adservers.txt#L63


Khrin:
Hi, I haven't accepted this PR because I prefer to keep in the list that filter, as (I think) was added from an English-language site and likely is still active. I have removed the other filters you've suggested. Thank you!

PR #3076, opened 2019-03-16
A: https://zorka.mobi/en

smed79:


peace2000:
"fanboy-addon/fanboy_social_general_hide.txt"

Shouldn't this be added to Easylist Cookie list instead?

peace2000:
Not in use on that domain anymore, but used on this: https://cerealia.ch/ I'll make a new pull req.

PR #3092, opened 2019-03-17
A: (algeriatenders.com) #3091

smed79:
```
http://www.algeriatenders.com/publicite.xhr.php
```


ryanbr:
Probably not worth it, even for a generic.

PR #3094, opened 2019-03-18
P: https://olymptrade.com/

smed79:


smed79:
@ryanbr

peace2000:
This rule `###to_top` breaks this site completely. This is important fix therefore as other sites could be affected as well.

smed79:
https://github.com/easylist/easylist/blob/c607113b121f8f7d5b239b8e2126887ef7638b7e/fanboy-addon/fanboy_annoyance_general_hide.txt#L2835

PR #3120, opened 2019-03-22
M: (ebay.com) Update 22fd4a9

smed79:
```
https://www.ebay.de/scl/js/ScandalLoader.js
https://www.ebay.es/scl/js/ScandalLoader.js
https://www.ebay.fr/scl/js/ScandalLoader.js
https://www.ebay.it/scl/js/ScandalLoader.js
https://www.ebay.co.uk/scl/js/ScandalLoader.js
```


Khrin:
it doesn't seems to be necessary, for now.

Issue #3188, opened 2019-04-04
https://en.wiktionary.org/wiki/ads


mk-pmb:
### List the website(s) you're having issues:

```text
https://en.wiktionary.org/wiki/ads
https://en.wiktionary.org/wiki/advert
https://de.wiktionary.org/wiki/advert
```

### What happens?

The main content area of the dictionary seems missing.

Since the rule is `##.page-ads`, it likely affects lots of other wiki pages that happen to use a CSS class based on the page name.

### List Subscriptions you're using:

`easylist/easylist_general_hide.txt` affected since b34682f617ec02b3365aedd237452dfd9cc9da2d

### Your settings

skipped because the rule is obviously overly general.

### Other details:

Reporting on behalf of the affected user. [Original IRC log](https://wm-bot.wmflabs.org/logs/%23wikimedia-tech/20190404.txt)

RhinosF1:
Hi, I'm part of the team on IRC that reported it

https://help.getadblock.com/support/tickets/62765 was filed at first during investagations and is now closed.

https://en.wiktionary.org/w/index.php?title=Wiktionary:Information_desk/2019/April is also a discussion on it

mk-pmb:
Adjusted title because the other pages are blanked by similar but different rules. Unfortunately, Github blame seems to not work for me atm.

Khrin:
This will fix the issue.

RhinosF1:
@Khrin  Thanks!!!

RhinosF1:
How long for adblockers to start pushing this @Khrin ?

Khrin:
Change is already live. All adblockers should get the update in a few days.

RhinosF1:
Thanks, Fix live for me

PR #3190, opened 2019-04-04
Add official Wikimedia domains to whitelist

quiddity-wp:
Add to whitelist all canonical domains for the Wikimedia movement, cf. https://www.wikimedia.org/ 
Note: w.wiki is the forthcoming official URL shortener, cf. https://meta.wikimedia.org/wiki/Wikimedia_URL_Shortener
Sidenote: See also the recent addition for Wiktionary at https://github.com/easylist/easylist/issues/3188 (thanks!)

### DIFF
--- a/easylist/easylist_whitelist.txt
+++ b/easylist/easylist_whitelist.txt
@@ -1905,6 +1905,7 @@
 @@||vtstage.cbsinteractive.com/plugins/*_adplugin.swf
 @@||vzoozv.com^$domain=nme.com|trustedreviews.com
 @@||w.org/plugins/adsense-plugin/screenshot-$image,domain=wordpress.org
+@@||w.wiki^$generichide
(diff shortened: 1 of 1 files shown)


RhinosF1:
@khrin - Seen as you handled #3188, would it be possible you could push this?

ryanbr:
This seems a bit of an overkill to whitelist so many sites.

RhinosF1:
@ryanbr - Given that WikiMedia sites dont show adds and #3188 took out multiple pages on one of the sites for 10 hours while it was fixed. I don't think whitelisting them is overkill.

Khrin:
Personally I do agree that's a overkill at the moment, unless there's a specific issue for every site. I'm willing only to add a whitelisting rule for Wikipedia, since we already have one in the Annoyances List, but not for the other sites.

PR #3211, opened 2019-04-09
whitelist ipb v4 forum's buttons for roads.ru

Bars-ik:
after forum upgrade to v4 some useful buttons became blocked. This commit unblock them.


Khrin:
I don't think this is necessary. Could be also easily abused.

PR #3247, opened 2019-04-11
ibit.to wrong blocking fix https://forums.lanik.us/viewtopic.php?f=64&t=42876

Ibit-to:
Allow google fonts on https://ibit.to/ ref: https://forums.lanik.us/viewtopic.php?f=64&t=42876


peace2000:
This seems to be fixed already (mentioned in the forum link also), and when I access that domain, no fonts are blocked.

PR #3350, opened 2019-04-28
Fix network panel for GADWP plugin

chriscct7:
Whitelist already fixes the admin panel, but network panel is blocked


chriscct7:
As a sidequestion, there needs to be another whitelist for font files (woff2, woff, ttf, oof).

How does one write a rule for that?
If someone can show me I can do another pr to fix those

PR #3355, opened 2019-04-30
Remove datadome.co

gilleswalbrou:
Hi, I'm Gilles, Head of Engineering at DataDome.
datadome.co is a cybersecurity solution to protect websites against the bad bot activity. We don't use information for any other purpose than detecting fraud and bad bots.
Blocking our JS may generate blocking access to some of the major french websites (PagesJaunes, LeParisien, Blablacar...). more information on detection [here](https://datadome.co/how-to-detect-malicious-bots/)
This is something already discussed a few months ago on those PRs : #673 and #997

This PR is a proposal to revert a [commit](https://github.com/easylist/easylist/commit/6843b5ff0d5733052dce19c1a70c42e2b7947fef#diff-1a58c0a4c8e32e93aa7240cd96af3279) delivered a few days ago.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_international.txt
+++ b/easyprivacy/easyprivacy_trackingservers_international.txt
@@ -343,7 +343,6 @@
 ||compteur.org^$third-party
 ||count.fr^$third-party
 ||countus.fr^$third-party
-||datadome.co^$third-party
 ||deliv.lexpress.fr^$third-party
 ||do09.net^$third-party
 ||edt02.net^$third-party


ryanbr:
Bot's won't use a Adblock Extenstion, Adblock List and Browser, this list is used by actual people. And the fact you're hiding behind obfuscated code, shows its untrustworthy.

gilleswalbrou:
Basic bots are not using real browser, Advance Bots are using browser automation like PhantomJS or Chrome headless for instance, this is the reason why we need to challenge the browser to detect all bot generations.
For the same reason, we obfuscate our JavaScript code because we are fighting hackers that use bots for account takeover and try to reverse engineer our challenges.
Because Easylist users won't trigger our Javascript, they may be detected as basic bots and be blocked on many websites.
We really want to avoid degrading Easylist user experience.
As mentioned, we don't use information for any other purpose than detecting bad bots and don't share any information to thrid party providers. This is part of our customers agreements.

paullaleu:
Hi, I'm Paul, DevOps for one of the french media company using Datadome to protect our infrastructure from bad bots and hackers. (our website is https://www.nouvelobs.com) 

We have feedbacks from easylist users (using ublock origin for exemple) that are blocked because they wont be prompted by the captcha page Datadome uses to protect our website when it detects a bad bot connexion. 

Datadome use an algorithm to detect if a connexion is a bot or not, some times it can be wrong and it is very important to us that any actual human can validate the captcha on Datadome page to access our website wether or not they use an AdBlocker. It acts like a firewall, that any company use to protect their servers.
I can reaffirm here that our intention in using Datadome is not at all to block any AdBlock users, and Datadome is not here to collect any data about users, only to filter bad bot(crawler, ad companies, data intelligence companies, etc)  traffic before it reach our website.

In fact Datadome even helps with preventing such companies to read our users comments and collecting their data without our/their consent, which is precisely the same fight Adblockers are made for, so you should find a common interest ;)

dgifork:
Hello, I am security manager for LaFourchette.
We use datadome to block bots.
We have complaints from customers because they cant access to our websites because Adblockers dont allow to show captcha to prove they are not bots.
https://twitter.com/pg_xocolatl/status/1128326224114323456

ryanbr:
The script is untrustworthy.

SachaMorard:
Hello @ryanbr 
I'm Sacha Morard, CTO of Le Monde Group. I understand your point of view, it should be more simple if the datadome JS wouldn't be obfuscated. But as a security solution, they have to fight about hackers who steal our user's data and journalist's contents, so they have to be "under cover" ;). I assure you, it is a true security solution that protects our readers.
Our subscribers pay to access our premium items, and we significantly reduce advertising pressure for them, but since this blocking, they cannot read our premium articles and they don't understand why.
Could you review your opinion?

ryanbr:
As a response/update.

Bot's don't use adblock lists. So no matter what is added here, won't affect bots. It wouldn't stop bots from manually adding the DataDome domain regardless (if they so wish). 

Inspecting `https://js.datadome.co/tags.js` , you'll see UA, referrer, event checks, mouse movement checks. I understand datadome is trying to do something 'good', but it's still tracking. Easyprivacy primary reason to block all tracking, limit many 3rd-party unnecessary scripts.

spirillen:
> Hello, I am security manager for LaFourchette.
> We use datadome to block bots.
> We have complaints from customers because they cant access to our websites because Adblockers dont allow to show captcha to prove they are not bots.
> https://twitter.com/pg_xocolatl/status/1128326224114323456

Then your code is broken. And needs to be rewritten into a working state.

As @ryanbr already has replied numerous times, this program is used by humans not bots. S in reality you might actually do a check for if users is using Ublock or similar programs to protect them self against Cyber Criminals like your self trying to harvest everything about tour users, by therefore actually deliverers a spyware site.

And how would you like users to get any of your contents if they uses www.mypdns.org privacy enhanced DNS servers? The reality, as I understand you replies, is that user using uBlock or DNS Firewalls isn't you audience in the first place, as they requires higher standards.

If you are from France, this should be so hard to think European(ish) and not like that Google|facebook|Trumpsk, that's the US way, and no others.

Repeating your self, and (trolling) this as turning chose we more than to simultaneous working brain cells against you.

Have I misunderstood something?

gilleswalbrou:
@ryanbr As an ad-block user, I understand your position: you have to draw a "line" somewhere to determine what's trustworthy and what's not.

Bad bot threats are going through the roof, especially credential stuffing where bots are trying to steal personal information. This is a very serious security issue today.
Most of the largest websites are deploying bot protection to keep their users safe.

On the detection side, the JS tag is one of the components used to verify the authenticity of the request. The lack of data from the JS tag can be explained by two reasons:
- It’s a robot 
- It’s an EasyList user

This can generate an increase in the blocking decision for EasyList users.
And it’s the situation we would like to avoid.

We don’t track users, we just want to detect human vs robot.

Although we clearly understand the need to block any unnecessary third party tracker, we would like to raise the question if protecting users’ credentials is not a separate category?

Of course, if you have any idea on how we could collaborate to avoid this, we’ll be glad to discuss this more in detail.

vincentbernat:
After upgrading to Firefox 85, I have been running a lot in captcha by datadome. I have whitelisted it specifically with:

```
! 2021-02-07 https://geo.captcha-delivery.com
@@||js.datadome.co/tags.js$script,domain=geo.captcha-delivery.com
@@||api-js.datadome.co^$domain=geo.captcha-delivery.com
```

However, I am still getting the captcha. So, removing datadome seems to not be enough to fix that and I suppose they rely on third-party cookies that are now blocked by Firefox.

PR #3511, opened 2019-05-27
A: http://www.bombuj.eu/prehravace/verystream.com.php?url=john-wick-3

smed79:
```
https://pastebin.com/raw/jt3hRRmn

http://www.bombuj.eu/prehravace/verystream.com.php?url=john-wick-3&version=11&code=SNRW3FqlEvM&id=YuVuiXdBcz2&vip=
```

PR #3512, opened 2019-05-27
A: #3511

smed79:
```
http://www.bombuj.eu/prehravace/verystream.com.php?url=john-wick-3&version=11&code=SNRW3FqlEvM&id=YuVuiXdBcz2&vip=
```


ryanbr:
domain is dead

spirillen:
`||prehraj.to^*/banner` isn't dead...

PR #3522, opened 2019-05-29
A: https://www.avito.ma/

smed79:


Khrin:
I don't think `schibsted.com` is a tracker.

spirillen:
> I don't think `schibsted.com` is a tracker.

Funny I have the same :question: [here](https://gitlab.com/my-privacy-dns/matrix/matrix/issues/1412) :smiley:

PR #3543, opened 2019-05-31
A: https://www.hunterboots.com/us/en_us/

smed79:


ryanbr:
Used for text auto-complete. seems okay?

smed79:
I will leave the judgment to you.

PR #3599, opened 2019-06-10
A: https://davidwalsh.name/

smed79:


Khrin:
Proposed filter was good, but is blocking more unwanted stuff if less specific.

PR #3664, opened 2019-06-23
P: https://www.gartenjournal.net/

smed79:


ryanbr:
No longer needed.

`##.show-cookie-popup:not(body):not(html)`

PR #3710, opened 2019-07-03
R: bstats.org (Metrics for MINECRAFT SERVER, OPEN SOURCE, false positive)

antt0n:


ryanbr:
Keep in mind that filter has an exemption already. Is there an example where its broken?

antt0n:
bstats.org is a site that displays statistics retrieved from Minecraft plugins.
This data is retrieved with the consent of the users, and this project is Open Source.
Blocking this site prevents users who have your lists from seeing images on external sites (such as https://botsentry.eu/).

PR #3716, opened 2019-07-04
A: https://www.windy.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=64&t=43192


smed79:
@ryanbr @Khrin https://forums.lanik.us/viewtopic.php?p=149060#p149060

ryanbr:
Btw, P: used here.

PR #3723, opened 2019-07-05
Remove /api/v1/collector

omriiluz:
This is causing far reaching issues with hundreds of sites.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -707,7 +707,6 @@
 /api/pixel?
 /api/stat?
 /api/tracking/*
-/api/v1/collector
 /api/v1/metrics
 /api/v1/stat?
 /api/x.gif?


Khrin:
I have to guess that https://github.com/easylist/easylist/commit/701b01951de87a7494cbf45c91ca9497da289c69#commitcomment-34201247 and https://github.com/easylist/easylist/pull/3724 are related. 

Anyway, what a bad, I mean, bizarre choice for a "security service" name... What's the issue for real? The site is working perfectly.

omriiluz:
Because the system can't profile you as a user, you will be facing challenges a lot more often. Try to browse quickly through a site and you'll get the block page. 

This is the service - https://www.perimeterx.com/

omriiluz:
Here's an example on how it's being used and the open source server component handling the calls https://github.com/PerimeterX/mod_perimeterx/blob/master/examples/index.html

omriiluz:
One more datapoint - a similar case has been resolved by removing another vendor's domain from the list - https://github.com/easylist/easylist/pull/997

brahma-dev:
@omriiluz This pull request is futile. You should try adding the website to whitelist instead.

omriiluz:
@brahma-dev this rule is blocking the path on every website, not just one. Using a whitelist wouldn't work. Why not limit the rule to only one website?

brahma-dev:
@omriiluz  As I understand you want to unblock `//collector-$APP_ID.perimeterx.net/api/v1/collector/pxPixel.gif`, and whitelisting `perimeterx.net/api/v1/collector` is enough for that.

omriiluz:
@brahma-dev, I am *not* looking to unblock the perimeterx.net domains (to do that whitelist would work). I am looking to remove the too generic rule that is blocking /api/v1/collector on any website/domain. It is blocking the local installation of the product on websites.

For example, you are blocking:
https://www.zillow.com/HYx10rg3/xhr/api/v1/collector
https://www.hulu.com/OB5TwiRQ/xhr/api/v1/collector
and hundreds other.

brahma-dev:
@omriiluz Well, unfortunately that's how bad the internet has become that one has to use generic rules. It would be an impossible task if one had to write individual rules for every tracker on the internet. So, you seriously can't expect easylist to remove such generic rules just because your product happens to fall under one of them. That would quickly slip into demands for allowing tracking pixels, because somebody is using them for good purpose.

The easier thing would be to change the route in your product and get your customers to update it.

omriiluz:
@brahma-dev I completely understand and we're changing the path as soon as we possible. Just one last comment - this rule wasn't created as a generic rule, it was created to block our product on one domain (seekingalpha.com - https://github.com/easylist/easylist/commit/701b01951de87a7494cbf45c91ca9497da289c69) and hence shouldn't have been placed in the first place.

PR #3724, opened 2019-07-05
PerimeterX isn't a tracking or ad service

safruti:
PerimeterX is a security service, protecting websites from automated attackers. 
information isn't sold or shared with third parties and is used solely in order to protect the web sites.
We suggest removing the perimeterx.net domain from the list.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1615,7 +1615,6 @@
 ||performanceanalyser.net^$third-party
 ||performancerevenues.com^$third-party
 ||performtracking.com^$third-party
-||perimeterx.net^$third-party
 ||perion.com^$third-party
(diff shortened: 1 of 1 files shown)


omriiluz:
@brahma-dev this is the pull request to remove perimeterx.net it self. Happy to change it to a whitelist rule if you think it's more appropriate.

omriiluz:
@brahma-dev wanted to surface this again, would it be possible to get this domain removed? We are happy to provide every guarantee you need that we are not a tracking or advertisement service

DRSDavidSoft:
@omriiluz Your service is major pain in the butt for bot developers such as me. Please, kindly use other methods to label user data than 3<sup>rd</sup> party collectors. :(

Not everybody has whitelisted `*.perimeterx.net` domain, and not everyone has javascript enabled.

PR #3728, opened 2019-07-07
Add perimeterx.net to whitelist

safruti:
PerimeterX is a security service protecting sites from malicious automated attacks/bots, and inspecting malicious scripts (see www.perimeterx.com).
the scripts served from the perimeterx.net domain and information sent to that domain is used solely for security and fraud detection purposes, and isn't shared with any 3rd party. 
by blocking the scripts or preventing the data collection it will impact hundreds of large sites using the perimeterx services, and may block users from accessing the sites.

adding the domain to the whitelist per @brahma-dev recommendation, and following issue https://github.com/easylist/easylist/pull/3724

### DIFF
--- a/easyprivacy/easyprivacy_whitelist.txt
+++ b/easyprivacy/easyprivacy_whitelist.txt
@@ -558,6 +558,7 @@
 @@||paypalobjects.com^*/opinionLab.js$domain=paypal.com
 @@||paypalobjects.com^*/pixel.gif$domain=youngcons.com
 @@||pbskids.org/js/ga-current.js
+@@||perimeterx.net^
 @@||petametrics.com^*.js?$script,domain=nylon.com|space.com
 @@||pillsbury.com/Shared/StarterKit/Javascript/ntpagetag.js
 @@||pillsbury.com/Shared/StarterKit/Javascript/UnicaTag.js


Khrin:
Don't even think about it... This request is absolutely unacceptable.

safruti:
can you please explain? 
I read the policy, the collected data isn't intended for tracking or monetization of users. blocking will impact users. 
I've seen other domains and urls added to whitelist and removed from the privacy lists on similar requests. It will help if you can give a better answer, as currently this is impacting hundreds of sites.

DRSDavidSoft:
@safruti While I agree with the fact that it may cause the PerimeterX to fail, personally I'm against the idea of whitelisting these types of domains.

It doesn't matter what their privacy policy says, _at all_. It's a fact that hundreds of users are generating these data (which contains user-identifiable data such as screen size/resolution) that are shared with PerimeterX, that they use to refine their blocks -- and may inadvertently be used to track the users.

Personally, I believe companies such as PerimeterX should keep their ideas to themselves, and use other means of rating to label user-generated data.

_Disclaimer_ This is just a third-party user's opinion; I am not in association with neither EasyList or PerimeterX.

PR #3744, opened 2019-07-09
Remove Shariff from fanboy_social_general_block.txt

Luensche:
Shariff is a data privacy protected sharing possibility. It seems wrong to block it.

### DIFF
--- a/fanboy-addon/fanboy_social_general_block.txt
+++ b/fanboy-addon/fanboy_social_general_block.txt
@@ -1954,8 +1954,6 @@
 /sharetw_
 /shareTwitter.
 /shareWidgetBackgrounds.
-/shariff-sharing/*
-/shariff.min.js
 /sharing-buttons.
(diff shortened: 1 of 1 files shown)


ryanbr:
Looking at the original commit; https://github.com/easylist/easylist/commit/f6136a921c48

Shariff is used for share buttons (can be tested by just whitelisting #@#.shariff-button on that sample site. It seems fine to leave it.

Luensche:
Indeed it is used for share buttons. But as I already wrote, I think it is wrong to block sharing possibilities which are privacy protected. If you look for the http requests, you will find no traffic to heise/ct (developer of shariff) neither to Facebook, Twitter and so on.

ryanbr:
That is the whole point of the social list, it'll hide all the social buttons. The privacy aspect would be secondary, and any privacy blocking is an Easyprivacy specific solution.

PR #3745, opened 2019-07-09
A: https://www.xrel.to/

smed79:


kbirger:
What is the rationale for blocking jsnlog? It seems to be a legitimate framework for logging errors from client to server side. I have observed negative consequences on pages legitimately using this.

I'm all for blocking trackers, but we can't simply just block any call that can be used to send information back to a server.

AkosLukacs:
We use [jsnlog ](http://jsnlog.com/)in production: In case any client-side error happens, an error report is sent back to the server, so we know it.
We have chosen jsnlog because it sends error reports back to our application, not to some third party.
It has caught and helped fixing a couple of errors for users with obscure browsers, unexpected circumstances, and deployment errors.

It is deliberately kept as a single file (original file name), not bundled, so any bundling / minification errors don't affect it, and we get error reporting back.

Blocking that file just broke our production site.

PR #3786, opened 2019-07-14
A: (discordbots.org) #3780

smed79:

PR #3780, opened 2019-07-14
A: https://discordbots.org/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=43272

PR #3804, opened 2019-07-19
R: (https://discordbots.org/)

Oliyy:
This is because of #3780 and #3786


Khrin:
Oh no, why do you want to do that? I'm becoming quite attached to this filter.

Oliyy:
This filter doesn't even work @Khrin did you even look at the linked issues?

PR #3839, opened 2019-07-28
A: nzz.ch adblock notice

matbech:
This removes the adblock notice on nzz.ch. Screenshot of notice: https://ibb.co/9pxpG5w


Khrin:
EasyList doesn't block anti-adblock systems.

PR #3859, opened 2019-07-30
R: fast.appcues.com

seanedwards:
Posted a thread about this on the forums here, happy to discuss in either place. https://forums.lanik.us/viewtopic.php?f=64&t=43331

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -901,7 +901,6 @@
 ||facebook.com^*/placementbid.json$third-party
 ||facebook.com^*/tracking.js$third-party
 ||fairfax.com.au/js/track/
-||fast.appcues.com^
 ||fastcounter.bcentral.com^
 ||fastcounter.onlinehoster.net^
 ||fastly.net/collect?


Khrin:
Script code looks like very suspicious indeed.

seanedwards:
Hey @Khrin, can you be more specific about what's suspicious about it? I'm on the platform team, and I directly work on the backend APIs and databases that power this SDK, so I can definitely answer anything you'd want to know about how Appcues works.

If you'd like to see what the script does, you can take a look at https://projectcu.es - this is a fake project management app that we built for sales, to show demos of various patterns we support. (Obviously you will need to disable ad blockers for it to work property :) ) You can use the "Examples" menu on the left to launch different content, and you can use `Appcues.debug()` in the js console to get a window that will give you more information about what the SDK is doing. Like I said in my forum post, product managers use our tool to show welcome tours and feature announcements to their users.

For whatever it's worth, I am also a uBlock Origin user, and I also use the EasyPrivacy lists. I appreciate the work you all are doing, and would like to be an advocate for the privacy community within Appcues, and help the company keep a positive relationship with ad blockers.

PR #3888, opened 2019-08-06
A: https://verystream.com/e/abhYfpBg3cf/

smed79:
https://github.com/abp-filters/abp-filters-anti-cv/issues/152


ryanbr:
Would this cause false positives. https://github.com/easylist/easylist/commit/f93a6f60ee0db8fe8076ec7cd53030a7d43efda6

Would be better.

PR #3890, opened 2019-08-06
R: https://www.bradyunited.org/

lebowitz:
This blocks embedded donation forms for many charities

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -370,7 +370,6 @@
 /acfp.js
 /acounter.php?
 /act_pagetrack.
-/actiontag/*
 /activetrackphp.php?
 /activities/logger/*
 /activity-track/?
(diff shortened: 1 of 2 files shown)


Khrin:
Please can you be more specific about what's supposedly broken? The "Donate" page seems to be loading correctly.

lebowitz:
Was unbroken by uBlock for us https://forums.lanik.us/viewtopic.php?f=62&t=43350

Khrin:
That's should be an answer?

itsdrewmiller:
@Khrin the at.js file is being blocked on https://www.bradyunited.org/donate (and many other sites) which generates an HTML form to fill out with contact info and (appropriate iframed) credit card info.  In addition to that specific file, there are other assets on the cloudfront domain such as client-uploaded images that are being erroneously blocked.  ublock origin has reverted the change for that blocker which is great, but we would like this to be fixed for adblock and others who use easylist.

Khrin:
Thank you.

itsdrewmiller:
Thanks!

Issue #3912, opened 2019-08-11
Wera.de - Blocked pages (EasyList) [Issue #3788 reopen]


THEtomaso:
The issue hasn't been fixed!

Details here:
https://github.com/easylist/easylist/issues/3788#issuecomment-520211267

zoreu:
Got ublock right?
click update now
![ublock origin](https://i.imgur.com/VJooPnj.jpg)

THEtomaso:
You're missing the point, zoreu.
Take a look at ryanbr's commit, and read my post (which I refered to above).

zoreu:
I do not understand, should work

https://github.com/uBlockOrigin/uAssets/commit/eea75769dc2516c07629279a3dce56af59537090

https://imgur.com/3JdiFYQ

THEtomaso:
> `https://imgur.com/3JdiFYQ`

That exception rule was added to '**uBlock Filters**'.
The issue still needs to be addressed at the source of the problem, which is in '**EasyList**'!
Currently, only people that use '**uBlock Filters**' is able to take advantage of the fix.
The great majority of adblock users only use '**EasyList**'!
Hope this makes things clear.

zoreu:
PR: #3915

Issue #3788, opened 2019-07-16
Wera.de - Blocked pages (EasyList)


THEtomaso:
### Affected site:
`https://www.wera.de/`

### Issue:
Some pages are prevented from loading!

### Examples:
`https://products.wera.de/en/screwdrivers_kraftform_plus__series_300.html`

`https://products.wera.de/en/tools_for_the_aerospace_segment_screwdrivers_screwdrivers_kraftform_plus__series_300.html`

(All language versions of the same pages are affected)

### Screenshot:
Not needed, as the problem is obvious.

### Problem filter:
* EasyList

### Problem rule:
`_300.htm`

### Possible fix:
`@@_300.htm$domain=wera.de`

--

### System info:
* OS/version: Windows 8.1 Pro (x64)
* Browser/version: Pale Moon v28.6.0.1 (x64)
* Adblock Extension/version: uBlock Origin v1.16.4.11 (XUL)

zoreu:
add:

`@@||products.wera.de^$xmlhttprequest`

llacb47:
Stop telling maintainers to add existing filters. https://github.com/uBlockOrigin/uAssets/commit/eea75769dc2516c07629279a3dce56af59537090

THEtomaso:
@ryanbr:

Removing `/300.html$subdocument` doesn't help.
Look at my initial report..
This is the rule that is causing the problem!: `_300.htm`

Screenshot from uBO's logger:
![filterissue-wera](https://user-images.githubusercontent.com/9382674/62831823-21e55400-bc25-11e9-84e8-4eede782f4d8.jpg)

zoreu:
PR: #3915

krystian3w:
PR name should be: `P: wera.de` for better readability?

https://github.com/easylist/easylist/pull/2280

zoreu:
@krystian3w thanks for reminding me
Khrin solved the problem

https://github.com/easylist/easylist/commit/e6d73534bb1ed2030a14ba8a33b23bd9d3cbd1b1

THEtomaso:
@okiehsch:
> https://github.com/uBlockOrigin/uAssets/commit/1aad10970154d813df9e48f0e16388dfdb2b820d

I was wondering if you'd notice this, before I made you aware of it. :)
Anyway, guess you can delete it now.

PR #3915, opened 2019-08-12
wera.de #3788 #3912

zoreu:
fix loading


Khrin:
No need to copy such bad filters.

PR #3925, opened 2019-08-16
A: thevid.tv

zoreu:
add thevid.tv domain
Example: https://thevid.tv/v/d0cke6iel1t4bv0mpyxzn


zoreu:
@ryanbr check if it's right

PR #3957, opened 2019-09-04
A: https://pasteca.sh/Ht8qB7Q9

smed79:


smed79:
@ryanbr `pasteca.sh` could be added to popads filter.

PR #3992, opened 2019-09-10
A: na.op.gg

gh-bct:
Minified device fingerprinting tracking script

https://static.developer.riotgames.com/js/rgea.min.js


llacb47:
Do not merge this, way too broad and will probably have false positives. Why would you block the whole js directory?

gh-bct:
A different hostname and path is used for their other stuff. Analytics files are the only thing I have ever seen in that folder on numerous sites.

PR #4038, opened 2019-09-19
remove hscta from tracking servers list

scottrw93:
Could `hscta.net` please be removed from this list? This script will not track events without `hs-analytics.net` also on the page, which is correctly blocked.

PR adding this https://github.com/easylist/easylist/commit/e8f1ea5be7b490a155534b62cc8a34d9c18cf656

@ryanbr

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1080,7 +1080,6 @@
 ||hqhrt.com^$third-party
 ||hs-analytics.net^$third-party
 ||hsadspixel.net^$third-party
-||hscta.net^$third-party
 ||hubvisor.io^$third-party
 ||hudb.pl^$third-party
 ||humanclick.com^$third-party


ryanbr:
This looks like a tracker to me; https://js.hscta.net/cta/current.js

https://pastebin.com/VbuU2K7v

scottrw93:
It relies on hs-analytics.js to do any tracking, which is already correctly blocked. current.js pushes any tracking events to an array, window._hsq. This is polled by hs-analytics.js which sends the events, but only if hs-analytics.js isn’t blocked. Without hs-analytics it’s just a button builder

ryanbr:
There is just too much in the script that is too suspect. Can't in good faith accept this pull request.

> It relies on hs-analytics.js to do any tracking, which is already correctly blocked. current.js pushes any tracking events to an array, window._hsq. This is polled by hs-analytics.js which sends the events, but only if hs-analytics.js isn’t blocked. Without hs-analytics it’s just a button builder

PR #4043, opened 2019-09-20
A: https://gamesmagnet.net/efootball-pes-2020/

smed79:


ryanbr:
Don't think there is harm in steam widget? Re-open if you think otherwise

smed79:
I trust your judgement.

PR #4059, opened 2019-09-23
remove ||ekstat.com/js/ek$script filter

enverarslan:
this filter causes many problems, effects users' surfing experience. Because website main javascript files are serving from this resource.

example blocked resource urls are:

- https://ekstat.com/js/ek$i-combo.js
- https://ekstat.com/js/ek$i-defer.js

we must have very specific blocker filters if this resource contains ads.

affected website: https://eksisozluk.com


ryanbr:
there was delays on the site, and blocking this script also helped.

https://github.com/easylist/easylist/commit/6e93d94429f4

PR #4085, opened 2019-09-27
A: https://krebsonsecurity.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=43442


Khrin:
I've modified a bit the proposed filter, if you don't mind.

smed79:
no objection.

PR #4098, opened 2019-09-29
A: optinmonster scripts

pavelloz:
Optinmonster is extremely user-hostile set of scripts, blocking content, detecting adblocks, forcing people to give up their emails to see what they have already downloaded into their browser. In general it behaves more like malware than something you voluntairly download to your computer. 
 
I think its very beneficial for people to get rid off this if this is not what they have been looking for.

I noticed that they are trying very hard to not be detected, by buying more and more domains, ie:

```
||optinmonster.com^$third-party
||optmnstr.com^$third-party
||optmstr.com^$third-party
||optnmnstr.com^$third-party
||optnmstr.com^$third-party
```

But lets try to keep them in check, because what they do is pretty high on the scale of being completely useless and counterproductive on the web.

I suspect they will be buying `op*m?n?s?t?r?.com` domains now, since they reached peak on `opt*` so maybe a good regexp would be more appropriate. Unfortunately im pretty bad at regexp and im certainly not an expert on regexp performance, so this is the best i can suggest - https://regex101.com/r/8zUn2F/1

https://optinmonster.com/

Test site: https://contentdistribution.com/seo-content-writer/

Before:
image
 
Edit: I removed the "after" screenshot (for possibly copyright reasons).

### DIFF
--- a/fanboy-addon/fanboy_social_thirdparty.txt
+++ b/fanboy-addon/fanboy_social_thirdparty.txt
@@ -148,6 +148,7 @@
 ||networkedblogs.com^$third-party
 ||newsharecounts.com^$third-party
 ||opensharecount.com^$third-party
+||opmnstr.com^$third-party
 ||pinterest.com/images/$third-party
 ||pinterest.com/js/pinit_main.js$third-party
 ||pinterest.com/v1/urls/count.json$third-party


spirillen:
Hi @pavelloz thx for the update :smiley:

Khrin:
These are already listed in the Annoyances List: https://github.com/easylist/easylist/blob/master/fanboy-addon/fanboy_annoyance_thirdparty.txt#L226 Personally I don't think these should be listed in the Social List.

pavelloz:
@Khrin the domain I have added is not added.

I added `opmnstr.com` which is missing as i described in original post.

Khrin:
Sorry about that, has been added now. https://github.com/easylist/easylist/commit/103327483e6303e3adb376d46f4eb6f13ced58b7 I've added in the Fanboy's Annoyances List, as you can see, which also include "newsletters" stuff like those offered by Optinmonster. Social List include sharing and social media -related stuff.

pavelloz:
Haha, thanks, i was just preparing second PR with correct addition instead of the wrong one.

This is my first time contributing and im not an expert on syntax yet -- next time ill use my brain more to not pollute the wrong list :)

Khrin:
No worries, always in time to make changes if required.

spirillen:
We all started by learning how to turn on the damned thing by pushing a button :smiley: :o:

Issue #3583, opened 2019-06-07
investigationdiscovery.com


BlazDT:
https://www.investigationdiscovery.com/tv-shows/true-crime-with-aphrodite-jones/full-episodes/scott-peterson-aphrodite-jones-investigates

Site requires US IP and EasyPrivacy list:

Video will not play unless these two requests are whitelisted:

```
@@||src.litix.io/core/*/mux.js$domain=investigationdiscovery.com
@@||fusion.ddmcdn.com/app/eos-usnets^$domain=investigationdiscovery.com
```

ryanbr:
https://github.com/easylist/easylist/commit/03643a2b1d482ea45500ae276a651f97711939c2

PR #4102, opened 2019-09-29
Update whitelisted domains

llacb47:
Add more Discovery media domains needing whitelisting


ryanbr:
Too generic, need a pull request per domain and what are we actually fixing? A bit more descriptive.

llacb47:
I am not going to make a PR for each domain... these sites are all Discovery Media sites and the players are broken without these scripts whitelisted. See 

https://github.com/easylist/easylist/issues/3583 

https://github.com/easylist/easylist/issues/3636

PR #4126, opened 2019-10-06
Block bolt.com logging

benwaffle:
[Bolt](https://www.bolt.com) is a checkout UI
Example: used on https://dita.com


Khrin:
Is this the only element loaded? Do you have the full url?

benwaffle:
It's not a DOM element. Telemetry is reported to https://api.bolt.com/v1/log. The full URL is `https://api.bolt.com/v1/log?`

Screenshot from 2019-10-06 16-03-28

Khrin:
Added: https://github.com/easylist/easylist/commit/104337dcf677fa3f70c6dfa07fc62b66c1540500

PR #4167, opened 2019-10-14
block metrics.readme.io

benwaffle:
Used on https://readme.io, and any other documentation sites that use readme, such as https://developer.box.com


ryanbr:
https://github.com/easylist/easylist/commit/1551e3fa288bce7db02958a97f267a1af30e222f

PR #4170, opened 2019-10-17
A: https://www.kapwing.com/

smed79:


kylefox:
👋 Hi, I run Rewardful — the company these blocked domains belong to.

Blocking `getrewardful.com` has broken the application for our customers who are using a custom domain to host their Rewardful page because it prevents our application script/styles from being loaded.

**Example:** https://affiliates.goreminders.com/signup

This page is technically hosted on our subdomain (`goreminders.getrewardful.com`). The customer's domain (`affiliates.goreminders.com`) points to our Caddy server so we can reverse proxy with SSL.

As you can see, this prevents our application's scripts/styles from loading for our customers:

image

Could you please consider removing `getrewardful.com` from this list?

_**Note:** I tried registering for [the forum](https://forums.lanik.us/viewforum.php?f=64) to make this request, but my account was immediately banned at signup — maybe because I used my `@getrewardful.com` email address?_

kylefox:
We'd also love to ask about getting our `wdfl.co` domain unblocked — our customers add this script (https://r.wdfl.co/rw.js) to their website so they can attribute affiliate link clicks.

I realize this case is not as clear-cut as unblocking `getrewardful.com` because we technically _do_ use it for tracking clicks. However, we don't set any third-party cookies or collect personal data.

Is there another channel I should go through to discuss `wdfl.co` getting unblocked?

PR #4232, opened 2019-11-02
a general filter ###euConsent added

peace2000:
This kind of element encountered on `www.virustotal.com`

That message might not be visible for users outside of EU despite using an eu based proxy. I live in EU and I tried foreign proxies (US, Indonese) but still saw that message even though I should have not.

My personal screenshot: https://user-images.githubusercontent.com/17256841/67633368-bfacfe80-f8b7-11e9-83c5-4a88b722b2f5.png

Others have seen that notice also: https://github.com/NanoAdblocker/NanoFilters/issues/349

### DIFF
--- a/fanboy-addon/fanboy_cookie_general_hide.txt
+++ b/fanboy-addon/fanboy_cookie_general_hide.txt
@@ -259,6 +259,7 @@
 ###ECScookiepanel
 ###EPDirectiveInfo
 ###EP_cookiesPopup
+###euConsent
 ###EUCOOK
 ###EUCookie
 ###EUCookie-Banner


ryanbr:
Its already in the list. Should search the list first.

peace2000:
Sorry didn't realize that because that general filter doesn't work for Virustotal. No matter how much I refresh the page.

https://github.com/ryanbr/fanboy-adblock/issues/1173

krystian3w:
chat with bot on virustootal:

`virustotal.com##.blue.action-button[role="button"]`

peace2000:
> 
> 
> chat with bot on virustootal:
> 
> `virustotal.com##.blue.action-button[role="button"]`

Sorry I didn't understand how this is related to this cookie issue?

krystian3w:
Also need specific block in Fanboy Annoyance.

krystian3w:
Maybe it is time report into uAsset as bug in uBO (I suppose generic works propery with ABP 3.7 / AdBlock 3.66).

krystian3w:
@gwarser can you test how ABP / AdBlock hide cookie bar with generic filter?

peace2000:
@krystian3w I tested with ABP and this works there so yes, I believe this could be an ubo bug after all.

PR #4235, opened 2019-11-02
a generic filter ##[class*="mobilead"] added for mobile ads

peace2000:
Sample page: `https://www.cpubenchmark.net/` (browser window needs to be resized small enough so that the mobile ad will appear on the bottom)

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -19725,6 +19725,7 @@
 ##FBS-AD
 ##LEADERBOARD-AD
 ##[ad-id^="googlead"]
+##[class*="mobilead"]
 ##[href*="//xml.revrtb.com/"]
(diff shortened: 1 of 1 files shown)


ryanbr:
Too generic, I didn't see any ad divs here (even on a mobile UA)

peace2000:
@ryanbr This is the exact name of the element: `##.mobileadunit`

But the displayal of that element is not based on the user agent but the browser window size.

I'm running Windows 10 & Firefox with full screen -> ad is not visible
I'm running Windows 10 & Firefox with small, resized browser window -> ad element becomes visible

Näyttökuva (60)

Here is a screenshot. That element is grey becomes that ad is blocked but the element is still there, that¨s why a cosmetic filter is needed. That screenshot is taken without user agent switcher. Just plain Windows 10 and Firefox, but resized window.

PR #4238, opened 2019-11-03
easylist_whitelist_popup.txt: o2 Germany phone bills

max-foss:
Bill downloads on o2online.de are being blocked, this whitelist fixes it and keeps all other nastyness blocked.


ryanbr:
Resubmit into Easylist Germany would be better here.

https://github.com/easylist/easylistgermany

PR #4258, opened 2019-11-08
Alma media cookie script

peace2000:
The cookie notification is blocked on Alma media's websites by these hiding rules:

`##.alma-data-policy-banner`
`###alma-data-policy-banner`

But Alma media's cookie message has recently started to quickly blink at times, meaning that these hiding rules don't work 100 % efficiently. Blocking a cookie script in the background will work better.

A couple of Alma media owned websites to test with:
`https://www.iltalehti.fi/`
`https://www.tivi.fi/`


ryanbr:
https://github.com/easylist/easylist/commit/099a104eb039f74002824c157dd5f2e1cfb2a4c2

PR #4298, opened 2019-11-18
A: https://www.lemonde.fr/

smed79:


spirillen:
What does `buf.lemonde.fr` do? asking to put it to the right category on the [matrix](https://gitlab.com/my-privacy-dns/matrix/matrix)

smed79:
> What does buf.lemonde.fr do?

Tracking.

PR #4302, opened 2019-11-18
Test pull request

kiboke:
Added a single rule cookie general hide list.


ryanbr:
v1  not bad, just need a sync

PR #4332, opened 2019-11-25
A: https://two-movies.me/watch_movie/Noelle

smed79:


spirillen:
Hey @smed79 you should try 

```
dig look.opskln.com
```

That should response with the `CNAME   ads.ad-center.com` so blocking this would apparently be sufficient!, but your choice, Only sharing my foundings

PR #3582, opened 2019-06-06
A: (klaviyo.com) opened email tracker

smed79:
```
<img src=3D"https://trk.klclick.com/wf/open?3fBYNGeLrtBzu01NGeKDc-2Fsl2Ywpb=
vMN1-2F-2F4nrEM-2Bf4Sr4NwY16Oh1K9C7vOFLGxCVhog4dzy-2BndMEQaifhB-2Bi-2FkaNY5=
PFpcLGBTv-2FmDEuIuUJI08dN-2FS8sYoejDcSL38dUxDdBc-2F0pe5uIqFIzUSHQvz0FMlt9KF=
5gGvRcJl-2BI78ju70EPkDmHaZShhpipO5uTStyjbwYWgDi2Q4xwY20e0aOETzEmO9hQn929R-2=
FoD888RLH4rqCNw5FUWwObqjqdN8h3YPtloG4CI2PufdQTT240MoJL0c83Rv9QxkzIws1ghsRei=
Vt4VSEaL06rhXSjLJYZXiXPSaDF2AIL80yYEHVtQwsfIHk35JKYF1N5CzJf7j2znMWKo8A8SPhz=
V5xgH1HVZeQ0sXJW55pNiuMdcrWRGA-3D-3D" alt=3D"" width=3D"1" height=3D"1" bor=
der=3D"0" style=3D"height:1px !important;width:1px !important;border-width:=
0 !important;margin-top:0 !important;margin-bottom:0 !important;margin-righ=
t:0 !important;margin-left:0 !important;padding-top:0 !important;padding-bo=
ttom:0 !important;padding-right:0 !important;padding-left:0 !important;"/><=
```

PR #4358, opened 2019-11-27
A: (klaviyo.com) opened email tracker

smed79:
`<=`


spirillen:
https://github.com/easylist/easylist/pull/3582#issuecomment-537064282

You can easily do a full blocking of `klclick.com` the domain is only used for all kind of tracking

PR #4346, opened 2019-11-26
A: https://map.baidu.com/

smed79:
https://forums.lanik.us/viewtopic.php?p=151400#p151400

PR #4392, opened 2019-11-30
Fix #4346 - More precise match on baidu tracker

tomac4t:
https://github.com/easylist/easylist/pull/4346#issuecomment-559910228


smed79:
@ryanbr

PR #4412, opened 2019-12-01
Block only cdn.speedcurve.com so the webapp still works

wildlyinaccurate:
SpeedCurve is a performance monitoring tool used by many software developers. The current rule makes the web app inaccessible and breaks images in emails. It should only block cdn.speedcurve.com, which serves the data collection script on other websites.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -2103,7 +2103,7 @@
 ||spamanalyst.com^$third-party
 ||spectate.com^$third-party
 ||speed-trap.com^$third-party
-||speedcurve.com^$third-party
+||cdn.speedcurve.com^$third-party
 ||spklw.com^$third-party
 ||splittag.com^$third-party
 ||splurgi.com^$third-party


llacb47:
Why not add an exception for the web app or a specific endpoint?

wildlyinaccurate:
@llacb47 because cdn.speedcurve.com is the only domain that serves a tracker. Same reason we don't just block google.com - there are more valid domains than tracking domains.

The current rule also breaks images in emails and people (like me) don't realise the ad blocker is interfering.

llacb47:
That didn't answer the question..., wouldn't it be easier to provide the specific affected URLs so they could be whitelisted?

wildlyinaccurate:
Sorry, I probably didn't make it clear that speedcurve.com is one of the
affected URLs.

wildlyinaccurate:
Example of URLs that need to be whitelisted:

- https://speedcurve.com/ (web app)
- https://img.speedcurve.com/home/synlux.png (images from web app)
- https://assets.speedcurve.com/css/speedcurve.444.css (assets for web app)
- https://alerts.speedcurve.com/n.php?hash=e694db2db21d55702aabfe1636f42f25d81a62abde511e97baec8706874a714f&width=600 (image from emails)
- https://wpt1.speedcurve.com/ (part of the web app)
- https://api.speedcurve.com/ (API documentation)
- https://support.speedcurve.com/en/ (web app documentation)
- Probably more? I don't have an exhaustive list.

Example of URL that needs to be blacklisted:

- https://cdn.speedcurve.com/js/lux.js?id=73376842 (a script that does real user performance monitoring)

llacb47:
I see. Well, the rule only blocks `third-party`, so the *only* issue should be images in emails, and even then, most big mail clients serve third-party images through their own CDN, meaning that the email images should be a non-issue in some cases. 

But, to fix that issue, `@@||alerts.speedcurve.com/n.php?$third-party` could be added @ryanbr. 

I don't think PRs like this are often accepted because you could circumvent this list easily-- If your boss decided that adblockers were chewing into your bottom line, what's to say you wouldn't serve the "data collection script" from another subdomain?

wildlyinaccurate:
Ah, that's interesting. Thanks for the explanation! I actually noticed this
with my pi-hole blocking the web app, but I guess that's because it blocks
the entire domain and has no context to determine whether it's first- or
third-party.

I'll close this issue for now :)

wildlyinaccurate:
Also, on this point:

> I don't think PRs like this are often accepted because you could circumvent this list easily-- If your boss decided that adblockers were chewing into your bottom line, what's to say you wouldn't serve the "data collection script" from another subdomain?

A lot of services these days configure their ads & trackers to use either a random top-level domain or a first-party domain. I won't drop names but one big ads platform now serves ads through `<hash>.g00.<host-domain>` and a big analytics platform uses `<random-string>.<tld>`.

So blocking an entire TLD on the basis that the service could switch to another subdomain is, while understandable, possibly quite futile and to the detriment of user experience.

llacb47:
Ah, yes. Instart Logic 😂

spirillen:
> @llacb47 because cdn.speedcurve.com is the only domain that serves a tracker. Same reason we don't just block google.com - there are more valid domains than tracking domains.

Speak for yourself :angry: https://gitlab.com/rpz-zones/google.tld

PR #4432, opened 2019-12-03
A: https://www.dcplanet.fr/

smed79:


Khrin:
Filters were good candidates to be added in my opinion, just I preferred to make them specific to scripts only.

smed79:
Ok, thank you.

PR #4460, opened 2019-12-07
A: https://www.foumovies.net/

smed79:
https://github.com/easylist/easylist/pull/4461


Khrin:
Are you sure it's necessary?

PR #4471, opened 2019-12-07
A: Cookie blocking remnant, shirtsofcotton.com

peace2000:
On `shirtsofcotton.com` there is a dark transparent cover. Obviously it's a leftover of the hidden cookie message.


smed79:
@peace2000 for commit message see https://forums.lanik.us/viewtopic.php?p=131958#p131958

peace2000:
@smed79 All right I changed this. I'll check my other pulls also.

PR #4485, opened 2019-12-08
A: Spam chat by justanswer.com

### LINKED ISSUE #4128: justanswer.com: Intrusive and deceiving pseudo-popup
This is a follow-up to https://github.com/easylist/easylist/issues/2897, as the pop-up has returned:

image

The pop-up doesn't always appear on that site; I've been browsing that site for about an hour or two and it has only just appeared.  It is clearly deceptive as it wouldn't be clear to the average user that the prompt is an advertisement for a third-party service.

I'd like to suggest that attempts to load scripts and stylesheets from `*.justanswer.com` be blocked on third-party sites.

peace2000:
Fixes #4128

It's widely used.


peace2000:
Wrong place.

PR #4508, opened 2019-12-10
A: zz.bdstatic.com & push.zhanzhang.baidu.com

tomac4t:


spirillen:
Thanks for sharing :+1:

ryanbr:
would remove `||push.zhanzhang.baidu.com^` since its replaced by `||baidu.com/push.js`

Issue #4512, opened 2019-12-12
Curalate.com EasyPrivacy List rules


bvellek:
### List the website(s) you're having issues:
There are thousands of websites utilize our javascript experiences that have this issue. Here are some examples:
- `https://www.oakandfort.com/`
- `https://www.macys.com/ce/splash/macyslove/`
- `https://www.bestbuy.com/site/electronics/idea-center/pcmcat345300050032.c?id=pcmcat345300050032`

### What happens?
Curalate develops embeddable javascript experiences for ecommerce sites. These experiences are mistakenly being blocked by EasyPrivacy. These experiences emit events via a jsonp and png (pixel) for metrics aggregation. In the past we've provided the specific routes that can be blocked, but we need to remove the wholesale `*.curalate.com` third party block (from this [commit](https://github.com/bvellek/easylist/commit/8732578dd8342d28edaa9fb484ceafd949e54f70#diff-dbc51871bdd5cf7971a65ef9684d2248R585)) as it prevents not only the metrics event calls, but the entire experience from rendering on the page. 

![image (5)](https://user-images.githubusercontent.com/18407299/70672575-8f85a900-1c34-11ea-8b91-e0b333bc63d6.png)

Here is an example with uBlock disabled:
![Screen Shot 2019-12-11 at 4 48 08 PM](https://user-images.githubusercontent.com/18407299/70673082-2f900200-1c36-11ea-9351-00e4252bbb5e.png)

Here is an example with uBlock enabled: 
![Screen Shot 2019-12-11 at 4 48 39 PM](https://user-images.githubusercontent.com/18407299/70673095-3880d380-1c36-11ea-9660-e9840dc454c7.png)

### List Subscriptions you're using:
EasyPrivacy

### Your settings
- OS/version: MacOS Sierra
- Browser/version: Chrome (73.0), Brave (1.1.20)
- Adblock Extension/version: uBlock Origin (1.18.16)

### Other details:
We'd like to replace the `curalate.com` block with a block for `curalate.com/*/events.png` and `curalate.com/*/events.jsonp` which are the only endpoints that receive our anonymous, non PII events for tracking the number of clicks that occur on our widgets.

ryanbr:
There is a few pixel/metics bits in `https://cdn.curalate.com/sites/century21stores-nzvwxr/site/latest/site.min.js` could this be seperated off?

bvellek:
Hey @ryanbr thanks for taking a look!

[...]

PR #4513, opened 2019-12-12
ISSUE-4512: Update curalate.com EasyPrivacyList block.

bvellek:
Details in: https://github.com/easylist/easylist/issues/4512

Instead of blocking all traffic to curalate.com, it specifically blocks Curalate's tracking urls. An example can be seen on this page which uses Curalate's javascript experience: https://www.macys.com/ce/splash/macyslove/


ryanbr:
Further work needed, see; https://github.com/easylist/easylist/issues/4512

PR #4518, opened 2019-12-13
A: http://blog.niliushui.com/

tomac4t:


ryanbr:
Maybe also include `||baidu.com/push.js` and trim the filter to just `||qihucdn.com/11.0.1.js$script`

tomac4t:
Done. @ryanbr

PR #4528, opened 2019-12-16
fix: https://forums.lanik.us/viewtopic.php?f=64&t=43903

jakubjosef:


jakubjosef:
Previous version was incorrect, see: https://forums.lanik.us/viewtopic.php?f=64&t=43903&p=151879#p151879

ryanbr:
resolved

ryanbr:
Are these filters still needed @jakubjosef ?

PR #4564, opened 2019-12-24
P: A Facebook game won't load, cookappsgames.com

peace2000:
A Facebook game called Buggle2 won't load due to Easyprivacy rule.

https://www.facebook.com/search/apps/?q=buggle2&epa=SERP_TAB

It's likely that other games could be affected so I whitelisted all cookappsgames as there are many of them.


ryanbr:
Wrong list, recommit to EP. Also posting the full url you're trying to whitelist.

peace2000:
All right, I corrected my mistake: https://github.com/easylist/easylist/pull/4627

PR #4567, opened 2019-12-24
A: https://blogs.constantcontact.com/html-email-template/

smed79:


spirillen:
I think you need a :gift: to open.... you are spending to much time here like me....

smed79:
Too much less than before this last months (_busy to do other thing_). 
I will not regret the time spent here if my contributions go back to the users with a positive profit (_improve their browsing experience even a little_).

good wishes for the coming year ... :smile_cat:

PR #4586, opened 2019-12-25
A: http://shopselect.net/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=44059


spirillen:
Hi @smed79 would it be possible to have you ask the original contributor about what the `thebase.in` is about? 

Something is suggesting (visually) that it would be better to block this at a second level rather than only third level.

But I would need the translation for getting the real idea...

smed79:
`thebase.in` is an eCommerce hosting platform that lets your create your own online store.

examples
```
http://amijed.thebase.in/
http://echa.base.ec/
http://scar.theshop.jp/
```

spirillen:
ok, thanks @smed79 :smiley:

PR #4596, opened 2019-12-27
A: http://i-bbs.sijex.net/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=44055


spirillen:
Hmm that would need some translation :smile:

PR #4613, opened 2019-12-30
A: general cookie hide, www.kik.com

peace2000:


llacb47:
Why not just `.cookie_policy_outer`?

peace2000:
Good point. I'll change it.

PR #4621, opened 2020-01-02
A, a general cookie hide, www.similarsites.com

peace2000:


Khrin:
Prefer to have specific filter, in this case.

PR #4624, opened 2020-01-03
A: https://www.speedtest.net/

smed79:


uBlock-user:
@ryanbr This breaks speedtest. Why was this merged without testing ?

uBlock-user:
The go button is misplaced, mid-body of the webpage gets eaten away and the go button doesn't work either. @smed79 You didn't test this either before submitting for a merge ?

PR #4627, opened 2020-01-05
P: A Facebook game won't load, buggle2.cookappsgames.com

peace2000:
A Facebook game called Buggle2 won't load due to Easyprivacy rule.

https://www.facebook.com/search/apps/?q=buggle2&epa=SERP_TAB (here you can test the game if you have a Facebook account).

It's likely that other games could be affected so I whitelisted all Cookappsgames as there are many of them.

If I wanted to fix only this particular game, this rule would be enough:

`@@||pro.ip-api.com/json/$xmlhttprequest,domain=buggle2.cookappsgames.com`

But I consider it better to whitelist all Cookappsgames with this rule:

`@@||pro.ip-api.com/json/$xmlhttprequest,domain=cookappsgames.com`

### DIFF
--- a/easyprivacy/easyprivacy_whitelist.txt
+++ b/easyprivacy/easyprivacy_whitelist.txt
@@ -211,6 +211,7 @@
 @@||player.siriusxm.com/assets/AppMeasurement/VisitorAPI.js
 @@||player.sundaysky.com^$subdocument
 @@||pnas.org^*/fingerprint.js$script,~third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
$xhr is a ubo-ism, please change

peace2000:
@ryanbr It's updated now. Good to learn new things... :D

peace2000:
@ryanbr ?

PR #4633, opened 2020-01-06
A: general cookie block, www.blogit.fi

peace2000:
Easylist cookie already has rules to hide "oil cookie" notification:

`##.as-oil[data-qa="oil-Layer"]`
`##.as-oil-content-overlay`

I added a rule to block the script behind them.


peace2000:
@ryanbr

PR #4634, opened 2020-01-06
M: rule generalization /advertisements/

peace2000:
I think it would a good idea to generalize this rule.


peace2000:
Sorry this was mistake, I now realize that the rule just means that "every other domain but ellefanningham". I'll close this...

PR #4646, opened 2020-01-08
A: (nsfw) voyeurflash.com

smed79:


easylist/easylist_adservers.txt
```diff
@@ -4441,6 +4446,7 @@
 ||seiya.work^$third-party
 ||sekindo.com^$third-party
 ||selectablemedia.com^$third-party
+||selfpua.com^$third-party
```
spirillen: Hi @smed79 

Have tested this domain, and can't get it to resolve with `selfpua.com` or `www.selfpua.com`. Do you have another url, from which you have chosen to block this at  second level?

smed79: Hi,

==> `https://static.selfpua.com/mnpw.js`

spirillen: Thanks :+1: added it to own source :)

PR #4666, opened 2020-01-09
A: magnetdl.com search result pages

mimmiepower:
**Ad link**

To reproduce: search for anything on https://www.magnetdl.com/ or visit one of the categories. There is a box on top of the results that is not meaningful and probably only there to lure people to click on the ad hyperlink.

Maybe not the most elegant solution code-wise.


mimmiepower:
Oops. Ignore this request. It breaks the site.

PR #4685, opened 2020-01-10
P: messaging system breakage, bethesda.net

peace2000:
https://github.com/ryanbr/fanboy-adblock/issues/981
https://github.com/ryanbr/fanboy-adblock/issues/1017

Both are referring to the same issue. There are some discussion how to block it properly, but now it's most important just to disable the problematic rule. Instead of removing it, it's better to leave it there as disabled so that no one else won't add it some day, being unaware of problems it causes.


Khrin:
Better remove it, if necessary.

krystian3w:
Yes remove, if we not add version with `abp-has` construction.

PR #4705, opened 2020-01-13
A: google mobile ads hide, google.fi

### LINKED ISSUE #4004: Google ads not blocked when searching through a mobile browser

`www.google.fi`, `google.com` (not just those, all language variants are affected)

### What happens?

Google ads are not blocked

Easylist

### Your settings

- OS/version: Android 8
- Browser/version: Firefox 68.1
- Adblock Extension/version: Ublock Origin 1.22.2

### Other details:

This rule seems to block ads whichever language variant is being used: `##div#main > div.ZINbbc.xpd.O9g5cc.uUPGi`

You can also reproduce this problem on desktop by using an user agent switcher. More comfortable than using a cell phone to investigate this. :D

This user agent can used, for example:

Mozilla/5.0 (Android 9.0; Mobile; rv:68.0) Gecko/68.0 Firefox/68.0

peace2000:
Google ads for mobile users are not hid.

Here is my issue https://github.com/easylist/easylist/issues/4004 report I reported earlier.

Fixes #4004 

Earlier it seemed that this element was in use in all language variants but it doesn't seem to be so anymore. But it's definetely being used on google.fi.

To reproduce:

1. Use a mobile user agent on web browser. Here is one user agent that can be used:

`Mozilla/5.0 (Android 9.0; Mobile; rv:68.0) Gecko/68.0 Firefox/68.0`

2. Use a search word that will spark ads, such as "tietokone" (computer in English). Word "mainos" means "an ad".

3. You might need to have a Finnish ip and browser.

65819492-d6fbbc00-e225-11e9-8a89-12e37a8e68d3


peace2000:
Ads are blocked when I have this rule  `www.google.fi##div#main > div.ZINbbc.xpd.O9g5cc.uUPGi` and it activates in the logger:

Näyttökuva (99)

And mobile ad results (mainos) are blocked.

llacb47:
You can also show the cosmetic filters applied by clicking the code icon in the toolbar...

peace2000:
@ryanbr ?

ryanbr:
Only specific to one google domain? are you sure?

monzta:
Seems to be working for google.de as well.

peace2000:
Easiest and to-be-sure solution would be to add every Google's language variant. It's hard to test every single one of them by using slow free proxies...

monzta:
I checked some with a VPN and it seems to work. Feel free to create a PR with all domains for this filter but please add to the `! msn.com` section. I added another one I just found: https://github.com/easylist/easylist/commit/c61c116

peace2000:
Hi! I updated this pull req. It has all google's regional domains and it's not in the msn section anymore. :D

llacb47:
www not needed?

peace2000:
Guess not but that's what I have copied directly from Easylist itself.

monzta:
I'd suggest to leave www. as it could cause false positives on subdomains.

PR #4723, opened 2020-01-17
Remove false positive fluctuo.com

PierrickP:
Hello,
I'm the CTO of fluctuo. We are a young (mai 2019) company in shared-mobility data https://fluctuo.com/.
We don't tracks peoples :)

I don't know why our domain is there, it look to be related to our previous domain owner ¯\_(ツ)_/¯
This rules break the use of subdomain (like https://cdn.fluctuo.com) on fluctuo.com.

Best


ryanbr:
Have removed, though if we get any reports or any tracking/performance data/metric scripts, it'll be re-added.

PR #4745, opened 2020-01-19
A: cookie note hide, vapaasana.swedishforum.net

peace2000:
https://github.com/ryanbr/fanboy-adblock/issues/1160

(Yes, this is a Finnish site even though it has a name "swedishforum" as a part of it.)


peace2000:
@ryanbr  ?

PR #4807, opened 2020-01-27
M: Update easylist_specific_hide list for some bet sites in hltv

wolfcon:
- dongying
- onex
- thunderfire


ryanbr:
Is there a better way to filter these @wolfcon  and can you remove old unused filters.

wolfcon:
@ryanbr I am remove and modify unused filters.
hltv.org used <aside> / <div> / <span> etc for bet AD.

They use short url and changed url often. So there is no better block way that I ever find.

PR #4825, opened 2020-01-29
M: poki cookie element hide, 123pelit.com

peace2000:
This has changed. So that's why this update.

This is used on many domains, not on just `123pelit.com`

I however wonder if this rule would be better?

`##div#app-root.c > aside`


ryanbr:
Sure @peace2000 . do test, and submit a new PR if its safe

PR #4848, opened 2020-02-01
A: (update dd4b4bd) https://www.nextinpact.com/

smed79:


papci:
Hi, is it possible to stop  blocking our internal analytics ?  We try to be respectful with our visitors (we do not use Google analytics). We are using CNIL  (french regulator) settings, though we hold no personal data.

llacb47:
The only way it will be unblocked is by following what's written on this page https://easylist.to/2011/08/31/what-is-acceptable-first-party-tracking.html

PR #4850, opened 2020-02-01
A: https://tintin-thegame.com/

smed79:


spirillen:
Hey  @smed79 

I'm a little unsure whether is could be an idea to block this submission within DNS level, What is you thoughts on that?

PR #4856, opened 2020-02-03
A: Add new hidden class for hltv.org

wolfcon:


ryanbr:
I would recommend trimming older, unused filters before commiting new ones @wolfcon

wolfcon:
@ryanbr Yeah, I know. But hltv.org is the most professional website in CSGO. So many bet websites cooperate with them, and full of advertisements on it. Those filters are all useful.
If some of them are invalid, I will PR again and delete them.😁

PR #4861, opened 2020-02-06
Made it so that entries for 4 major sites in EasyList also covered their Tor addresses

DandelionSprout:
This chiefly covers four sites, namely `Facebook`, `New York Times`, `PornHub` and `DuckDuckGo`. A rule of thumb is that Tor versions of major sites have the exact same interface as the normal versions, so I felt it was considerably overdue to expand the entires to cover those alternative URLs.


ryanbr:
Missing pornhub in easylist_adult..

DandelionSprout:
I think I've added it to those categories as well now.

PR #4867, opened 2020-02-08
A: generic Alma Media data policy banner block, iltalehti.fi

peace2000:
Alma Media's data policy banner is being hid by these rules:

`##.alma-data-policy-banner`
`###alma-data-policy-banner`

I now added blocking to the script that loads that banner.

List of Alma Media's services: `https://www.almamedia.fi/palvelut/kaikki-palvelut`

(Not every Alma Media's service use this but many do).


ryanbr:
I don't feel its safe to block this script, I'll say no here

peace2000:
Ok. It probably then has other functions as well, not just this banner?

PR #4871, opened 2020-02-08
A: adservers, kaleva.fi

peace2000:
Encountered on kaleva.fi


spirillen:
The domain name of this submission `admp-tc-bauer.adtlgc.com` gives me the impression that it should/would be possible to block `adtlgc.com` entirely. If you try to visit that domain, you just get a timeout()

ryanbr:
We should be adding Non english adservers @peace2000  need to suggest filters that are common for english sites. Also the A: commit message needs to be cleaned up.

ryanbr:
Closing.

PR #4872, opened 2020-02-08
M: Moved an adserver Easyprivacy -> Easylist

peace2000:
It's listed under "Swedish" section. But it's headquarters are now in London so it's not "Swedish" anymore in that sense. :D

It can be encountered on this link: https://pakkotoisto.com/

It does connect to a domain that sounds like tracking (track.adtraction.com) but it's probably used for both ads and tracking. It's frontpage on adtraction.com suggests that it's for ads.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -676,6 +676,7 @@
 ||adtotal.pl^$third-party
 ||adtpix.com^$third-party
 ||adtrace.org^$third-party
+||adtraction.com^$third-party
 ||adtransfer.net^$third-party
 ||adtrgt.com^$third-party
 ||adtrieval.com^$third-party
--- a/easyprivacy/easyprivacy_trackingservers_international.txt
+++ b/easyprivacy/easyprivacy_trackingservers_international.txt
@@ -907,7 +907,6 @@
 ||adten.eu^$third-party
 ||adtlgc.com^$third-party
 ||adtr.io^$third-party
-||adtraction.com^$third-party
 ||bonnieradnetwork.se^$third-party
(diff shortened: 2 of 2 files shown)


mdanielsson:
Hi!

We use the domain track.adtraction.com mainly for conversion tracking, there are however ads served here as well. 

They are always served with https://track.adtraction.com/t/t?t=1 (query parameter t = 1). Would it be possible to move the domain back to EasyPrivacy and also limit the scope of the EasyList block from adtraction.com to https://track.adtraction.com/t/t?t=1?

Blocking all of adtraction.com in EasyList has led to various issues with third party widgets depending on for example api.adtraction.com to work properly. 

Thanks!

llacb47:
provide an example of a site with such "issues"

mdanielsson:
Our main concern is our conversion tracking is affected by the attempt to block adverts in EasyList. Can we meet half way here?

llacb47:
I always get puzzled by these requests, Easylist and Easyprivacy are enabled by default in many adblockers such as uBlock Origin and Adguard, moving the domain will not help you. Plus, it is conversion tracking for ads that will be blocked...  so no conversions will occur. It is a non-issue unless I misunderstand you.

mdanielsson:
Thanks for taking the time to respond @llacb47 

The main issue here is that the majority of our links are not ads but content based direct links provided by our affiliated publisher (and/or facebook or Google search third party tracking links) that points directly to the target (for example an e-commerce site). 

We also have an option where we actually deliver the content (for example a banner ad) as I wrote above. In this case we agree it's reasonable to add this domain (track.adtraction.com with the query parameter t=1) to EasyList.

PR #4873, opened 2020-02-08
A: cookie note block, alppimatkat.fi

peace2000:
Here is the link without an asterisk: https://www.alppimatkat.fi/api/v1/notifications

That text on that link speaks only about cookie policy stuff. It has no other functions.


ryanbr:
Need to clean up the A: commits. see other examples. use the url. and place comments in the commit message.  @peace2000

PR #4883, opened 2020-02-11
A: adserver, ts.fi

peace2000:
These Citymarket images (ads) are loaded from domain adswiper.eu.

image

Direct link to an ad sample: https://cdn-1.adswiper.eu/images/958cf588fe324f9a8aebee5d871d817a/8_kirjolohi.jpg

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -721,6 +721,7 @@
 ||adsvcs.com^$third-party
 ||adsvert.com^$third-party
 ||adsvids.com^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
Commit message needs fixing, and its non english. We're keeping Easylist for english sites

peace2000:
But this ad server can be used anywhere.

peace2000:
I just tried to prove that server loads ads and therefore it should be blocked. I can add this to the Finnish Easylist but I thought that adswiper.eu couls be used more internationally as well.

PR #4905, opened 2020-02-13
P: gigantti.fi

peace2000:
Test link: https://www.gigantti.fi/catalog/tietokoneet/fi_naytot/naytot

Näyttökuva (107)

dynamicyield.com filter causes breakages. Sorting products according to specific criteria doesn't work, nothing happens. After making this whitelist, sorting according to criteria works.


peace2000:
@monzta ?

PR #4913, opened 2020-02-14
A: hardware.fi

peace2000:
a general hiding rule.

Sample link: https://www.hardware.fi/uutiset/artikkeli.cfm/2020/02/12/huikean-halvat-ssd-levyt-120gb-parilla-kympilla-teratavun-levyt-reilun-satasen

kuva


Khrin:
This proposal cannot be accepted, the proposed filter is too much generic.

peace2000:
Oh. Has ##.advertisement caused breakages or do many pages use that for non-ads?

PR #4938, opened 2020-02-21
A: https://apkcombo.com/adblock-plus/org.adblockplus.android/

smed79:


peace2000:
I would ad this rule: apkcombo.com##div.advertisement as it blocks also those "Advertisement" texts. :)

krystian3w:
👓 

`(...),apkcombo.com,(..)##.advertisement` so added..

peace2000:
Oh, yea I didn't notice to scroll to the end. :D Removed my comment.

PR #4950, opened 2020-02-21
A: (#3313) webconfs.com

smed79:


peace2000:
Those 2 letters at the end of the filename have to be capitalized, otherwise it won't get blocked.

||webconfs.com/wp-content/uploads/2017/08/300x250bw.png -> ||webconfs.com/wp-content/uploads/2017/08/300x250BW.png

krystian3w:
Letters "size" in path is ignored:

https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters#match-case-modifier

IMO only AdGuard added `$match-case` modifier if we no use regex.

krystian3w:
@peace2000  the best test this in "My filters" panel and force refresh site.

peace2000:
I just tested, yes it seems that no capitalization is needed. It seems that Ubo's element picker cares about case but if it's in "My filters", it doesn't care.

peace2000:
@gorhill, should Ubo's element picker act case sensitively or is this a bug?

gorhill:
> is this a bug?

Yes; the element picker is case-sensitive, I need to make it case-insensitive.

PR #4954, opened 2020-02-22
M: spot-a-shop.fi | cookie notification filter update

peace2000:
Cookie banner hiding rule, cookie banner located at the top.


krystian3w:
Will you remember to check this compressed class name for random characters every 2 weeks on average?

I see no logic in avoiding:

`spot-a-shop.fi###__next > div > div:nth-of-type(1) > div[class^="css"]`

peace2000:
There is point on what you say if that's true but i don't know if they change it or not.

PR #4956, opened 2020-02-22
M: linkedin.com | cookie notification filter update

peace2000:
Note! This will only hide the banner that is visible for users not logged in:

Linkedin ulkona

Remade this old pull req https://github.com/easylist/easylist/pull/3277 that is partly redundant. The redundant rule it has is this one: `###global-alert-queue` that is not in use anymore in Linkedin.

That pull req also has this rule `###artdeco-global-alert-container` that is relevant, but it's a bit problematic. It's a filter to hide the cookie banner at the top for logged users, but if it's applied, the top banner will be detached from the top when the site is scrolled down. That's why I would not apply it. (This is occasional problem you probably are aware of). Check the screenshot below:

­
­
­kuva


ryanbr:
This won't break the scroll bar @peace2000 ?

peace2000:
No it wont. The one that I mentioned in the comments, will break, that's why I didn't add it to my own pull req.

peace2000:
I would also close that old pull req: https://github.com/easylist/easylist/pull/3277

PR #5009, opened 2020-02-28
Delist simply.com

tomsommer:
We aquired the domain simply.com and are launching a webhosting project on it. However the domain is blocked by you, due to past advertising activity (before we bought it).

Can you delist the domain from your blocklists?


Khrin:
Website isn't alive yet? I got redirected somewhere else instead.

tomsommer:
It's live. We are renaming the UnoEuro brand to simply.com - we just haven't flipped the hostname yet. Hard to do when we know it will be blocked in most adblockers :(

Khrin:
Removed the entry.

PR #5020, opened 2020-02-28
A: jpost.com

peace2000:
Specific ad container hide on the top.


Khrin:
Not all Outbrain stuff is considered as advertisement, as far I can see their sponsored links are correctly blocked.

PR #5024, opened 2020-03-01
A: https://www.file-up.org/y2lhcmug0bjl

smed79:


peace2000:
Old pull req that has some explanation: https://github.com/easylist/easylist/pull/2835

ryanbr:
we probably dont need this, since the adservers are blocked already. Maybe a fix for the popup

PR #5027, opened 2020-03-01
A: https://yready.net/

smed79:
> The push monetization is a form of advertising that relies on push notifications. This means that the ads are delivered to the browser or the device of the subscriber as push notifications...


ryanbr:
Should this go into Fanboy Notifications List? @smed79

smed79:
The push notification serve ads.

PR #5075, opened 2020-03-17
A: https://7-up.net/logjs5gnm223

smed79:


Khrin:
I don't see the need for these rules.

PR #5084, opened 2020-03-17
P: ryanbr/fanboy-adblock#613

tvc:
Fixing same issue referenced.

Blocked script is: https://www.abc.net.au/assets/radio-2.52/js/components/newsletter/wcms.newsletter.jquery.min.js


Khrin:
Do you have a valid working link? That one in the issue report you've used as reference is expired and others are working correctly, like for example here: `https://radio.abc.net.au/programitem/pgoGRNdJ4G` , nor I've seen this element blocked at all.

tvc:
https://www.abc.net.au/triplej/programs/drive/listen-to-lewis-do-an-exit-interview-with-his-ex-girlfriend/11758226

The on demand radio clip on this page will not load without the blocked script.

PR #5098, opened 2020-03-19
A: https://alahlionline.com/

smed79:


Khrin:
I believe that's too much generic.

smed79:
Ok, thank you.

PR #5137, opened 2020-03-25
A: (nsfw) porno333.com

smed79:


ryanbr:
Should specify $domain= on the other whitelist, to be safe

smed79:
@Khrin

PR #5181, opened 2020-04-01
M: Update hltv bet ads in widget.

wolfcon:


ryanbr:
Removed these, create a new PR

PR #5192, opened 2020-04-03
A: https://translate.google.com/

smed79:
https://imgur.com/Nt2Bq3I.png


Khrin:
I don't think this deserves to be blocked.

smed79:
Ok, thank you.

PR #5225, opened 2020-04-12
A: gamefaqs.gamespot.com

peace2000:
Cookie hiding rule. Specific rule needed, because gamefaqs.gamespot.com has generichide-filter.


ryanbr:
where is this generichide, shouldnt this be fixed there? @peace2000

peace2000:
I saw it in my logger back then, it was in the Easylist If I recall correctly. But seems that something has been changed. Therefore, this pull req isn't relevant anymore. Thank you anyway.

PR #5267, opened 2020-04-23
A: https://www.androidweblog.com/huawei-nova-plus-root-install-twrp-r…

smed79:


ryanbr:
Does this need to be generic? @smed79

PR #5292, opened 2020-04-27
Reallow ?type=pageview&

foygl:
This breaks some sites that use Zendesk for support.

This rule was added as part of https://github.com/easylist/easylist/commit/ca1225e827cadd4929e1ecc38e4a85dc02f60f3b. I would argue this was added erroneously as it's far too generic a change for the intention of the original commit.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -5055,7 +5055,6 @@
 ?trackGroup=*&referrer=
 ?trackingCategory=
 ?triggertags=
-?type=pageview&
 ?type=performance&
 ?type=ping&$~subdocument
 ?url=*&id=*&res=*&ref=


Khrin:
What is exactly breaking and where?

foygl:
An example is on https://www.bugsnag.com/ where it causes the following error in the browser logs (Firefox 75.0 with uBlock Origin 1.26.0):
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://bugsnag.zendesk.com/embeddable_blip?type=pageView&... (Reason: CORS request did not succeed).
```

llacb47:
`/embeddable_blip?type=pageView&...` is view tracking. Why would it be removed?

foygl:
Ok. Maybe it's blocking a legitimate tracker then. I will close this PR.

PR #5299, opened 2020-04-28
M: Update bet ads class in hltv.org

wolfcon:


ryanbr:
You need to re-check all the filters, before adding new ones.

wolfcon:
They use random class for antiblock.

plz need help. How to block this `<a>` & `<div>`. 
image
image
It doesn't work in my `MY FILTER LIST` with Adblock plus.
```
||out.hltv.org/*.php?ref=*
||bbfun88.com^
```

wolfcon:
Already checked and change some filters.

PR #5300, opened 2020-04-28
A: (nsfw) Fix popunder leading to syndication.exoclick.com on anybunny.tv

Yuki2718:
Steps to reproduce [NSFW]

<details>
1. Go to `https://anybunny.tv/` and click on any video.

anybunny1

2. Then click on any related video from the below images.
anybunny2png

3. Popunder is made only at the first visit. You need to delete cookies to reproduce again.
anybunny3

</details>


Yuki2718:
@ryanbr The  popunder still appears with the rule added, please check.

PR #5305, opened 2020-04-30
M: Admez should be blocked only for .com domain

janbarasek:
AdBlock is now blocking my decent page in Czech. I think it should be blocked only for the com domain.

Snímek obrazovky 2020-04-30 v 9 29 47

Historically, only the com version should have been blocked, I think an error occurred:

Snímek obrazovky 2020-04-30 v 9 31 41

Results: https://github.com/easylist/easylist/search?q=admez&unscoped_q=admez

Thanks!


janbarasek:
Thanks!

PR #5312, opened 2020-04-30
Add specific installs of piwik analytics.

YipYup:
Updated.

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -1190,6 +1190,7 @@
 ||matchesfashion.com/js/Track.js
 ||mate1.com^*/iframe/pixel/
 ||mate1.com^*/reg.logging.js
+||matomo.rdforum.org^
 ||mayoclinic.org/js/gconversion.js
 ||mcloud.to/log/
 ||mdmpix.com/js/connector.php?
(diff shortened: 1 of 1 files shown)


peace2000:
`matomo.rdforum.org` - 404

`meismyname.rdforum.org` - alive but where it's in use?

YipYup:
@peace2000 - `matomo.rdforum.org` was used up until about 7 days ago. It had been used for the past ~8 months to host piwik/matomo analytics. The hostname was changed to `meismyname.rdforum.org`. I added the matomo in case they revert it back.

They are used at `https://www.rdforum.org/`, called in the HTML HEAD as in-line script.

```
<script type="text/javascript">
		var _paq = _paq || [];
		
			_paq.push(['setUserId', 'ForumUserNameGoesHere']);	
			_paq.push(["setCookieDomain", "*.rdforum.org"]);	
		
		_paq.push(['setRequestMethod', 'POST']);
		_paq.push(['trackPageView']);
		_paq.push(['enableLinkTracking']);
		(function() {
			var u="https://meismyname.rdforum.org/";
			_paq.push(['setTrackerUrl', u+'p.php']);
			_paq.push(['setSiteId', '1']);
			var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
			g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'p.js'; s.parentNode.insertBefore(g,s);
		  })();
	</script>
```

PR #5358, opened 2020-05-15
Remove: honeybadger.io

joshuap:
Hey y'all, I emailed with @monzta in 2018 and they removed us from this list once before. Could you remove this entry again? [Honeybadger.io](https://www.honeybadger.io) is an error reporting service for web developers. We also have a strong [anti-tracking policy](https://www.honeybadger.io/do-not-track/) ourselves.

Our customers use our honeybadger.io domain in two ways:

1. Some use our CDN, js.honeybadger.io to serve our [client-side JavaScript error reporting library](https://docs.honeybadger.io/lib/javascript/index.html). Others install this package via NPM.
2. Once installed, our library reports client-side errors to api.honeybadger.io

Both of these operations are currently blocked, which is negatively affecting our customers (especially the second operation: access to api.honeybadger.io).

Could you please remove us again? Thanks!!

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1063,7 +1063,6 @@
 ||hivps.xyz^$third-party
 ||hlserve.com^$third-party
 ||hmstats.com^$third-party
-||honeybadger.io^$third-party
 ||hoood.info^$third-party
 ||hopurl.org^$third-party
 ||hospitality-optimizer.com^$third-party


ryanbr:
I have looked at this, again. its fine to keep in Easyprivacy.

Due to referrer, UA checks and setting cookies. Anti-tracking Policy's can change without notification, so they hold little sway.

ryanbr:
Left open, so unless there is a 3.0, version 2.0 of this script isn't acceptable.

joshuap:
@ryanbr thanks for your response! Can I request a compromise?

We provide the same service as sentry.io and rollbar.com, which both appear in `easyprivacy_general.txt` but not in `easyprivacy_trackingservers.txt`. Their entries are much less restricted than ours. For example, Rollbar's root domain isn't blocked, only their CDN-hosted scripts:

https://github.com/easylist/easylist/blob/d13eff186bd30af63e0582895e95443f2f02b8b7/easyprivacy/easyprivacy_general.txt#L3503-L3504

Would you merge this PR (removing `honeybadger.io` from `easyprivacy_trackingservers.txt`) if I add the following entry to `easyprivacy_general.txt`?

```diff
+ /honeybadger.js
+ /honeybadger.min.js
```

Our main problem is that our **root domain** is blocked, which means that clients cannot connect to our API at all. I'd really appreciate a fair resolution on this!

---
PS: You mentioned that we set cookies, but _we do not set cookies_. If you scanned our library code for the term "cookie", we *do* have some code which parses a cookie string using a regexp, but that is only if the user of our library intentionally provides it. We do use the referrer and UA strings by default, however.

> Due to referrer, UA checks and setting cookies.

ryanbr:
Okay, adjusted then. Does that look better @joshuap ?

joshuap:
yes, thanks @ryanbr!

Yuki2718:
@joshuap Does https://github.com/easylist/easylist/commit/f8467a715b0bf50a8aea3b79c484eecc4a655815 no more affect your problem?

joshuap:
> @joshuap Does [f8467a7](https://github.com/easylist/easylist/commit/f8467a715b0bf50a8aea3b79c484eecc4a655815) no more affect your problem?

Can someone remind me what the removal of the `$third-party` suffix means?

Yuki2718:
@joshuap You said
> Our main problem is that our root domain is blocked,

Without `$third-party` these scripts will always be blocked, while with the option blocked only if the scripts come from third-party sources.

joshuap:
> @joshuap You said
> 
> > Our main problem is that our root domain is blocked,
> 
> Without `$third-party` these scripts will always be blocked, while with the option blocked only if the scripts come from third-party sources. But I don't know what your comment actuall meant, the removed `||honeybadger.io^$third-party` rule couldn't block root domain itself thanks to `$third-party`.

In general it would be _nice_ if people could host a script called honeybadger.js themselves and it wouldn't be blocked, but the main thing we want to avoid is requests to the honeybadger.io and/or api.honeybadger.io **domains** being blocked, which I think that's not an issue here?

Yuki2718:
> I think that's not an issue here?

Sure, no.

PR #5379, opened 2020-05-20
M: Removed https://www.moscarossa.biz/ links hiding (nsfw)

stefanoio:
Hi, I've noticed a strange behavior in some Google search results and decided to investigate it. Every search result for moscarossa.biz is blocked, that's just an example:

Schermata 2020-05-20 alle 13 51 40
After some digging I found out that the same goes in many other legit websites (other search engines, directories, forums, blogs...) where links to moscarossa.biz are hidden. So I tracked back the blocking rule and I think it might have been added by mistake?
It's part of this commit 6eb61a06fd724d63d4be3ea3e978c61463ff5d45 and I don't get why it's there. The commit is about some xcafe.com website, but it also adds to easylist_general_hide links to: freeadult.games (I checked and it seems to be some sort of free trial with credit card subscription scam website), track.trkinator.com (no idea what that is, I can't reach the domain), and finally www.moscarossa.biz, which is one of the top 3 escort listing websites in Italy. Escort listing is adult business but is totally legal in Italy and AFAIK there's no reason to ad-block it. I have to admit it's not a good looking website, being https://torino.bakecaincontrii.com/donna-cerca-uomo/ and https://www.escortforumit.xxx their competitor, but still it should be a normal website, no advertising or tracking or something.
Thank you!


llacb47:
Related: `https://forums.lanik.us/viewtopic.php?f=64&t=44483`

stefanoio:
Thank you @llacb47 for the additional information. Sorry, I didn't check the whitelist files, I've added a commit to remove that rule as well. By reading the forum thread I've seen that the owner (?) of moscarossa reported some examples of legit websites linking them but those have not been whitelisted. But my point in fact is: there are too many websites (first of all google, it looks like they have 151k blocked results from google.it) linking it to add every single one of them, and they are not ads. If there's some actual ad for moscarossa that should be blocked, non the entire website.
Schermata 2020-05-21 alle 09 22 10

ryanbr:
sure.

stefanoio:
Thank you @ryanbr !

PR #5407, opened 2020-05-26
[FIX] fanboy_social_general_hide remove tag-social rule

mart-e:
This rule was added at a7d474d019e8ed99f as hidding part of
https://www.ibtimes.co.in/ice-age-roundworms-frozen-47000-years-wake-start-eating-if-nothing-happened-776230

This class no longer seems to appear in the given page and may block
wordpress content.
The `<article>` entries on wordpress will have class in the form
`tag-<name>` where `<name>` is the name of the tag used on the article

https://unodieuxconnard.com/2020/05/26/le-monde-dapres-naura-pas-lieu/
has several tags, including one "social", making `<article class="tag-social">`

### DIFF
--- a/fanboy-addon/fanboy_social_general_hide.txt
+++ b/fanboy-addon/fanboy_social_general_hide.txt
@@ -11477,7 +11477,6 @@
 ##.tab_social
 ##.tafSocialButton
 ##.tag-line-social
-##.tag-social
 ##.tags-share
 ##.talk-sharing__facebook
 ##.talk-sharing__other


ryanbr:
Its a good generic, even if not used on the site.  Is there an actual issue with the element?

mart-e:
> Is there an actual issue with the element?

Yes, the whole article is hidden, see the below screenshot with the result. You get directly the comment section without the main content.

Screenshot_2020-06-01 Le monde d’après n’aura pas lieu

while with uBlock disabled you can read it:

Screenshot_2020-06-01 Le monde d’après n’aura pas lieu2

To show you it's generic, take the following article on my personal wordpress blog (with a default wordpress theme), I have added a tag "social" on the article and the whole article is hidden too!

https://mart-e.be/post/2014/02/23/supprimer-des-utilisateurs-indesirables-sur-gnu-social/
Screenshot_2020-06-01 Supprimer des utilisateurs indésirables sur GNU Social – mart-e

Every wordpress blog can potentially have this issue. Maybe the rule ` .tag-social` must be a bit more precise if a good generic rule. Is it possible to find other examples on where it's used?

peace2000:
This rule doesn't seem to be on Fanboy Social anymore, so this can be closed.

mart-e:
Removed at 7462518acf9ee4121af8e7eab3ce75b43c512706 indeed

PR #5435, opened 2020-05-29
A: (onebitco.com) (#5434 & #5436)

smed79:


ryanbr:
I prefer to not add domains to this filter.

PR #5436, opened 2020-05-29
A: https://onebitco.com/ (#5435 & #5436)

smed79:


ryanbr:
I prefer to not add domains to this filter.

PR #5437, opened 2020-05-30
pageviews.toolforge.org/ad_block_test.js

tbertels:


llacb47:
Why?

tbertels:
On https://pageviews.toolforge.org:

> Looks like you are using an ad blocker!
Pageviews Analysis shows no ads, but the ad blockers blacklist websites that collect metrics. In our case, we're just trying to show you metrics, not collect them!

>This issue may be resolved by simply **updating your ad blocker, specifically the EasyPrivacy list**.

Unless it gets added to the whitelist, updating the EasyPrivacy list won't change anything.

llacb47:
Easyprivacy doesn't fix anti adblock.

tbertels:
Why does it say `Anti-Adblock`
at https://github.com/easylist/easylist/blob/master/easylist/easylist_whitelist.txt#L475?

ryanbr:
Those are existing rules, and very limited to certain sites. To Limit liability.

PR #5440, opened 2020-05-30
A: outlook.live.com

peace2000:
Remnant banner on the right, after ads inside of it are blocked.

remnant banner


ryanbr:
No pull, anti-adblock message..

peace2000:
Ok, however if ads inside that bar would have not been blocked, that ad bar would just be a normal ad bar. In that case it could have been hid?

It always depends how things are interpreted. :)

peace2000:
outlook mainos

Still no? @ryanbr

ryanbr:
I checked outlook here, and no advert was seen.

peace2000:
If you disable Easylist, then it's visible.

peace2000:
My point just was that I consider this a legitimate target because when ads are not blocked, this is just a normal ad bar, that couöd be hid as a normal ad hiding process.

PR #5456, opened 2020-06-01
A: facebook.com

peace2000:
Adblocking for the Finnish version of Facebook.

Or should this to be done differently?


ryanbr:
Shouldn't this be left to the Finnish list? @peace2000

krystian3w:
also you ate "` `" (space) very old ABP need this space to property use filter.

peace2000:
@ryanbr 

That's a good question. Facebook isn't a Finnish site though it does provide a Finnish UI. Personally I have thought that sites would be handled according to their nationality even though they would have multiple languages available. That has been the case with Twitter as well:

`twitter.com#?#div[data-testid="UserCell"]:-abp-contains(/Promoted|Gesponsert|Promocionado|Sponsorisé|Sponsorizzato|Promowane|Promovido|Реклама|Uitgelicht|Sponsorlu|Mainostettu|推廣|推广/)`

`twitter.com#?#div[data-testid="placementTracking"]:-abp-has(svg + div[dir="auto"] > span:-abp-contains(/Promoted|Gesponsert|Promocionado|Sponsorisé|Sponsorizzato|Promowane|Promovido|Реклама|Uitgelicht|Sponsorlu|Mainostettu|推廣|推广|推薦|推荐/))`

`twitter.com#?#div[data-testid="trend"]:-abp-contains(/Promoted|Gesponsert|Promocionado|Sponsorisé|Sponsorizzato|Promowane|Promovido|Реклама|Uitgelicht|sponsorluğunda|Mainostettu|推廣|推广/)`

`twitter.com#?#h2[role="heading"]:-abp-contains(/Promoted Tweet|Gesponserter Tweet|推廣推文|推广推文/)`

Is this a policy change?

I could of course add that rule the Finnish list but the problem is that Eyeo still haven't added it as part of ABP's stock lists... So it's only avaible for those who have added the Finnish list manually. Ubo does have it as a stock list...

peace2000:
@krystian3w Ok, I added a space. (Though personally I think that old ABP users should rather update their versions to newer ones.)

krystian3w:
Some people still like use Firefox 36 (some site still works) on Windows XP with ABP `v1.X.X`.

peace2000:
Added this to the Finnish list, hopefully ABP will accept the Finnish list soon.

PR #5535, opened 2020-06-11
A: #5534

smed79:


ryanbr:
the btn- stuff was removed because it causes too many false positives with font awesome.

PR #5539, opened 2020-06-12
A: Gawker Affiliate Ads

thedannywahl:
Hides sponsored/affiliate sidebar ads on Gawker media properties.

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -20154,6 +20154,7 @@
 ##[onclick^="window.open('https://www.brazzersnetwork.com/landing/"]
 ##[onclick^="window.open('window.open('//delivery.trafficfabrik.com/"]
 ##[src^="/Redirect.a2b?"]
+##a[data-ga*="commerce"]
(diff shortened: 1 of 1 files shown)


llacb47:
I think it's too generic, you should restrict it to specific domains. Also it's called Gizmodo Media Group now.

ryanbr:
Needing examples, @llacb47  is right, probably too generic.

thedannywahl:
`##a[href*="click.linksynergy.com"]`

is a more specific selector which hides the affiliate ads on G/O Media properties.  There's an example of this with a specific domain prefix in fanboy's annoyance:

https://github.com/easylist/easylist/blob/dae2d1271649dd10b62a8132228c9c41328ff72e/fanboy-addon/fanboy_annoyance_specific_hide.txt#L3171

I'm not sure if it's better to prefix every domain where this is known to be used or if it should be applied generically as above. Also, if generically, I'm not sure which list to create a PR for.  Here's how the content looks with and without the rule applied:
without filter

with rule applied:
with filter

the text above and below the placeholder appears to be part of the CMS and not generated by the affiliate code.

PR #5543, opened 2020-06-13
A: www.op.fi

peace2000:
Cookie banner general blocking rule and a specific hiding rule.

I left "_www_" intentionally to the beginning of the specific rule in order to avoid false positives as the url is very short and "_op_" could be an ending of another domain also.

https://www.op.fi/


ryanbr:
www. isn't needed and `/consent.js`  seems to be too generic.

PR #5555, opened 2020-06-16
P: mehilainen.fi

peace2000:
Hiding the cookie element breaks scrolling.

https://www.mehilainen.fi/


ryanbr:
Rebase needed @peace2000

peace2000:
I have to make a new one, I cannot fix this req.

PR #5563, opened 2020-06-17
thehackernews.com ads

SeriousHoax:
Ads: 

<details><summary>Screenshot:</summary>

</details><br/>

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -6173,6 +6173,8 @@ gadgetsnow.com##.nativecontent
 9gag.com##.naughty-box
 animenfo.com##.nav2
 thetechpoint.org##.navigation-banner
+thehackernews.com##body > .google.cf
+thehackernews.com##.right-box > .zoho-box
 typo3.org##.navigationbanners
 zeenews.india.com##.navnw-banrpad
 newsbusters.org##.nb-a-p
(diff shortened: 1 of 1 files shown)


llacb47:
do you really need to use `>`?

SeriousHoax:
> 
> 
> do you really need to use `>`?

You're right. I don't need to. I just changed it. Is it alright now?

llacb47:
I mean, isn't it enough to use the class selectors by themselves?

SeriousHoax:
> 
> 
> I mean, isn't it enough to use the class selectors by themselves?

Suggest me what you think should be the better option and I'll make the change.

Khrin:
https://github.com/easylist/easylist/commit/e71eac111e83417a526271c51d4e2e84edd8fb3b

Yuki2718:
`##body .google.cf`
is less efficient than
`##body > .google.cf`
but in this case the simple class was enough as @llacb47 commented unless they caused false positive. Also you should check if those ads can't be blocked with dev tools before committing cosmetic filters, but in this case they're called from inline-script so unblockable in standard ABP syntax EasyList honors (though regex can be used, it's not worth as these images' paths will change soon).

I make PR only if I'm confident the rules are optimal, otherwise PR can take more time than issue reports by requesting correction and waiting for it, if they chose not to ignore the request and fix by themselves - often the case in AdGuard. Well, this should be reported back to AG so they can remove the redundant rules.

@llacb47 How do you think about
```
thehackernews.com##+js(acis, images, thn.li)
thehackernews.com##+js(acis, group, Math.random)

```
of kinds, worth reporting to uAssets?

PR #5581, opened 2020-06-21
P: mehilainen.fi

peace2000:
Hiding the cookie element breaks scrolling.

https://www.mehilainen.fi/


peace2000:
@spirillen Just being curious, what is the point of that "anti-whitelist"?

spirillen:
@peace2000 Nice to see some curiosity :smiley: 

The answer is rather simple if you follows my [projects](https://www.mypdns.org/project/), it's all about privacy, non tracking, and if that "breaks" a site, well [the site is badly written](https://www.mypdns.org/Q12)

This means if I find something that "might(¹)" be opening up for any kind of tracking, it gets a line in the anit-whitelist.

But as mentioned other places, I was rather astonished, that any other noticed that list, as it was never thought as anything else but a private list shared with a few friends and family.


**(1)**: Notice the **_might_**, as that is a "private" project, so evidence is not that mandatory :smiling_imp:

peace2000:
@spirillen Oh, ok. :D

Yeah, it's annoying that sometimes tracking can't be blocked without breaking a site. That happens a lot...

> But as mentioned other places, I was rather astonished, that any other noticed that list, as it was never thought as anything else but a private list shared with a few friends and family.

That's what happens on the internet :D Github nicely mentions every linking that happens on Github.

spirillen:
> Github nicely mentions every linking that happens on Github.

Yep, that's one of the reasons I put the ref link inside the commit....  For who can remember where the stuff came from and why in 6 month? I can't in a week :smiley: there is too much going on to be blocked.

> it's annoying that sometimes tracking can't be blocked without breaking a site.

I say it's annoying site owners doesn't know how to write there source code without breaking there sites, but hey, it's not my wallet that's bleeding from badly coded sites :smiling_imp:

llacb47:
This syntax is invalid, but since I'm blocked by spirillen I guess he doesn't need my help :DD https://github.com/spirillen/ublockorigin-rules/commit/ef45c695d143c7e53e323765298eebaf55505c35

peace2000:
Site has changed, not relevant anymore.

PR #5599, opened 2020-06-26
Blocked tracking for spot.im

danielunited:
Spot.im is a commenting system implemented on various websites such as foxnews and techcrunch.


ryanbr:
We don't need these, already blocked here.

PR #5624, opened 2020-06-29
A: https://areena.yle.fi/

peace2000:
Tracking.


ryanbr:
Don't we have `/b/ss/*&aqe=` in the list to cover the `dax.yle.fi/adobe/b/ss/yleproduction/$image,domain=yle.fi` @peace2000

peace2000:
Oh yeah. Sorry I got confused.

PR #5636, opened 2020-06-30
A: (gossip-room.fr) 3014e54

smed79:


peace2000:
I don't see this applying anymore.

PR #5650, opened 2020-07-01
A: fonecta.fi

peace2000:
Tracking, sample link: https://stats.fonecta.fi/log?s=fonectaweb&a=homepage&page=homepage&e=home&i=191349&ex=desktop&lng=fin&rnd=0.4969946427925994

Note, you must have an account on that site in order to see that address in your logger, but that address is accessible without an account. (I don't know if that sample link will expire later.) That link should show 1x1 gif image.


peace2000:
@monzta Any change to check this? It's bit old but merging ok (no conflicts) and it's easy to review.

PR #5657, opened 2020-07-02
A: is.fi

peace2000:
Tracking. Sample page: https://www.is.fi/formula1/art-2000006559945.html


ryanbr:
Does it need the $xml switch?

peace2000:
@ryanbr Not necessarily, I took it off.

PR #5658, opened 2020-07-02
A: www.tori.fi

peace2000:
https://www.tori.fi/

Note, I think it's better to retain prefix `www.` in this case because the word `tori` is a general Finnish word that could be used as a suffix in other domain names.

One example: http://keskustori.fi/ (could be more)


ryanbr:
Unless there is issues with subdomains under tori.fi, I don't understand why www isn't removed.  @peace2000

peace2000:
@ryanbr Sorry I had wrong conceptions. I removed www part.

PR #5669, opened 2020-07-05
A: https://www.opensubtitles.org/

peace2000:
Hides "Watch online" ad box. I had to concatenate two rules because I could not block that ad fully by using one rule.

opensub

Sample link: https://www.opensubtitles.org/en/subtitles/5639181/a-nightmare-on-elm-street-5-the-dream-child-en


peace2000:
deleted.

PR #5692, opened 2020-07-07
why isn't piwik.js being blocked?

fabianski7:


Khrin:
We already have https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_general.txt#L3187

Yuki2718:
@monzta That must be intentional, see https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_general.txt#L3199. If it's no more needed, better to remove the exception than adding a new rule.

PR #5693, opened 2020-07-08
yu.xyz.mn NSFW

Yuki2718:
URL: `http://yu.xyz.mn/`
Issue: an useful images (discount coupon) blocked

yu

yu2

Env: Brave 1.10.97 + uBO 1.27.10 default lists + AdGuard Japanese


Yuki2718:
Sorry I changed a wrong file.

PR #5728, opened 2020-07-15
A: http://www.fiches-auto.fr/

smed79:


peace2000:
Both of these entries are already in the list by now.

PR #5736, opened 2020-07-16
A: pantip.com

mishari:
Removed annoying PDPA notice


Khrin:
Hi, thanks for your suggestion! I only added instead in EasyList Cookie, which is included also in Fanboy's Annoynces List.

PR #5742, opened 2020-07-17
A: Google

peace2000:
Data privacy notification hide for Google and for it's tld's.

Sample link used: https://www.google.fi/

kuva


peace2000:
Not working anymore.

PR #5756, opened 2020-07-20
A: https://www.opensubtitles.org/

smed79:


DandelionSprout:
I can confirm the existence of such a filepath file on that site. However, I couldn't confirm the exact extent to which it causes ads to spawn.

PR #5759, opened 2020-07-20
A: (ettv.to & ettvdl.com) #5758

smed79:


ryanbr:
both domains dead, safe to close

PR #5787, opened 2020-07-23
A: https://malwaretips.com/blogs/browser-toolbar-removal/

smed79:


Khrin:
We prefer to use a specific filter to block this element: https://github.com/easylist/easylist/blob/master/fanboy-addon/fanboy_social_specific_hide.txt#L706

Issue #391, opened 2017-05-02
LaunchDarkly in privacy list breaks applications that use their service


dcousineau:
Recently launchdarkly was added to the EasyPrivacy list (see [commit here](https://github.com/easylist/easylist/commit/c858d871d2fa0d38ce0ea0db75fec89845d67ee0)).

LaunchDarkly is a third-party service the application authors (such as myself) opt-in to and pay for. It uses the concept of a user-id and attributes to return whether or not certain flags are enabled or not. We use it to roll out access to beta features in a controlled fashion before removal.

As it stands to use my own application I have to turn off EasyPrivacy and recommend all users turn off EasyPrivacy until either (a) we write and deploy a proxy to bypass this rule or (b) this rule is removed from the list. I would prefer not to instruct users to turn off the privacy list and frankly do not like turning it off for myself.

adamlaunchingdarkly:
This issue can now be closed based on a commit that restricts the scope of the listing (https://github.com/easylist/easylist/commit/0356394b6210143110e2226018aafbabb5acc3b7)

PR #5814, opened 2020-07-27
Remove app.launchdarkly.com block

oskarolsson-jimdo:
Hi!

It seems Launchdarkly was blocked in [this commit](https://github.com/easylist/easylist/commit/951cb820f215a981649bc53d29898133f6a1dffd).  

This block happened [once before](https://github.com/easylist/easylist/issues/391) and was rolled back in favor of only blocking [tracking](https://github.com/easylist/easylist/commit/0356394b6210143110e2226018aafbabb5acc3b7) via Launchdarkly.

In short Launchdarkly is a thirty party service one can integrate with to turn certain features on or off. It's not directly tied to ads or  banners and the tracking aspects are completely optional. The newly implemented block will definitely affect some services negatively, since they rely on this to e.g. roll out new features to a subset of users.

ping @ryanbr

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -265,7 +265,6 @@
 ||api.wipmania.com^
 ||apm-engine.meteor.com^$third-party,xmlhttprequest
 ||app.cdn-cs.com/__t.png?
-||app.launchdarkly.com^
 ||app.link^$third-party
 ||app.opmnstr.com/v2/geolocate/
 ||app.yesware.com/t/$third-party


jbyler:
As a user of LaunchDarkly, I agree with what @oskarolsson-jimdo says.  LaunchDarkly is not an ad service or a tracking service; it's a "feature flag" service with the primary use case of doing progressive rollouts.  The tag line is "Dynamically control the availability of application features to your users."  I think this PR should be approved and merged.

ryanbr:
Looking at this,

`https://app.launchdarkly.com/snippet/ldclient.min.js` A few event trackers/monitoring
`https://app.launchdarkly.com/sdk/evalx/58455d83460121090e244f2b/users/eyJrZXkiOiIwMDAwMDAwMC0wMDAwLTAwMDAtMDAwMC0wMDAwMDAwMDAwMDAiLCJhbm9ueW1vdXMiOnRydWUsImlwIjoiMzcuMjM1LjEwOC43OSJ9?h=ca06ac173e05cd741b9c1f64551a520e615f0da9e89851019167b38c7ea91109`  (More event trackers/monitoring)
`https://app.launchdarkly.com/sdk/evalx/599be38364af310acf2b3854/users/eyJrZXkiOiJiYTE1MDFiOS1jZjAxLTQ5YjQtOWU4NC0xOWRlYzY4ZTdiNzUiLCJhbm9ueW1vdXMiOnRydWUsImN1c3RvbSI6eyJ1c2VyQWdlbnQiOiJNb3ppbGxhLzUuMCAoV2luZG93cyBOVCAxMC4wOyBXaW42NDsgeDY0KSBBcHBsZVdlYktpdC81MzcuMzYgKEtIVE1MLCBsaWtlIEdlY2tvKSBDaHJvbWUvODQuMC40MTQ3Ljg5IFNhZmFyaS81MzcuMzYiLCJzY3JlZW5XaWR0aCI6MjU2MCwic2NyZWVuSGVpZ2h0IjoxNDQwLCJpc0FmZmlsaWF0ZSI6ZmFsc2UsImxhbmd1YWdlIjoiZW4iLCJjdXJyZW50VGltZXN0YW1wIjoxNTk1OTA1MjM4MzE0LCJ1c2VySGFzU2VlblByaWNlcyI6ZmFsc2UsImhhc05ld0NoZWNrb3V0U3VwcG9ydCI6dHJ1ZSwiY29va2llSWQiOiJiYTE1MDFiOS1jZjAxLTQ5YjQtOWU4NC0xOWRlYzY4ZTdiNzUifX0` (More event trackers/monitoring)

Taken from sample site: https://minetilbud.dk/tilbudsaviser/bordershop/1 https://www.vivaldichor.de/ and https://www.sevencorners.com/#start

Event trackers/monitoring should still be blocked.

twooster:
@ryanbr Can you explain how the situation is changed since the last time `app.launchdarkly.com` was pulled out of EasyList?

My understanding is that LaunchDarkly has split out the feature-flag app from the events domain, making it easy to block tracking without blocking app functionality. As far as I can tell from the code you linked, the JS from `app.launchdarkly.com` does have event tracking capability, but only if it can reach `events.launchdarkly.com`. Otherwise the event tracking should not function.

We're currently having to hot-patch our web-app, where we use LaunchDarkly to subset and trial new features, and handling a flood of customer tickets as a result of this block. Would love to come to an amicable solution. Perhaps we should cc in some LaunchDarkly folks to comment and maybe provide some remediation?

ryanbr:
I have temporarily removed the filter, the event monitoring/tracking **still** needs to be resolved.

Regardless what occurred historically, sites change, site reviews change. Every tracker and/or adserver could complain we "break too much", but it doesn't stop them from conducting their business.

I'll re-review this commit in a few months, and if things don't improve... I'm not a launchdarkly customer, and most people browsing the sites aren't either. So why should we (the enduser) care about it?

jo-sm:
@ryanbr I'm a bit confused by your response here - did you look at the actual request data, or assume that the opaque URL is tracking outright?

If you inspect the payload for `https://app.launchdarkly.com/sdk/evalx/...` requests, it's apparent that it's getting feature flag variation data. For example, this is part of the payload for one of the requests in our app:

```
{
    "new-navigation-styles": {
        "version": 1096,
        "flagVersion": 16,
        "value": false,
        "variation": 1,
        "trackEvents": false
    }
}
```

Besides the key and actual value object data, I'd be surprised if other sites have fundamentally different response payloads for these URLs. 

Contrast that to requests to `events.launchdarkly.com`. Requests to URLs like `https://events.launchdarkly.com/events/bulk/...` contains request data that looks like this:

```
{"default":null,"counters":[{"value":true,"count":1,"variation":0,"version":3}]}
```

And has no response payload. 

These two being separate are on purpose, so that the tracking that happens from `events.launchdarkly.com` is separate from the actual functionality of the product, and that hasn't changed anytime recently. With that in mind, I'd like to repeat the question that @twooster asked above again -- can you explain what's changed, with more detail?

jnolen:
Hi @ryanbr ,

This is Jonathan from LaunchDarkly, I run Engineering and Product here. We got quite a few support tickets on Monday from Easylist users about broken applications due to the appearance of LaunchDarkly on the blocklist. We followed this PR and are grateful that you removed LD from the blocking list. Normal behaviors seem to have been restored, but we want to make sure we can prevent this problem from recurring in the future.

LaunchDarkly is a feature flagging platform. Our customers use it to manage which features their end-users have access to. By blocking LaunchDarkly entirely, end-users will lose access to the features they expect to be there. An example might be features or plans that they have paid for, or new functionality that is in the midst of a progressive rollout. In the worst-case scenario, we’ve heard about end-users being completely unable to use a service they pay for since LaunchDarkly isn’t there to manage their access.

When we designed LaunchDarkly, we explicitly separated the delivery of flag values for each user (which are necessary for your users' apps to function properly) and the return of telemetry data about those flags. We maintain different hosts for this purpose:

* app.launchdarkly.com / api.launchdarkly.com / sdk.launchdarkly.com: the hosts we use to deliver flag values to the clients.
* events.launchdarkly.com: the host we use to receive data from clients

The evalx calls to app you mention above includes a client object that contains an ID as well as context information provided by the host application that is necessary to determine the set of flag rules for this client. That client or context object is only used to retrieve the flag rules payload and never stored.

The posts to events.launchdarkly.com record which flags were evaluated and which behaviors the host application chooses to record. Regardless of the host application, we respect the end users' Do Not Track setting and do not send analytics data in that case.

Applications like yours can block only events.launchdarkly.com to ensure no analytic data is returned from the client to us – while allowing the host application to function successfully as outlined by @twooster above. This is what we would recommend here.

If you have any questions or there is anything we can do to help clarify, please let us know. We’ve done our absolute best to honor the privacy of the end-users of our SDK. We’d be happy to hop on a screen share to show you the product and answer any concerns you might have.

Thanks for your time and consideration,

Jonathan Nolen (and the LaunchDarkly Team)
VP Eng & Product at LaunchDarkly

spirillen:
Hey @jnolen Let me ask you about these domains then, found them in the archive, are they 1. actively used, 2. what's there purpose. I'm asking as I can see from the id they are from an initial import.

|id|name|
|--|----|
|3504229|clientstream.intuit.launchdarkly.com.rpz.mypdns.cloud|
|3504230|app.intuit.launchdarkly.com.rpz.mypdns.cloud|

jnolen:
Hi @spirillen,

`clientstream.intuit.launchdarkly.com` and `app.intuit.launchdarkly.com` are domains actively used by us (as are `sdk.intuit.launchdarkly.com`, and `clientstream.intuit.launchdarkly.com`). They represent a customer-specific instance of LaunchDarkly, and correspond exactly to `clientstream.launchdarkly.com` etc.

But we don't recognize the `rpz.mypdns.cloud` domain.

spirillen:
Thanks for your reply @jnolen 

The `.rpz.mypdns.cloud` are coming from the PowerDNS authoritative server, which stores fully qualified domain records vs ex. bind9 which often uses the partials.

So used within a RPZ zone it becomes `$domain.$zone`.

Full example. To block `example.org` will become

`example.com.rpz.mypdns.cloud    IN CNAME .`

Hence the appended `$zone` in the table above.

There is a cheat cheat available here <https://www.mypdns.org/w/rpz_record_types/>

PR #5847, opened 2020-07-31
Remove Nation Media Group specific hide

roelfjan:
Hi! Will building a new website for Nation.co.ke (for the Nation Media Group), I encountered the issues that our footer (with classname `.footer`) was getting removed by ad blockers. Since this was a hide for previous versions of websites of the Nation Media Group, this can be removed.


ryanbr:
Sure, seems legit

PR #5863, opened 2020-08-01
A: https://www.latimes.com/science/story/2020-08-01/inside-the-clinical-trial-testing-the-healing-powers-of-blood-from-covid-19-survivors

llacb47:


llacb47:
I don't know if it is preferred to put this alphabetically, if so you can move it.

PR #5867, opened 2020-08-01
A: https://search.playgamenowtab1.com/

smed79:


ryanbr:
Can you add another entry in the Notifications list, under the `! also in Easyprivacy` heading @smed79

smed79:
Done!

PR #5875, opened 2020-08-02
A: https://besancon.onvasortir.com/

smed79:


Khrin:
I think it can be made as generic filter.

smed79:
There is already a generic filter https://github.com/easylist/easylist/blob/75bf7212a49a8c631ee095d85c7928bd7b2c598e/easylist_cookie/easylist_cookie_general_hide.txt#L4435
I added the filter to fix an _[old entry](https://hg.adblockplus.org/listefr/rev/0590f973b47b)_ `$generichide` rule found in **liste FR**.

PR #5924, opened 2020-08-08
Proof-of-concept of filesize reductions (or lack thereof) by removing " from entries that don't need them

DandelionSprout:
I noticed that EasyList is doing a ***huge*** summer cleaning this week, so I figured I should suggest something that hasn't traditionally been thought about much in public.

However, I think I only managed to save a grand total of 1KB (if even that), in a list of 3~4MB. Thus I presume this wouldn't even be worth to submit as a proper PR, but I felt the results were worth writing about nevertheless.


ryanbr:
Have you tested without the `""` in all the Adblockers? Adblock Plus, Ubo, Adblock etc. it could break compatibility with anyone parsing these rules.

DandelionSprout:
I've tested in Nano and uBO. I can double-check and test in *AdGuard for Windows* and Adblock Plus later this afternoon.

DandelionSprout:
Using https://example.org/ as the example page, I've got `example.org##a[href*=example]` and `example.org##a[href*=EXAMPLE i]` to work in AdGuard for Windows 7.4.2, ABP 3.9.4, and AdBlock 4.18.0, all of them on Vivaldi.

Yuki2718:
Also please make sure that all the block-allow matching are kept. I mean, e.g. if you changed `##div[id^="div-gpt-ad"]` to `##div[id^=div-gpt-ad] but left an allowlist `example.com#@#div[id^="div-gpt-ad"]` then this allowlist doesn't work. I've recently experienced this and fixed in my list.

DandelionSprout:
A quick look through https://github.com/DandelionSprout/easylist/blob/patch-7/easylist/easylist_allowlist_general_hide.txt, tells me that this draft isn't affected by that problem, thankfully.

PR #5944, opened 2020-08-11
A: https://www.hackingloops.com/phish-tracking-targets-hiddeneye/

smed79:


ryanbr:
Might be too generic, I would make this specific

PR #5952, opened 2020-08-12
A: aristeguinoticias.com

jabonsolo:


Khrin:
Wouldn't be better to have these filters added in Easylist spanish?

jabonsolo:
I was thinking about it, but as this is not to block an ad, it is actually to prevent a page to break, but if you consider it is better to add it there I'm ok with that.

Khrin:
Yes, if it's possible, also fixing filters for easylist should be added to supplemental filters lists

jabonsolo:
Added filters to EL spanish.
https://github.com/easylist/easylistspanish/commit/19435d1bedd53eb8078940c37a73634eb87acde8

PR #5965, opened 2020-08-14
A: tivi.fi

peace2000:
Alma media's cookie notification block for several websites, such as:

https://www.tivi.fi/
https://www.mikrobitti.fi/
https://www.arvopaperi.fi/

(There are more but those samples are probably enough.)


ryanbr:
can this be `||almamedia.fi/almacmp/$script` ?

peace2000:
@ryanbr Sure! It's edited now.

PR #5976, opened 2020-08-15
P: fix https://github.com/AdguardTeam/AdguardFilters/issues/61255 det…

Yuki2718:
…ail.chiebukuro.yahoo.co.jp

https://github.com/AdguardTeam/AdguardFilters/issues/61255

I can't reproduce the issue but four independent reporter (three in the thread and one in a Japanese community) have reported the same issue and three of them confirmed that this fixed the issue.


Yuki2718:
Thank you for quick response!

PR #5986, opened 2020-08-19
A: (nsfw) tits-guru.com

smed79:


easylist_adult/adult_specific_block.txt
```diff
@@ -802,6 +802,7 @@
 ||spankbang.com^*/ero_loader.
 ||sss.xxx/ps/
 ||static.kinghost.com^
+||static.tits-guru.com/js/istripper2.js
```
spirillen, line 805: I think this should be replaced with `static.tits-guru.com/js/istripper*.js` as I've just found the `static.tits-guru.com/js/istripper4.js`

I will even suggest ``tits-guru.com/*/istripper*.js` simply to be on forehand of relocations/renaming. However I even believe it should be possible by only doing `istripper*.js$script` as I only located this js script in relation with the `istripper.com` pop-ups

PR #6002, opened 2020-08-24
A: Address https://old.reddit.com/r/uBlockOrigin/comments/ied4x6/ads_…

Yuki2718:
…appear_on_google_search/

URL: `https://www.google.com/search?client=firefox-b-d&q=super+smash+bros+ultimate` (search "super smash bros ultimate" at Google top page)
Issue: missed ads, as reported in `https://old.reddit.com/r/uBlockOrigin/comments/ied4x6/ads_appear_on_google_search/` and `https://forums.lanik.us/viewtopic.php?p=156198#p156198` (the same reporter). ~~Now I can reproduce with US and German IPs~~ EDIT apparently IP is irrelevant, I can reproduce with whatever IP/locale:

google1

google2

Env: Firefox 79.0 + uBO 1.29.2 default lists

Note: I used the same class name in the thread as it seems to be constant, but what's unique to the ads is `mf8UVb` so this only.


Yuki2718:
I found the ads also on local version of Google:

google

PR #6026, opened 2020-09-07
P: Add two domains to ###pb_block (NSFW)

Yuki2718:
`tube8.com`

<details>

tube8
</details>

`redtube.net`
(Covered by uBlock filters so I disabled the rule for this SS.)
<details>

redtube
</details>

You need to play any video and pause to reproduce.


Yuki2718:
It may be better to make the rule generic.

Yuki2718:
I don't see `#pb_block` on tube8 any more.

PR #6062, opened 2020-09-15
P: www.sony.jp can't switch tabs

Yuki2718:
URL: `https://www.sony.jp/`
Issue: You can't switch tabs on the page:

sony

Env: Firefox 80.0.1 + uBO 1.29.2 default lists

Note: another fix is needed on uBlock Unbreak to allow tiqcdn.


Khrin:
Is it necessary the `www.`?

Yuki2718:
> Is it necessary the `www.`?

Asked the same in uBO issue, it's more of my preference to make whitelists as precise as possible if the domain has other subdomains than www. But I don't see requests to adobedtm in other subdomains of sony so will fix it.

PR #6064, opened 2020-09-15
P: https://www.tennisassa.fi/

peace2000:
Access to the website is blocked, added a fix.

https://www.tennisassa.fi/


peace2000:
I don't want too much to rush anyone, but as this is a critical issue, I hope this would be checked as quickly as possible. @Khrin

peace2000:
I don't want too much to rush anyone, but as this is a critical issue in terms of filter issues, I hope this would be checked as quickly as possible. @ryanbr ?

PR #6065, opened 2020-09-16
A: warkaudenlehti.fi (generic Easylist hide)

peace2000:
I think this would do as a general rule. 

Sample link: https://www.warkaudenlehti.fi/uutissuomalainen/3106612

kuva


ryanbr:
Not sure if `##div[class~="card--native"]` isn't good filter, not a very adlike element. @peace2000

peace2000:
@ryanbr That element which is hid on warkaudenlehti.fi by using that filter, is a native ad. That's why I suggested this filter as a general one.

(Word "mainos" means "an ad" if translated.)

PR #6072, opened 2020-09-18
A: https://post.tv-asahi.co.jp/post-128983/ anti right-click, copy

Yuki2718:
Right-click, copy, drag etc disabled.


ryanbr:
`||tv-asahi.co.jp/wp-content/themes/favclip-master/js/front.js`

Would be a better filter, if that is what you're aiming

Yuki2718:
@ryanbr  Can you please add the rule on your side? Since mine is days old, fixing it on my end would cause conflict.

spirillen:
Just a thought :thought_balloon: 

Wouldn't it be better to make it a more global rules as

```
/wp-content/themes/favclip-master/js/front.js
```

To prevent it from all similar sites (themes)?

ryanbr:
Sure, okay.

PR #6075, opened 2020-09-18
update filter

llacb47:
akamai fingerprinting not being blocked on `https://www.pepperfry.com/`

`https://www.pepperfry.com/public/77b550f9eno2052d455ab5f13646afa`

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -5334,7 +5334,7 @@ cgi-bin/counter
 /pg*.js?PID=$xmlhttprequest
 ! Akamai fingerprinting
 ! https://publicwww.com/websites/%22_cf.push%22/
-/^https?:\/\/.*\/(public|resources|static|assets)\/([a-f0-9]){28,30}$/$script,~third-party,xmlhttprequest,domain=~cand.li
+/^https?:\/\/.*\/(public|resources|static|assets)\/([a-o0-9]){28,31}$/$script,~third-party,xmlhttprequest,domain=~cand.li
 ! Cloudflare fingerprinting
 /cgi-bin/bm/cv/*$script
 ! Admiral


Yuki2718:
@llacb47 See mine https://github.com/Yuki2718/adblock/blob/master/japanese/jp-filters.txt#L5154, 31 is not enough, there's a case of 32 letters (not confirmed if [a-o] is enough, maybe latter). Probably you can check as I added domain=. There are still the same Akamai staff I couldn't include in regex, easily spotted in my list by their random-looking.

llacb47:
Closing in order to apply fixes mentioned above

darioseidl:
Hi! This filter rules in "uBlock filters – Privacy"

```
! https://github.com/easylist/easylist/pull/6075
/^https?:\/\/.*\/(public|resources|static|assets)\/([a-f0-9]){28,30}$/$script,1p,xhr,domain=~cand.li,badfilter
/^https?:\/\/.*\/(public|resources|static|assets|utilities)\/([a-z0-9]){28,33}$/$script,1p,xhr,domain=~cand.li
```

Is blocking some legitimate XHR requests on our website, for example, one called `/public/fetchCompanyRegistrationTemplate` among others. This filter seems to be very generic, anything starting with public or assets and between 28 and 30 characters being blocked? Shouldn't this be restricted to some specific domains?

Yuki2718:
@darioseidl URL?

darioseidl:
For example here: https://www.bus.busworld24.com/partner/registration when clicking on "Register now as bus operator" we fetch a text template with the request `/public/fetchCompanyRegistrationTemplate`. The text currently says "This text was blocked by UBlock Origin" for demo purposes and should be shown on the bottom of the page, but is not shown when uBlock with this filter rule is active.

Anyway, this is just one of many examples. I'm sure there are countless valid requests that start with `public|resources|static|assets` and happen to be between 28 and 30 characters long that should not be blocked.

Yuki2718:
@darioseidl Thank you for the detailed info. I'll disable the filter on uBlock and we'll discuss what we can do.
@ryanbr Probably needs to be disabled on EP too.
@Alex-302 AGTPL too.

Yuki2718:
@llacb47 I believe Akamai BM's `(public|resources|static|assets)` part always follows directly after TLD. Do you think this is enough?

darioseidl:
Thank you for the quick response!

Yuki2718:
@ryanbr @Alex-302 I can reproduce `https://www.reddit.com/r/uBlockOrigin/comments/k9a5mh/costcocom_login_failure_with_ublock_origin/gf73zd0/`. I think the filter should be disabled ASAP. (already badfiltered on uAssets)

Alex-302:
@Yuki2718 Which rule? Missed this discussion:( I don't see `public|resources|static` in filters.

Yuki2718:
@Alex-302 `/\/(public|resources|static|assets)\/([a-f0-9]){28,31}$/$script,xmlhttprequest,~third-party` in TPL. (Un)fortunately Costco changed the script's path to "utils" so now it's not blocked, but if you block you'll get login error as reported in Reddit. This Akamai BM script is deployed widely all over the world, if you need I can provide a dozen of examples most of which are shopping, banking, or other sensitive sites. Who knows what else has similar problem.
BTW thanks ryanbr for disabling the script. I think the only safe way to keep the filter is to add domains one by one on which blocking doesn't cause any issue.

Alex-302:
got it

PR #6081, opened 2020-09-22
A: www.365businessfinance.co.uk

FozzieHi:
A: www.365businessfinance.co.uk cookie popup


ryanbr:
Why `www.365businessfinance.co.uk` and not `365businessfinance.co.uk` @FozzieHi

FozzieHi:
@ryanbr The root domain just seems to redirect to the www domain. This is my first time making a rule so please excuse my ignorance!

PR #6090, opened 2020-09-26
A: ampparit.com (generic Easylist hide)

peace2000:
Generic filter.

kuva

https://www.ampparit.com/


ryanbr:
It's been made specific, rather a generic element @peace2000

PR #6107, opened 2020-10-01
fimg-resp.seznam.cz

vasimi:
`fimg-resp.seznam.cz` is just an image storage, it isn't involved in any user tracking


ryanbr:
Example site where it's breaking @vasimi  ?

vasimi:
@ryanbr https://search.seznam.cz/?q=andrej+babis

spirillen:
Why not specify it a bit more to enhance the rule as whitelist to prevent bad stuff from happening?

```
! 2020-10-11 seznam.cz
@@||fimg-resp.seznam.cz^$1p,image,domain=search.seznam.cz
```

vasimi:
@spirillen allowing this only for search.seznam.cz doesn't make sense, there are other domains that may use this storage, but I'll update the PR with the image filter

spirillen:
> @spirillen allowing this only for search.seznam.cz doesn't make sense, there are other domains that may use this storage, but I'll update the PR with the image filter
> 
> update: the filter you suggested doesn't work (it still blocks all the images from fimg-resp.seznam.cz) and I'm not able to make it work without removing the whole domain from the list. However, the only purpose `fimg-resp.seznam.cz` is to serve images, it's not capable of tracking users and/or invading their privacy, so IMHO, it's totally safe to just remove the whole domain from the list.

The though is/was to be a step ahead of what ever the future might bring :bulb: no more, no less

But weird the rule didn't work for you, it did for me. Could you give this variant a spin?

```
@@||fimg-resp.seznam.cz^$image,domain=seznam.cz
```

Removed the `1p`

vasimi:
@spirillen Works for me, I've updated the PR.

vasimi:
@ryanbr hey, could you please take a look at this and eventually merge it?

PR #6124, opened 2020-10-09
Add new google ads carousel match format.

jameshilliard:


easylist/easylist_general_hide.txt
```diff
@@ -21271,6 +21271,7 @@
 ##.l-container > #fishtank
 ! Google
 ###atvcap + #tvcap > .mnr-c > .commercial-unit-mobile-top
+###atvcap > div:has(h3[role="heading"] > span:has-text(/^Ads/))
```
krystian3w, line 21274: - `:has-text(foo)` and `:has(foo)` is no supported by ABP/AdBlock

- omitted domains is no supported by ABP/AdBlock for procedural filters (like `:-abp-has(foo)`)

   - also no supported simplify procedural filters to only TLD like `com##...`

krystian3w (changes requested):

krystian3w:
Filter need move to other file, if no possible remove useage `:-abp-has()` and `:-abp-contains()`

It would also be appropriate to add around ~220 google domains with their various public suffixes ("TLD") before `#?#`.

krystian3w (changes requested):

jameshilliard:
Think it's safe to unconditionally block `#atvcap`?

ryanbr:
We already have `###atvcap + #tvcap > .mnr-c > .commercial-unit-mobile-top` (which #atvcap would overrule.) How safe is blocking #atvcap outright?

jameshilliard:
> How safe is blocking #atvcap outright?

Well the class name sounds to me like it's ad specific and not a random div name, but hard to say for sure.

PR #6267, opened 2020-10-22
M: Remove new Instagram cookie wall

Forage:


peace2000:
This will break scrolling, unfortunately.

Forage:
Darn. Sorry, I don't have an account so I did not test that. It appears to be working fine without logging in, but I realize that's not sufficient.
Let me know if you want me to delete the change.

krystian3w (changes requested):

krystian3w:
Better close, EasyList Cookies can break page if try block pop-up with `$csp=...`

krystian3w:
@Forage disable filter from `uBO Annoyances` when you should locked scrolling.

PR #6282, opened 2020-10-25
A: (nsfw) https://git.io/JTwhs

smed79:


DandelionSprout:
I can confirm that the overlay is indeed present on video pages.

PR #6283, opened 2020-10-25
A: https://downdetector.com/status/youtube/

smed79:


Yuki2718:
https://forum.adguard.com/index.php?threads/resolved-downdetector-it-problem-with-disqus.37852/
IDK if this still matters.

DandelionSprout:
I am unable to reproduce any difference in ad elements with or without `/adscript.`, but I like the idea of a generic blocking rule like that.

PR #6393, opened 2020-11-06
A: (rustica.fr) #6392

smed79:


ryanbr:
used for fontawesome, would cause issues

smed79:
Ok, thank you.

PR #6490, opened 2020-11-21
A: Update mining domainds

smed79:
https://github.com/hoshsadiq/adblock-nocoin-list/issues/410

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -2821,6 +2821,8 @@
 ||35.239.57.233^$third-party
 ||45.32.105.134^$third-party
 ||77.162.125.199^$third-party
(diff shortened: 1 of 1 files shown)


spirillen:
Hi @smed79 

After looking at the domain names of you submission, wouldn't it be more right that I would add them to a [coin-miner](https://www.mypdns.org/project/view/27/) list rather than actually classify them as [TrackWare](https://www.mypdns.org/w/trackware/) for Cryptojacking?

smed79:
@spirillen Datamining use scripts to discover what "features" your browser has.

@ryanbr Correct me if im wrong.

spirillen:
The reason I asked is there could be a misunderstanding between datamining vs coinmining, have seen cases where the terms are mixed :wink: 

So to clarify, these are **DATA-miners**, not COIN-miners, correct?

PR #6601, opened 2020-12-04
A: (nsfw) playvids.com

smed79:


spirillen:
Related to this domain, not your rule, as I've tested with and without.

Videos are not played (Tor-browser)

image

smed79:
> Videos are not played

No issue here `https://ibb.co/HLcbq26`

PR #3462, opened 2019-05-20
A: cdn.granify.com

rsadwick:
Adding cdn.granify.com to privacy specific

PR #6662, opened 2020-12-10
Removed incorrectly added Granify

richardhodsdon:
Originally added in #3462

### Filter affected: 
Privacy Filter

### 1st Party sites affected: 
[www.hsn.com](www.hsn.com), plus many other clients

 
### How is it broken? 
Our integrated solution is now blocked from loading. The our initial script is blocked.

### Description why it should be removed: 
We run a 1st Party integrated system with our clients. Granify provides valuable site functionality above and beyond "ads". We're internal messaging for the client itself, and being on this ad blocking list is harming the legitimate customer user experience.

Secondly the original site referenced above in #3462 is no longer a client and we are not running on the mentioned site: https://shop.islandcreekoysters.com/

Thanks

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1054,7 +1054,6 @@
 ||gostats.ro^$third-party
 ||gotrk.net^$third-party
 ||govmetric.com^$third-party
-||granify.com^$third-party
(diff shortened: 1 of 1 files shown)


richardhodsdon:
Hi @spirillen this has been a while and looking for some help to bump this PR up?

PR #6720, opened 2020-12-17
A: (nsfw) https://tuberel.com/ (fixes https://github.com/AdguardTeam/…

Yuki2718:
…AdguardFilters/issues/70112)
https://github.com/AdguardTeam/AdguardFilters/issues/70112
Also fixes https://github.com/AdguardTeam/AdguardFilters/issues/69956

PR #6733, opened 2020-12-20
M: update #6720

Yuki2718:
https://github.com/easylist/easylist/pull/6720/
https://github.com/AdguardTeam/AdguardFilters/issues/70112
https://github.com/AdguardTeam/AdguardFilters/issues/69956
Also replaced `||tubev.sex/static/js/abb_test.js` with the current form. Not sure poppy is used on pornj, but kept it just in case.


spirillen:
Nice work, thanks :+1:

PR #6789, opened 2020-12-21
A: https://hackernoon.com/

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=45520

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -1046,6 +1046,7 @@
 ||informer.yandex.ru^
 ||infosniper.net/locate-ip-on-map.php
 ||infusionsoft.com^*/getTrackingCode?
+||ingest.sentry.io^
 ||innogamescdn.com/media/js/metrics-
 ||inphonic.com/tracking/
 ||inq.com/tagserver/logging/


jan-auer:
I would like to move for reverting this rule. Sentry is not a third-party tracking service, it reports **bugs** to developers and provides extensive tools to remove all personal identifiable information (PII) from data that is being submitted. 

There has been prior discussion on the forums on this, including:
- https://forums.lanik.us/viewtopic.php?p=99380#p99380
- https://forums.lanik.us/viewtopic.php?f=64&t=43332&p=149006&hilit=sentry#p149006

If hackernoon is sending tracking information to Sentry, please add the offending domain like is done [here](https://github.com/easylist/easylist/blob/75faaea2e1f26b946ee98f6d51871ec6c20527a8/easyprivacy/easyprivacy_allowlist.txt#L72).

hyperknot:
If anything on Hackernoon get's sent to Sentry, it is because they have or had an error in their Javascript implementation. Today nothing gets sent there. It has nothing to do with tracking.

chaudois:
Can someone explain to other beginners like me why did sentry get listed in the first place? It looks like an attempt to solve a very minor issue on a very limited number of cases, that causes massive drawbacks for industries that are using sentry to log and debug their front-end applications, which is... realy bad, especially in this case, where we, the devs, cannot control if our users have an adblocker installed or not.

PR #6797, opened 2020-12-22
M: fix blocking of Matomo docker images

Findus23:
The current filter rules cause https://hub.docker.com/r/crazymax/matomo/ to show a 404 as the site loads its content from https://hub.docker.com/v2/repositories/crazymax/matomo/ which is blocked by the `/matomo/` rule.


Findus23:
Hi everyone,

Is there anything I can do or help to get this merged?
The issue still occurs due to the rule being too general.

PR #6803, opened 2020-12-23
M: fix blocking of Wikimedia image content

inductiveload:
See: https://forums.lanik.us/viewtopic.php?f=64&t=45509


llacb47:
This will not allow images on wikipedia, wikiquote, wikiversity, wikibook, etc. @ryanbr 

Suggest to change the rule to `@@||upload.wikimedia.org^$image` ?

inductiveload:
Looks like it's been fixed by another commit

PR #6848, opened 2020-12-27
A: https://linuxconfig.org/

smed79:


smed79:
@Khrin example `https://linuxconfig.org/images/banners/linuxcareers-get-alert_optim.png`

PR #6886, opened 2021-01-03
A: veoh.com (generic 3p cookie block)

peace2000:
Sample link: https://www.veoh.com/list/movies/movies_animation


peace2000:
Still relevant. Though, maybe this could be reduced to `js/gdpr/` ?...

PR #6899, opened 2021-01-06
A: (nsfw) freshscat.com

smed79:


spirillen:
May I suggest an append:

```
www.freshscat.com##.happy-sidebar
```

spirillen (approved):

PR #6992, opened 2021-01-14
A: https://www.thinksign.com/

smed79:


fanboy-addon/fanboy_social_general_hide.txt
```diff
@@ -12734,6 +12734,7 @@
 ##[aria-label="Share on facebook"]
 ##[aria-label="Share on linkedin"]
 ##[aria-label="Share on twitter"]
+##[aria-label="Social Bar"]
```
chirag127, line 12737: Are these rules case sensitive?

PR #7008, opened 2021-01-15
unblock hubspot.com widget

JLarky:
according to https://forums.lanik.us/viewtopic.php?p=129155 and https://knowledge.hubspot.com/reports/what-is-the-hs-scripts-embed-code-loading-on-my-website the domain `hs-scripts.com` is OK in terms of privacy. 

It seems that f70f0a77afb598ec5d5584b6b57fa3ae8d30855d was done without too much concern for that even though it even links pastebin with this exact information that this domain/script is OK :)

it's the first line here https://pastebin.com/aFiXQEq1

p.s. I have no affiliation with hubspot I was just trying to include its widget on my website and it was blocked by uBlock

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1139,7 +1139,6 @@
 ||hotjar.io^$third-party
 ||hotlog.ru^$third-party
 ||hs-analytics.net^$third-party
-||hs-scripts.com^$third-party
 ||hsadspixel.net^$third-party
 ||hscta.net^$third-party
 ||htlbid.com^$third-party


ryanbr:
No, it's not "OK". if you need a widget, implement it like any other js or html. Don't need hubscripts

Even comparing the current script to new script from https://github.com/easylist/easylist/commit/f70f0a77afb598ec5d5584b6b57fa3ae8d30855d  nothing has changed.  Embbedded javascript other than linking, has privacy issues.

JLarky:
I don't see any reason in what you are saying :) why don't you ban intercom widget then? :) or zendesk?

PR #7013, opened 2021-01-16
A: https://www.eltiempo.com/justicia/investigacion/margarita-cabello-…

isacurad:


Khrin:
Why this exception is needed?

isacurad:
The filter ##.social-col hides an empty element (the ads are blocked) that makes the main content container move to the left (it should be in the space marked in blue), so a fixed element (marked in red) overlaps the content.
image

PR #7023, opened 2021-01-17
A: prisma.fi (gdpr block)

peace2000:
Cookie script.

Sample sites:
https://www.prisma.fi/
https://www.alepa.fi/
https://www.sokoshotels.fi/
https://www.foodie.fi/
https://www.raflaamo.fi/fi


peace2000:
Still relevant

PR #7025, opened 2021-01-18
M: generic cookie hide filter expansion

peace2000:
Reason: "_gdpr-cookie-consent_" can be used as part of another cookie banner's name.

Sample link: https://www.finlandiatalo.fi/

Name of the cookie banner element on that site:
_##[id="db-gdpr-cookie-consent"]_

I don't know what "db" stands for and I don't think that it's necessary to make a separate filter for it. I think it's better to expand an existing filter. I think that  _##[id*="gdpr-cookie-consent"]_ is safe enough and doesn't produce false positives.


peace2000:
Still relevant.

PR #7030, opened 2021-01-19
M: expansion of 3rd party cookie filter

peace2000:
I think that this filter could be expanded so that it completely covers sites like this: https://www.fazer.fi/ There is now one unblocked cookie script:

kuva


peace2000:
I'll retract this, my expansion suggestion causes issues. E.g. images get blocked for some sites, such as: https://kuntoplus.fi/laihdutus/energiankulutus

PR #7040, opened 2021-01-21
A: seura.fi (generic ad hide rule)

peace2000:
Sample site: https://seura.fi/

kuva

("mainos" means "an ad").

Used widely:

https://publicwww.com/websites/%22sponsored-article%22/

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -19465,6 +19465,7 @@
 ##.sponsored-ad-ob
 ##.sponsored-add
 ##.sponsored-ads
+##.sponsored-article
 ##.sponsored-article-item
 ##.sponsored-b
 ##.sponsored-block


peace2000:
Still relevant

Yuki2718:
@peace2000 I think it's now obsolete, right? The rule is hiding the whole article of `https://www.tomshardware.com/desktops/mini-pcs/aoostar-mini-pcs-bring-elite-ryzen-power-in-a-tiny-footprint-for-under-usd400` so at least an exception will be needed, but I'm wondering if we can remove the rule.

peace2000:
@Yuki2718 

At least it's relevant on this site: `https://ng.se/stockholm`

Personally I would not remove, but that's up to Easylist maintainers to decide how readily old generics should be removed.

An easy way to fix that rule and globally prevent blocking body element:

`##.sponsored-article`
->
`##.sponsored-article:not(html, body)`

Yuki2718:
If it's only that site, I think it's better suited for regional list, either specific or generic. I'll remove the rule.

Yuki2718:
@lassekongo83 Sorry, can you please check the mentioned site?

lassekongo83:
I already have region specific filters for ng.se if that's what you mean.

PR #7051, opened 2021-01-24
A: puuilo.fi (generic cookie block)

peace2000:
A generic cookie blocking rule.

Sample page: https://www.puuilo.fi/

kuva

Used widely:
https://publicwww.com/websites/Amasty_GdprCookie/

Edit, also added whitelisting for `eckeroline.fi` because blocking this script causes some content not load on the home page.


peace2000:
Still relevant

ryanbr:
Include a wildcard at the end

peace2000:
@ryanbr Ok, added.

PR #7111, opened 2021-01-30
A: https://zaq2.pl/

audrey-dogman:


ryanbr:
English list, and we limit what anti-adblock is used due to legal reasons

PR #7152, opened 2021-02-03
Update easyprivacy_general.txt - removed an entry

yippeykeiyay:
Removed `/timingcg.` because that's the filename we use at RapidSpike for our Real User Monitoring detection. We don't serve ads from it and we don't record any PII data. The script is purely for recording timing metrics, browser, device and estimated location of the user and is installed by the website owner. We also use it for identifying data being sent to untrusted hosts in order to help protect the website's owners and its users from Magecart attacks.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -4285,7 +4285,6 @@
 /tide_stat.js
 /timeslog.
 /timestrend_
-/timingcg.
 /tiwik.
 /tjp_beacon.
 /tjx-tracking-


Yuki2718:
Did you read the template?
```
Easyprivacy requests:
** If a site implements any tracking or monitoring, UA/IP/Geo checks, browser detection, analytics, telemetry, linking to third-partys, pixels, referrers, fingerprinting, event/perf logging etc. Regardless how helpful or needed the script(s) are, it will be blocked in Easyprivacy. Privacy comes first and the block on these scripts will remain in place.
```
Ofc most of tracker/analytics do not record PII in straight sense.

PR #7292, opened 2021-02-26
ethereumads.com || https://www.mypdns.org/T3970

spirillen:
Found through https://www.mypdns.org/T3969

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -7753,6 +7753,7 @@
 ||theodorus-flo.com^
 ||theonecdn.com^
 ||therejockeryfu.fun^
+||ethereumads.com^
 ||thergoldru.club^
 ||thernouverge.site^
 ||thesister.club^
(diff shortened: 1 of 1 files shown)


ryanbr:
If its a legit site, making it $third-party would be preferable @spirillen

spirillen:
> If its a legit site...

:smirk: wouldn't that depend on who you asks?? :smirk: And how should we bw able to determine what is a Legit site, where are (if) the company registered etc.

If you would like me to add `$third-party` I'll do it, but on the other hand, or also in time noticed a lot of other noticed a number of "legit" adware servers/sites that don't have the `$third-party`. 

Non the less it's the same to me, they are blocked in my rpz...

ryanbr:
(and move it the third-party section also..) @spirillen 

Most, if not all non-third-party aren't hosting a homepage.  which is why `^` was preferred in that case.

spirillen:
> (and move it the third-party section also..) @spirillen
> 
> Most, if not all non-third-party aren't hosting a homepage. which is why `^` was preferred in that case.

Does it look more right for you now??

PR #7317, opened 2021-03-02
A: https://www.intel.cn/content/www/cn/zh/support.html

Yuki2718:
cname tracker of oracle:

<details>

intel

</details>

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -621,6 +621,7 @@
 ||ehow.com/services/jslogging/log/?
 ||ekg.riotgames.com^
 ||elance.com^*/emc.php?
+||elqtrk.intel.
 ||elsevier.com/pageReport?
 ||email-tickets.com/dt?e=PageView
 ||email.aol.com/cgi-bin*/flosensing?


easyprivacy/easyprivacy_specific.txt
```diff
@@ -621,6 +621,7 @@
 ||ehow.com/services/jslogging/log/?
 ||ekg.riotgames.com^
 ||elance.com^*/emc.php?
+||elqtrk.intel.
```
spirillen, line 624: Hey @Yuki2718 I do understand your intention and even tempted, but even I would make this rule in the DNSdist, I see to many risks of FP's.

Yuki2718: So what FP?

spirillen: If I knew up front, I would have given it, but only check against 2 matches `^elqtrk\.intel\.` just seems to be to wide/unsafe. On the other hand, I can give it a spin, and see how many hits it might get..

As said, I do follow you here :smiley: and even back you up on your idea, but i would prefer (personal opinion) it would check against 3 matches `^elqtrk\.intel\.(TLD1|TLD2|or something else)`

But I'm not the judge, I'm just spreading my thought, concerns

Issue #7331, opened 2021-03-05
speedostream.com


ghajini:
### List the website(s) you're having issues:

`https://speedostream.com/embed-f8kj1y6wwciw.html`

Embed link came from
`https://yomovies.mx/asli-rakhwala-natakam-2021-hindi-dubbed-Watch-online-full-movie/`

### What happens?

Push notification ads
![Screenshot_2021-03-05-16-09-10-618_org mozilla fenix](https://user-images.githubusercontent.com/20338483/110108564-5ce44500-7dd2-11eb-9cb0-76bcaf7723cc.png)

### List Subscriptions you're using:

Ublock origin default lists

### Your settings

- OS/version: android
- Browser/version: Firefox 88
- Adblock Extension/version: ubo 1.33.2

### Other details:

 certain filters help
```
||edxlewni.com^
||tgvbwjbnxz.com^
||adactioner.com^
||achilles-ena.com^
||coderformylife.info^
||notifypicture.info^
||a.speedostream.com/api/spots/*$script,first-party
```

spirillen:
Tested in FF and no hits on any of the OP's domains

<details><summary>Logger output FireFox</summary>

| | | | | | | | |
|:--- |:--- |:--- |:--- |:--- |:--- |:--- |:--- |
| +44 |  |  | speedostream.com | 1 | xhr | `https://s9.speedostream.com/hls/i7dmpqarqxb7o2nqgcsflp44ombvv44cpx4nvqca253iute5kipszthdbg3q/seg-3-v1-a1.ts` |
| +39 |  |  | speedostream.com | 1 | xhr | `https://s9.speedostream.com/hls/i7dmpqarqxb7o2nqgcsflp44ombvv44cpx4nvqca2sviwte5kip4styzpmaq/seg-2-v1-a1.ts` |
| +31 |  |  | speedostream.com | 1 | xhr | `https://s9.speedostream.com/hls/i7dmpqarqxb7o2nqgcsflp44ombvv44cpx4nvqca253iute5kipszthdbg3q/seg-2-v1-a1.ts` |
| +31 |  |  | speedostream.com | 1 | xhr | `https://s9.speedostream.com/hls/i7dmpqarqxb7o2nqgcsflp44ombvv44cpx4nvqca253iute5kipszthdbg3q/index-v1-a1.m3u8` |
[...]

PR #7335, opened 2021-03-05
speedostream.com | https://www.mypdns.org/T4272

spirillen:
From: https://www.mypdns.org/T4272

Related to: https://github.com/easylist/easylist/issues/7331


ryanbr:
Not an adserver, `easylist_specific_block.txt` @spirillen

PR #7337, opened 2021-03-06
P: (nsfw) https://www.freexcafe.com/erotica/femjoy/gallery1j5d6/pics/…

Yuki2718:
…photo2.php (fixes https://github.com/AdguardTeam/AdguardFilters/issues/76760)
https://github.com/AdguardTeam/AdguardFilters/issues/76760
Partly revert my bad commit https://github.com/easylist/easylist/commit/321370e826fd6a460a16bb42d28b953d1f8cebf6


ryanbr:
Whats the reason for revert, why is it bad? @Yuki2718

Yuki2718:
@ryanbr In the linked issue the photo is also hidden - surely clicking the photo leads to the tracker, however, hiding the photo will not be the desired result for the visitor of the site (and thus reported as Incorrect blocking). At the time of the commit I missed there's already specific filter, so I just made the filter specific.

Yuki2718:
To see the new ads, don't directly visit - only seen if you're navigated through the top page.

PR #7354, opened 2021-03-09
P: api(1,2,3...).fox.com

nathannaofficial:
Prevents breakage of fox articles, knowledgeable and custom support center features. Not sure how though.


ryanbr:
To generic to implement, need a test url to see it in action.

PR #7356, opened 2021-03-09
P: accounts.google.com (third-party)

nathannaofficial:
On some non-google websites the account login banner is broken. This is what it should look like for refrence.

Screen Shot 2021-03-08 at 11 55 37 PM


ryanbr:
Tested in medium and nytimes, we're not targeting accounts.google.com, no need to fix

PR #7360, opened 2021-03-09
P: purr.nytimes.com/v1

nathannaofficial:
Used to speed up article fetching on nytimes.com, takes longer to load otherwise.


ryanbr:
Seems fine for me, its just as quick.

nathannaofficial:
Okay. :)

PR #7362, opened 2021-03-09
P: yahoo.com/px.gif

nathannaofficial:
Yahoo Mail Beacon - Checks If Is Wifi To Reload Inbox (Non-Malicious) Could break yahoo.com mail icon or mail.yahoo.com itsself.


ryanbr:
Yes, its blocked. But its well tested, and works for me.

nathannaofficial:
ok

PR #7363, opened 2021-03-09
P: ...videodelivery.net

nathannaofficial:
Used to play video files, kind of like a vimeo or youtube but built in on websites. First detected on cloudflare.com's customer reviews page.

Videodelivery.net - Scripts
Embed.videodelivery.net - Iframes


ryanbr:
Sample site where its broken. And we're only blocking the stats bit for the domain

nathannaofficial:
https://www.cloudflare.com/case-studies/

ryanbr:
Videos played fine here from https://www.cloudflare.com/case-studies/ @nathannaofficial  

I would say like the previous ticket, if you want to be tracked, just unsub from Easyprivacy

PR #7447, opened 2021-03-17
P: datakauppa.fi (cookie banner hiding breaks the page)

peace2000:
The page cannot be used without a whitelist rule. Both rules are needed - the other for the banner to be visible, other for giving an access to cookie settings ("Evästeasetukset" in Finnish).

Sample link: https://www.datakauppa.fi/tuote/deltaco-uh-408-usb-2-0-hubi/


Nojuuu:
@peace2000 
Was it just a broken scrolling?

peace2000:
@Nojuuu I think so. This is pretty old PR, I think I made this before Easylist Cookie started to allow uBO specific fixes.

for ABP whitelisting is needed but as ELC allows now uBO specific fixes for this, they can be made.

Nojuuu:
Ok, I can look at this later when I have time.

Issue #4629, opened 2020-01-05
easylist cookie breaks sourceforge


momala454:
### List the website(s) you're having issues:

`https://sourceforge.net/projects/mpcbe/files/latest/download`

### What happens?

download doesn't start, timer doesn't count down

### List Subscriptions you're using:

easylist fr + easylist cookie

### Your settings

- OS/version: 
- Browser/version: 
- Adblock Extension/version:

Khrin:
Downloads starts for me. What browser/adblocker are you using?

momala454:
firefox
ublock origin

i think this filter is causing the problem 
||consentmanager.mgr.consensu.org^$3p

momala454:
i also have list uBlock filters and uBlock filters - unbreak active

spirillen:
`consensu.org` is a godaddy hijacked domain...

https://gitlab.com/my-privacy-dns/matrix/matrix/issues/271#note_267825322

momala454:
`https://consentmanager.mgr.consensu.org/`
is not displaying godaddy website

spirillen:
> `https://consentmanager.mgr.consensu.org/`
> is not displaying godaddy website

Weird, it does to me in the tor-browser as the screenshot shows

But your right, this is what I sees today:

<details><summary>Screenshot</summary>

![image](https://user-images.githubusercontent.com/44526987/71919084-166e6880-3184-11ea-9dc2-fcc068f3cd85.png)

</details>

momala454:
back to the topic, are you able to reproduce the problem on sourceforge ?

spirillen:
Hi @momala454 

I can confirm the counter ain't working, when consensu is blocked

<details><summary>ScreenShot</summary>

![image](https://user-images.githubusercontent.com/44526987/71926778-5d645a00-3194-11ea-9f8b-3a02b11c5736.png)

```
dig consentmanager.mgr.consensu.org

;; flags: qr aa rd ra; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;consentmanager.mgr.consensu.org. IN    A

;; Query time: 0 msec
;; SERVER: 127.0.0.1#53(127.0.0.1)
;; WHEN: Tue Jan 07 21:26:52 CET 2020
;; MSG SIZE  rcvd: 60
```

</details>

[...]

PR #7459, opened 2021-03-19
M: sourceforge.net (removing a cookie whitelist rule)

peace2000:
Due to this https://github.com/easylist/easylist/issues/4629 issue report, this whitelist rule was added. I tested and now downloading counter starts and I can download normally even if I block that address.

Sample link to test with: https://sourceforge.net/projects/megui/


krystian3w:
Possible hide again by scriptlet:

```
!##+js(no-setTimeout-if, window.cmpmngr.loadSettings, 30)
sourceforge.net##+js(set-constant, window.cmpmngr.setConsentViaBtn, noopFunc)
```

PR #7524, opened 2021-03-28
M: removed 404 domain

peace2000:
https://makujakauppa.fi/ is 404.


Khrin:
`https://www.makujakauppa.fi/` is still active and the filter is valid there.

PR #7616, opened 2021-04-09
A: https://leechpremium.link/index.php (fixes https://github.com/Adgu…

Yuki2718:
…ardTeam/AdguardFilters/issues/79850)
https://github.com/AdguardTeam/AdguardFilters/issues/79850


Yuki2718:
@ryanbr My bad, it seems to be self-promo. Can you move the rule to Annoyance?

PR #7649, opened 2021-04-12
A: asmand.wordpress.com (generic cookie block)

peace2000:
https://asmand.wordpress.com/

A generic cookie blocking rule for wordpress-sites.


ryanbr:
Got another example? asmand isn't showing this. @peace2000

peace2000:
Unfortunately I don't have another example. 

But I still see a banner on that page, here's a fresh screenshot:

kuva

This is the address in the logger: `https://s0.wp.com/wp-content/blog-plugins/wordads-classes/js/banner.bundle.js?id=69626cb5d25b886923fc`

PR #7701, opened 2021-04-19
A: Fix Tumblr cosmetic filtering

### LINKED ISSUE #7700: Tumblr: all cosmetic filtering broken

`tumblr.com/dashboard`

### What happens?

All cosmetic filtering is nonfunctional due to an update to the site's React CSS map.

Easylist 202104191755

marcustyphoon:
An update to Tumblr's CSS map means the current cosmetic filtering rules do not work at all. This fixes them.

`_1aQrn` used to be `instreamAd`, and it has been replaced with updated classes for `instreamAd`, `nativeIponWebAd`, and `takeoverBanner`. This also provides the functionality that the `._1DxdS:-abp-has( > div > header[role="banner"])` was supposed to provide, though it has been broken for a while.

This also hides the sidebar ad via `sidebarContent mrecContainer`, or `._3V0cR ._1s9MS`.

Using CSS map classes is certainly an imperfect solution, as updates like this will sometimes be necessary. It is possible to provide this functionality via human-readable but complex CSS rules, which can be discussed, but this PR fixes the functionality using the current method.

Resolves #7700.

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -839,7 +839,9 @@ wzzk.com##.Weather_Sponsor_Container
 this.org##.Wrap-leaderboard
 xbox.com##.XbcSponsorshipText
 rxlist.com##.Yahoo
-tumblr.com##._1aQrn
+tumblr.com##._3x1ep
+tumblr.com##.vd2CF
+tumblr.com##.kHo0b
 thequint.com##._1v3nU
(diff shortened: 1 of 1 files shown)


marcustyphoon:
Note: Tumblr stopped using `.kHo0b` (takeoverBanner) for the "takeover" style ads at the top of the dashboard shortly after I created this, but there is no unique CSS class on them right now.

PR #7723, opened 2021-04-22
A: https://mybobs.com

jdelman:
Fixes issue where customer service chatbot does not appear.

Screenshot:
Screen Shot 2021-04-22 at 5 12 05 PM

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -2172,7 +2172,6 @@
 ||smrk.io^$third-party
 ||smtrk.net^$third-party
 ||snapdeal.biz^$third-party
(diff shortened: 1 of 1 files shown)


Yuki2718:
Why remove the whole rule than adding exception, you mean the same issue occurs everywhere? If the service track user EP will block it anyway.

jdelman:
Yes, the issue occurs on multiple sites that have this same customer service widget

Yuki2718:
I can't access `mybobs.com` but on `nioxin.com` these two seem to be enough:
```
@@||pixel.snapsmedia.io/sp.js$domain=nioxin.com
@@||chat.snapsmedia.io/index.html$domain=nioxin.com
```

PR #7748, opened 2021-04-27
Hide ad on mainpage - https://allegro.pl

jarjar-creator:
allegro main page ads


Khrin:
Wrong place here, ask there instead: https://github.com/easylistpolish/easylistpolish

PR #7786, opened 2021-05-04
A: dcraddock.uk

Yuki2718:
Overlay causing popup:

dcraddock

IDK if the DOWNLOAD and PLAY NOW buttons should be addressed. Maybe in Annoyance?


Khrin:
@Yuki2718 Usually I consider such fake buttons as adverts, in this case I believe these should go to Liste FR. (@smed79 )

smed79:
> in this case I believe these should go to Liste FR

https://github.com/easylist/listefr/commit/76b9086a557ce35d8da601cfb850c71f018763ba

PR #7807, opened 2021-05-05
A: www.msn.com

Yuki2718:
msn


Khrin:
The generichide filter for MSN has been removed, this can be added as generic rule.

Yuki2718:
Okay.

PR #7905, opened 2021-05-20
M: Remove unnecessary filters

hirorpt:
https://github.com/easylist/easylist/commit/9bd4f0d5db20aa914f2201bf0c16ad661f4bc3e9

`https://web.archive.org/web/20160518165428/http://www.ntv.co.jp/`

I used the Internet Archive to check this commit and found that all the added general_block filters blocked non-advertising stuff.
In particular, `/ad-navi /*` still blocks non-ads and some of them are whitelisted.

### DIFF
--- a/easylist/easylist_allowlist.txt
+++ b/easylist/easylist_allowlist.txt
@@ -444,10 +444,8 @@
 @@||minigame.aeriagames.jp/css/videoad.css
 @@||mjhobbymassan.se/r/annonser/$image,~third-party
 @@||netmile.co.jp/ad/images/$image
-@@||news24.jp/ad-navi/$domain=search.news24.jp
 @@||nintendo.co.jp/ring/*/adv$~third-party
 @@||nnmclub.to/forum/misc/html/advert.html$~third-party
-@@||ntv.co.jp/ad-navi/$domain=search.news24.jp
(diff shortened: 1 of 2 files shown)


ryanbr:
remove generic blocks, why.

hirorpt:
> remove generic blocks, why.

Please show me such generic filters block real ads or tracking. 

As for`/ad-navi/*`, I can easily give you concrete examples of false positives which is not covered in EasyList.
1. https://github.com/AdguardTeam/AdguardFilters/issues/28185
2. `https://kinro.ntv.co.jp/`
3. `https://www.minnanokaigo.com/ad-navi/`

Also, you should not forget that adding `/ad-navi/*` in 2016 was probably @Khrin's mistake. It blocked a harmless script, `https://web.archive.org/web/20160519173455js_/http://www.ntv.co.jp/ad-navi/js_ntvtop/common.js`


About `-ad-text_` and `/column-ad-`, Please look at 
`https://web.archive.org/web/20150911203716im_/http://www.ntv.co.jp/top2010/images/column-ad-text_icon_no.gif`.
It is just a UI component. @Khrin accidentally blocked it in 2016. 
According to AdGuard statistics, these two generic filters is unused. In other words, It is unlikely to block ads, and potentially cause false positives.

PR #7907, opened 2021-05-20
M: Blocking Rakuten Automate scripts causes breakages of images or embed social media posts.

hirorpt:
`https://automate-frontend.linksynergy.com/minified_logic.js`
According to my observation, blocking this script occasionally prevents with image loading and always breaks embedded tweets and instagrams.
It is often used on Japanese sites, but is also used on English sites.
It can also be fixed with a scriptlet such as `4meee.com##+js(aopw, _rakuten_automate)`, which is the best solution, but cannot makes generic.

**Example**
Broken images
 `https://4meee.com/articles/view/30008569`
 `https://jamalouki.net/`
 `https://www.winecompanion.com.au/`
 
Broken embed tweets and/or instagram
 `https://www.gamespark.jp/article/2017/12/14/77430.html`
 `https://petpedia.net/article/894/neko_hetakuso`
 `https://www.sassyhongkong.com/lifestyle-culture-instagram-apps-insta-story-tips/`


hirorpt:
The label `Fanboy Social` is incorrect. This is a issue with EasyPrivacy.

PR #7939, opened 2021-05-27
A: (nsfw) https://www.pornjam.com/video/perky-tits-blonde-toying-cunt/

Yuki2718:
Not sure if banner under the player is self-promo or ads so didn't address.


spirillen:
Not bad, however these three have come to be interesting as well :smirk: 

```
pornjam.com##.ntv-disclaimer
pornjam.com##.r1ght-pl4yer-169
pornjam.com##.publ1s-b0ttom
```

PR #7980, opened 2021-05-31
A: Update adservers

smed79:


easylist/easylist_adservers.txt
```diff
@@ -4198,6 +4220,8 @@
 ||gagacon.com^
 ||gagpetwet.com^
 ||gagwithholdoutsider.com^
+||gaietybeautify.com^
+||gaietybeautify.com^
```
spirillen, line 4224: You have listed this one twice :smile:

Khrin: In any case, it's listed only once: https://github.com/easylist/easylist/blob/master/easylist/easylist_adservers.txt#L4227

PR #8123, opened 2021-06-18
M: generic cookie filter modified to avoid unwanted blockings

peace2000:
Rule `###ot-sdk-btn` might in some cases remove legit buttons, like it does on https://www.maaseuduntulevaisuus.fi/ it removes a button where you can access website's cookie settings. That should not in my opinion be blocked.

Modifying the rule to be like this `###ot-sdk-btn:not(button)` fixes the issue.

Sample picture: 
kuva

(Scroll to the bottom of the page. _Evästeasetukset_ means "Cookie settings" in English.


peace2000:
Sorry to directly mention @Khrin but this potentially affects numerous sites, not just this one.

Issue #8165, opened 2021-06-21
699pic.com


shenzhiming88:
### List the website(s) you're having issues:
699pic.com

### What happens?
Pop-up window appears.

### List Subscriptions you're using:
`EasyList`
`EasyPrivacy`
`Fanboy’s Annoyance`

### Your settings

- OS/version: --
- Browser/version: --
- Adblock Extension/version: --

### Other details:
![img](https://camo.githubusercontent.com/292ffb30265cb9002e140047fd18e931a9399d1b371ae05869daa6fbd8e9ebd1/68747470733a2f2f7265706f7274732d696d672e616467756172642e636f6d2f464976453048332e6a7067)

PR #8166, opened 2021-06-21
A: https://699pic.com/

shenzhiming88:
https://github.com/easylist/easylist/issues/8165


shenzhiming88:
@ryanbr Hello!
I think you'd better move these rules to fanboy_annoyance_international.txt, because this is a Chinese specific website.

PR #8217, opened 2021-06-29
##.sponsor-grid__sponsor + ##.sponsors

spirillen:
Both found on pypi.org

```
pypi.org##.sponsor-grid__sponsor
pypi.org##.sponsors
```

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -9234,7 +9234,6 @@
 ##.SponsoredLinksPanel
 ##.SponsoredResults
 ##.Sponsored_link
-##.sponsors
 ##.SponsorshipText
 ##.SquareAd
 ##.Squareadspot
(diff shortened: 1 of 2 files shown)


easylist/easylist_general_hide.txt
```diff
@@ -9234,6 +9234,7 @@
 ##.SponsoredLinksPanel
 ##.SponsoredResults
 ##.Sponsored_link
+##.sponsors
```
smed79, line 9237: Might be too generic, you could make this specific by adding it here :

https://github.com/easylist/easylist/blob/038e8259f4bbc0f4ab94ffb30b8125c67de7ce2b/easylist/easylist_specific_hide.txt#L1996

spirillen: I was thinking about whether it would be to generic...

I'll move it a bit later :smile: 

Suger fingers...

PR #8311, opened 2021-07-07
M: filters no more needed

Yuki2718:
Now video works fine without them.


stm-x:
Perhaps this PR should be changed.
https://github.com/AdguardTeam/AdguardFilters/issues/89007

PR #8333, opened 2021-07-11
A: sss.fi (cookie block, Ubo)

peace2000:
Cosmetic filter works too slowly, I see blinks of it occasionally. Added a scriptlet to block it.

https://sss.fi/


chocolateboy:
Did this get mangled at some point? Looks like the rule will always be discarded as a syntax error:

https://github.com/easylist/easylist/blob/1103ad6f8935add72d3e362f59e9831bd293eb8e/easylist_cookie/easylist_cookie_specific_uBO.txt#L149

peace2000:
@chocolateboy No issues on my end.

Sometimes uBO shows scriplet rules as invalid when viewing a scriplet rule on asset viewer. When I have refreshed the view, no syntax errors anymore.

chocolateboy:
Ah, thanks, my mistake. I thought the third argument was an expression. Didn't realize it's a pattern.

PR #8390, opened 2021-07-21
P: danskebank.fi (some pictures won't load, cookielist, Ubo)

peace2000:
https://danskebank.fi/ some pictures won't load when current rules are being used. Not serious breakage, but this will work better.

Also added * to cover it's other top level domains (such as `.dk`, `.se` etc.).


peace2000:
I'll remake...

PR #8393, opened 2021-07-21
M: redundant filter

peace2000:
https://www.alko.fi/ doesn't use cookie banner anymore.

Edit, it uses but is blocked via `||cookielaw.org^$3p`

So whitelist rule is useless anyway. I didn't notice issues on that site.


peace2000:
Sorry I was wrong, some race-condition I guess....

PR #8404, opened 2021-07-22
P: / A: vr.fi (cookie whitelist for ABP and addition for Ubo)

peace2000:
https://www.vr.fi/en

Easylist Cookie removes part of the cookie banner so it's not possible to accept cookies.

Added filters that will block the banner properly for Ubo (not possible for ABP currently).


ryanbr:
Rebase @peace2000

peace2000:
I'll make a new.

PR #8407, opened 2021-07-22
A: https://www.techpowerup.com/

smed79:


ryanbr:
Haven't checked, but it'll be out of date.

PR #8414, opened 2021-07-23
A: varusteleka.com|.fi (cookie hide, Ubo)

peace2000:
Added a specific filter to avoid blinking issues.

https://www.varusteleka.com/


ryanbr:
Just ensure its only uBO specific filters, and not used in `easylist_cookie_specific_hide.txt` @peace2000

peace2000:
@ryanbr it can be used in `easylist_cookie_specific_hide.txt`, the reason I added it here was to overcome blinking issues on Ubo. If that filter would be added to `easylist_cookie_specific_hide.txt`, someone might one day remove it, thinking that it's not neede as there's a generic one also...

PR #8428, opened 2021-07-24
A: bosch-home.fi | .* (cookie list, Ubo)

peace2000:
https://www.bosch-home.fi and all it's other top level domains as there are many of them.

Rule `bosch-home.*##.cookielaw-modal-layer` is needed at least in the Finnish site.

`bosch-home.*##[class="o-cookielaw"]` is needed at least here: https://www.bosch-home.com/?clear=1#/Togglebox=region-toggle-4/

`bosch-home.*##+js(rc, pinned|cookielaw-blur-background, body, stay)` will remove blur effect and overflow hidden properties from the body element.

Style rules
```
bosch-home.*##body:style(overflow: auto !important)
bosch-home.*##.g-container, .feedback-button:style(filter: none !important)
```

Are added in order to avoid blinking issues on visible blurred elements.
That site cluster uses blur effect on many elements and menus (Cookie notification related):

kuva

kuva

So it's not wise to individually apply style rules to every element on that site. Just applied it to those elements that are immediately visible to avoid blinking. Scriplet will take care of the rest without blinking.

Hopefully you got the idea. :D... This might look a bit complex but imo every filter is necessary.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -1,7 +1,6 @@
 ! uBO Sepecific fixes
 ! :style fixes
-bosch-home.*##.g-container:style(filter: none !important)
-bosch-home.*##.feedback-button:style(filter: none !important)
+bosch-home.*##.g-container, .feedback-button:style(filter: none !important)
 wetter.at##.full.blured:style(filter: none !important;)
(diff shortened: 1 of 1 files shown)


ryanbr:
rebase @peace2000

peace2000:
Gotta make new pull req...

Issue #8429, opened 2021-07-24
Elektroskandia.no - Overlay modal (Easylist Cookie List / Fanboy's Annoyance List) [GDPR]


THEtomaso:
### Affected site:
`https://webshop.elektroskandia.no/`

### Issue:
Grey overlay modal.

### Screenshot:
Not needed, as the problem is obvious.

### Problem filters:
* **Easylist Cookie List**
* **Fanboy's Annoyance List**

### Problem rules:
Various cookie blocking rules.

### Possible fix:
`elektroskandia.no##.modal-backdrop`

--

### System info:
* OS/version: Windows 8.1 Pro (x64)
* Browser/version: Pale Moon v29.3.0 (x64)
* Adblock Extension/version: uBlock Origin v1.16.4.30 (XUL)

THEtomaso:
@DandelionSprout

peace2000:
https://github.com/easylist/easylist/pull/8431 This pull request should be accepted in order to avoid such issues again with this filter.

peace2000:
This is a very common issue that Easylist breaks sites because filters match the body element.

https://github.com/easylist/easylist/issues/8367

THEtomaso:
> https://github.com/easylist/easylist/pull/8431

That's good.
`elektroskandia.no##.modal-backdrop` is still needed though, in order to remove the remaining overlay modal!

_First post updated_

PR #8431, opened 2021-07-24
P: webshop.elektroskandia.no (cookie list breakage, general issue)

peace2000:
https://webshop.elektroskandia.no/nor/


THEtomaso:
In the case of `elektroskandia.no`, an additional rule is needed, in order to remove the remaining overlay modal!:
https://github.com/easylist/easylist/issues/8429#issuecomment-886952900

PR #8449, opened 2021-07-26
A: audiovideo.fi (generic ad hide)

peace2000:
Sample site: https://audiovideo.fi/

Used widely: https://publicwww.com/websites/%22bunyad-widget-ads%22/


ryanbr:
`##[id|="bunyad-widget-ads"]` is invalid @peace2000

peace2000:
How so? it works fine in ABP and in Ubo. `|` is better than `^` because it's more specific and in this case it's enough.

In every known case that Publicwww lists https://publicwww.com/websites/%22bunyad-widget-ads%22/ `bunyad-widget-ads` is followed by a dash and a number.

PR #8455, opened 2021-07-27
A: hs.fi (social generic hide)

peace2000:
Sample page: https://www.hs.fi/fingerpori/car-2000008120137.html

Word "some" is a very common Finnish colloquial for "Social media", thus it should be included.


ryanbr:
Not a good generic. @peace2000

peace2000:
I wanted to edit this and also do some alphabetical ordering but decided to cancel that so I'll make a new pull req.

Issue #7872, opened 2021-05-15
arc.io untracking ads


c2xusnpq6:
https://arc.io/ ---> `{your_website}/arc-sw.js`
```
/arc-sw.js
```
![image](https://user-images.githubusercontent.com/75058881/118351118-9680bb80-b58c-11eb-9780-da43be4c76af.png)

it's untracking ads, or.... just let it go? ha~

it just like coin miner, but lite version?

c2xusnpq6:
![image](https://user-images.githubusercontent.com/75058881/118351416-8538ae80-b58e-11eb-8115-a3536597fada.png)

c2xusnpq6:
one more... https://joinmassive.com/ lite coin miner

ryanbr:
Sample urls please @c2xusnpq6

c2xusnpq6:
> Sample urls please @c2xusnpq6 

I don’t have a website, I can’t use that function, or you can apply for an account to find out

spirillen:
Hey @ryanbr 

For what I can read from the site "offering" this snippet, I would probably actually classify it as malware as it is "hijacking" the users bandwidth and browser cache without there knowledge and this can clearly become a costly affair on metered connection, like the one I'm on right now.

This said, I wouldn't be affraid trying to add @c2xusnpq6 suggestion for a filter on this on a global level as suggested. 

Better on the safe side than sorry, if it turns out that there somehow should appear a naming match later, we should take it up at that time.

```
/arc-sw.js
```

> Arc uses only a small portion of spare bandwidth, imperceptible CPU, 300MB of browser cache, and only runs when a device is connected to Wi-Fi or Ethernet. Cellular bandwidth is never used. 

> Arc creates a communal, distributed CDN with the latest HTML5 APIs — WebRTC to connect browsers, Service Worker to fulfill requests, IndexedDB to cache data, and WebCrypto to do it all securely.

c2xusnpq6:
similar mining platform:

> one more... https://joinmassive.com/ lite coin miner

gruns:
Hi guys. :wave:

I'm Ansgar. I work on Arc.

I want to clarify a few things.

> I would probably actually classify it as malware as it is "hijacking"
> the users bandwidth and browser cache without there knowledge and this
> can clearly become a costly affair on metered connection, like the one
> I'm on right now.

- Arc never uses cellular bandwidth. If a device is on a cellular
  connection, Arc does not activate and no bandwidth is used.

[...]

PR #8458, opened 2021-07-28
Fix Monzo Sentry

bentasker:
The Monzo (https://monzo.com/) payment verification process is broken by the widescale blocking of sentry (introduced in d610e11), because it uses sentry to send events to the payment page once the user has verified payment in their app.

With the block in place, the user will approve the payment in the Monzo app, but the merchants page will never update (you can force it by right clicking on the frame and choosing reload).

More discussion can be found on Twitter here - https://twitter.com/bentasker/status/1420354891453763588

### DIFF
--- a/easyprivacy/easyprivacy_allowlist.txt
+++ b/easyprivacy/easyprivacy_allowlist.txt
@@ -253,7 +253,6 @@
 @@||playcanvas.com.*/keen.min.js
 @@||player.ooyala.com^*/analytics-plugin/$script,domain=formula1.com|nintendo.com
 @@||player.sundaysky.com^$subdocument
-@@||playstation.net/event/$domain=playstation.com|playstation.net|sony.com
 @@||plugins.matomo.org^$image,~third-party
 @@||postex.com/api/ping?$~third-party
 @@||pro.ip-api.com/json$xmlhttprequest,domain=aljazeera.com|cookappsgames.com|sheee.co.il|walla.co.il
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -1012,6 +1012,7 @@
 /bsuite/worker.php?
 /bt.gif?
 /btn_tracking_pixel.
+/btrack.php?
 /buffer.pgif?r=
 /buffermetrics/*
 /bug/pic.gif?


ryanbr:
Was blocked long before that commit, if we can't reproduce it. It won't be fixed in EP. Privacy. Maybe have a fallback option here.

bentasker:
Honestly, that's a pretty crap response.

It doesn't matter when you blocked it, it's broken a service I use.

> Maybe have a fallback option here.

I'm a Monzo user, I have nothing to do with the org or their codebase.

> if we can't reproduce it. It won't be fixed in EP. Privacy

To repro, you'd have needed to be able to open a UK bank account (i.e. get a Monzo account), put money in, then attempt to make an online payment. So, this approach means (more or less) any banking service that gets broken by a EP Privacy overblock gets to stay broken.

In this case, it's moot, because Monzo have been [quite responsive](https://twitter.com/rubendura/status/1420655054344830980) and dealt with the issue, but this approach doesn't exactly serve to build confidence in using these lists.

At the end of the day, I'm a user of your lists, and of Monzo, if the detail I provided wasn't enough, maybe ask for more? Especially given it stems from a fairly wide block of something that's been reported as [an overblock](https://forums.lanik.us/viewtopic.php?f=64&t=31754&p=99380&hilit=sentry+cdn.com#p99380) on [multiple occasions](https://forums.lanik.us/viewtopic.php?f=64&t=43332&p=149006&hilit=sentry+cdn.com#p149006).

spirillen:
EasyList(+*)  Can never be responsible for whatever spyware/trackWare/malWare etc anydomain owner might add to there code base (website). 

In case there arise a conflict between Easylists attempt to protect users and your human right to privacy against evil greedy companies, then it is the Evil Webmaster who have chosen to "break" there own website by adding evil code to it.

If you then, despite this warning from @EasyList, still would like to have your internal organs harvested by mr. greedy :dollar: then you always have your personal right to do this by whitelisting such thing. In addition to this you could add the needed whitelisting rule here for sharing for other like minded.

The discussion of blocking `sentry-cdn.com` have ended.

The threads about that is to be located these places:
  - https://github.com/easylist/easylist/issues/4060 (sentry.io)
  - https://github.com/easylist/easylist/issues/6963 (sentry.io)
  - https://github.com/easylist/easylist/issues/7872
  - https://mypdns.org/my-privacy-dns/matrix/-/issues/1390 (sentry.io)
  - https://mypdns.org/my-privacy-dns/issues/-/issues/4349 (sentry-cdn.com)

bentasker:
Again, that's a pretty crap response.

The block _literally_ broke payment flows. Not everyone has the skillset to dig into what the issue was and what to whitelist - the pull would have resolved things for the cross-section of your list users who are also use Monzo as a bank.

I'm not overly interested in the wider discussion about whether Sentry should be blocked, I just want the stuff I use to work.

Thanks for the links - there's probably some selection bias, but certainly in #6963 the weight of opinion seems to be against, not with you. I'm not really sure of the relevance of #7872 though, other than that they briefly mention they use Sentry

Anyway, it's moot now because Monzo have fixed it on their end, but I can't lie - I find the response to this fairly disappointing. As user engagement goes, it's pretty poor.

But, ultimately it's your list to manage how you see fit.

Given it's something that keeps coming up (and will likely continue to do so), maybe update the README to state explicit policy on Sentry so that others don't waste their time creating a pull?

> In addition to this you could add the needed whitelisting rule here for sharing for other like minded.

We're literally commenting on a pull request, the filter's in the commit :)

spirillen:
> Again, that's a pretty crap response.

Not sure what you would have expected. I did my best as a contributor like your self on behalf of the issue linked

> I'm not overly interested in the wider discussion about whether Sentry should be blocked, I just want the stuff I use to work.

Me neither, just given answers to your question. :wink: 

> Thanks for the links - there's probably some selection bias, but certainly in #6963 the weight of opinion seems to be against, not with you. I'm not really sure of the relevance of #7872 though, other than that they briefly mention they use Sentry

Well I did actually just used the search within the repo :smiley: + plus added my own issue for the topic's

> Anyway, it's moot now because Monzo have fixed it on their end, but I can't lie - I find the response to this fairly disappointing. As user engagement goes, it's pretty poor.

Hmm, your opinion as I've linked to some rather long discussions on the topic you can read up on rather than repeating them.

> Given it's something that keeps coming up (and will likely continue to do so), maybe update the README to state explicit policy on Sentry so that others don't waste their time creating a pull?

Or use the search before opening a new issue to see if one already exist?

> > In addition to this you could add the needed whitelisting rule here for sharing for other like minded.
> 
> We're literally commenting on a pull request, the filter's in the commit :)

:smile: :rofl: sometimes that just disappear from my mind while doing searches for answers :shrug:

simonerni:
Hey @spirillen, sorry to open this again.

Some domains (orionprotocol.io, tesco.com) are allowed to use Sentry: 
https://github.com/easylist/easylist/blob/d2bad7914f85024c4c6b1bbfd2dd6890fc275aff/easyprivacy/easyprivacy_allowlist.txt#L227

Can you add our domains to that list as well?

Thanks!

spirillen:
Hey @simonerni You should ping @ryanbr  as I (as mentioned) only contribute my finds and opinions without power to this project.

But yes, it might be possible to find exceptions, isn't there exceptions to any rules?

Because I have something added to my own project, doesn't mean I use the tor-browser to overcome exceptions if required. Thanks for the comment anyway :smiley:

ryanbr:
We're not adding domains to this list without a very good reason, allowing analytics or tracking, event logs isn't one.

simonerni:
Hi @ryanbr, @spirillen,

Thanks for the replies! I completely understand.

In our use case, we‘d like to monitor the violations of the Content Security Policy (CSP) to spot any XSS / HTML Injection vulnerabilities in our applications as they happen. In effect, we can stop such attacks against our users pro- and retroactively, keeping their data safe.

It‘s an important topic to us, since these attacks are easy to execute, have a large impact, and can be easily prevented and ultimately detected - but we need these reports for this.

Is that something that could be allowed?

Thanks & Best Regards,
Simon

simonerni:
Sorry I forgot to mention the application itself - we‘re a pension fund in Switzerland with several offerings. We like to give our customers the most control about their finances and have all our servers in Switzerland.

Our domains are:

finpension.ch
yourpension.ch
valuepension.ch
myhostpoint.ch

Thanks again for looking into this!

Best regards,
Simon

ryanbr:
We're good, leaving.

PR #8472, opened 2021-07-29
P: danskebank.fi (some pictures won't load, cookielist, Ubo) (remade)

peace2000:
https://github.com/easylist/easylist/pull/8390/


peace2000:
I'll resolve conflicts after this https://github.com/easylist/easylist/pull/8474 is accepted as they try to edit the same line of code so no point to fix this yet.

PR #8388, opened 2021-07-21
A: calvinclein.* (Cookie, Ubo)

peace2000:
https://www.calvinklein.fi/

... and it's other top level domains.

PR #8474, opened 2021-07-29
 A: calvinclein.* (Cookie, Ubo) (remade)

peace2000:
Original https://github.com/easylist/easylist/pull/8388


peace2000:
This can't be squashed, as Github is buggy. All I did was that I edited few lines in uBO's cookie list file now GH thinks that I have made a commit to edit 70 files and 10 000+ characters of text. Some old pulls can't be squashed either because of this.

https://github.com/easylist/easylist/commit/132f3ed5ef621e962ae5109ca649ef95c5219f60

Issue #8367, opened 2021-07-18
Easylist Cookie (and Annoyance lists generally) body and button issues on many websites


peace2000:
It is a somewhat common thing that there is a need to add `:not` exceptions to generic cosmetic filters in the Easylist Cookie list in order to unbreak websites. A sample rule from the list:

`###gdpr:not(body):not(html):not(input):not(button)`

I see here and there cosmetic rules that have `:not(body)` or `html` or `input `or `button` or `.checkbox` exceptions added to them... Easylist Cookie either
- Blanks out websites
- Removes accept buttons / checkboxes

When it matches to those elements.

I have couple of active pull requests and some non-actives as well, concerning these kind of issues.

I suggest using `:not` exceptions as a common practice for generic hiding rules in Easylist cookie. This would preemptively unbreak countless websites and reduce repetitive maintenance work in applying exceptions and whitelist entries. Imo it's pointless to add `:not` one after another for Easylist Cookie's generic rules. Why not to apply them to all entries? I wonder how many unreported broken websites there are...

So what I suggest, is that after each generic cosmetic entry, this exception would be added:

`:not(body):not(html):not(input):not(button):not(.checkbox)`

It might not look elegant mut it would unbreak a lot and reduce maintenance work. Body and html elements should never be blocked, for obvious reasons. Input, button and checkbox elements should not be blocked either.

DandelionSprout:
Universally whitelisting `#@#button`, would cause problems for lists that block social media sharing buttons.

Whitelisting `#@#html` is actually a rather good idea, to the point I'll add it to https://github.com/DandelionSprout/adfilt/blob/master/FalukorvList.txt. I also see no real problems with `#@#body`.

I lack deep enough experience with `input` element to make any statements on that one.

peace2000:
> 
> 
> Universally whitelisting `#@#button`, would cause problems for lists that block social media sharing buttons.
> 
> Whitelisting `#@#html` is actually a rather good idea, to the point I'll add it to https://github.com/DandelionSprout/adfilt/blob/master/FalukorvList.txt. I also see no real problems with `#@#body`.
> 
> I lack deep enough experience with `input` element to make any statements on that one.

`#@#html` or `#@#body` would just whitelist rules like ##html` or `##body`. They cannot be, unfortunately, used to whitelist html and body elements generally, if they are matched by for e.g. `##.cookie-banner` when `##.cookie-banner` is a value in a body or html element. (I whish they could be easily whitelisted in that way).

[...]

PR #8508, opened 2021-07-31
P: iltalehti.fi (cookie accept button removed, and the button can't be clicked)

peace2000:
https://www.iltalehti.fi/kotimaa/a/7bda0d36-ad9d-4221-9f71-371881a1eb4f

kuva

```
Sisältöä ei voida näyttää evästeasetusten vuoksi. Nähdäksesi sisällön,
hyväksy evästeiden käyttö.

```
Translation:

```
Content can't be shown due to cookie settings. In order to see the content,
please accept cookies.
```

But the accept button is removed due this filter, so added exception...

EDIT:

I also found out that a whitelist rule (network filter) is needed so that the button can be clicked.


peace2000:
I'll remind again this https://github.com/easylist/easylist/issues/8367 issue report I opened. These issues with generic cookie filters are pretty common and it would be good to to apply generic fix for them as a preemptive measure.

PR #8519, opened 2021-08-02
A: https://www.deepl.com/translator

shenzhiming88:
Fixes https://github.com/easylist/easylist/pull/8412
Self-promotion
image
image


Khrin:

PR #8349, opened 2021-07-13
P: podleze-piekielko.pl (scrolling blocked by Cookie List)

peace2000:
https://podleze-piekielko.pl/

There needs to be whitelisting for other blockers: https://github.com/easylist/easylist/issues/8291

Edit, also added a rule for Ubo to counter this whitelisting rule.

Edit2: removed all dom and style rules and added scriptlets for both ABP and Ubo.

PR #8532, opened 2021-08-03
A: podleze-piekielko.pl (Ubo, cookie)

peace2000:
Remade of of https://github.com/easylist/easylist/pull/8349/

Style injection not needed as scriptlet deals the cookie banner.


krystian3w:
warta.pl has it

PR #8533, opened 2021-08-03
 A: podleze-piekielko.pl (ABP, cookie)

peace2000:
Remade of https://github.com/easylist/easylist/pull/8349

I don't know if ABP accepts this for Cookie list but will see.


ryanbr:
Acccepted but ABP doesn't support snippets outside of its own lists @peace2000

peace2000:
@ryanbr How then it worked for me when I tested it in ABP? I don't understand.

krystian3w:
warta.pl has it

PR #8562, opened 2021-08-05
P: https://hinagiku-u.ed.jp/

Yuki2718:
Another case of broken **ad**mission:

hinagiku1

hinagiku2


Yuki2718:
@ryanbr @Khrin `hinagiku-u.ed.jp#@##ad-box` is also needed.

PR #8564, opened 2021-08-06
Fixes https://github.com/uBlockOrigin/uAssets/issues/9697#issuecomment-892831828

DeepChirp:
The page will automatically pop up the login window.


ryanbr:
We won't break logins in Annoyances,

uBlock-user:
@ryanbr Login is hosted at `https://passport.csdn.net/account/login?spm=1001.2101.3001.5105`, nothing is broken by blocking that overlay.

PR #8585, opened 2021-08-08
M: changed generic filter to make it more safe

peace2000:
I think that instead of whitelisting this site, it'd better instead to edit the original rule to make sure it never breaks any sites by matching body elements.


peace2000:
I'm also promoting this suggestion https://github.com/easylist/easylist/issues/8367

PR #8586, opened 2021-08-08
A: (nsfw) titsbox.com

Yuki2718:


spirillen (approved):

spirillen:
I would add this rule as well as it is an ad/roulette not webcams...

`titsbox.com##[class="hidden-xs"]`

See examples in https://mypdns.org/my-privacy-dns/matrix/-/issues/2356

PR #8593, opened 2021-08-08
P: m.motonet.fi (product images blocked)

peace2000:
https://m.motonet.fi/


spirillen:
Why not make it global?? It sounds like it should be possible :smirk:

peace2000:
Sorry I didn't understand.

spirillen:
Oh sorry, I haven't notice the `@` for whitelisting...

PR #8602, opened 2021-08-09
M: https://healthy.kaiserpermanente.org/front-door - more granular blocking for Dynatrace

ghaydtner:
Extend the rule for Dynatrace. This still matches URLs like https://dynatrace-ag1.kp.org/jstag/managed/ea96fd35-4a91-4571-9999-391fba61c917/2244804940fc3be6_complete.js on websites like https://healthy.kaiserpermanente.org/front-door but avoids broken pages on websites like https://www.dynatrace.com/news/customers/.

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -1610,7 +1610,7 @@
 /DynamicAnalytics.
 /dynamictag.
 /dynamicyield/*
-/dynatrace-
+/dynatrace-ag
 /dynaTraceMonitor^
 /dynimpression?rkey=
 /e.gif?


spirillen:
????

image

ghaydtner:
@spirillen Please use HTTPS & the full URL: https://healthy.kaiserpermanente.org/front-door
Original commit adding the rule: https://github.com/easylist/easylist/commit/1b9cf1f662f704d3ee1d372bd577d6df03705200

spirillen:
> [https://healthy.kaiserpermanente.org/](https://healthy.kaiserpermanente.org/front-door)

The diff??

image

ghaydtner:
@spirillen I don't use a cache. The first screenshot didn't show the full URL, but the error message said http://... instead of https:// so that looked like the issue. For me, both URLs work. From my machine:
image
Tools like httpstatus.io show too that the URL works:
image
So I'm not sure why you get an "Access denied" error, but that may be out of proxy/VPN settings.

The issue is not with the https://healthy.kaiserpermanente.org/front-door page, the issue is with pages like https://www.dynatrace.com/news/customers/ as the original rule blocks the CSS files for that website. Adopting the rule makes sure that the blocking for Kaiser Permanente is still in place while the Dynatrace website works again.

ryanbr:
Dynatrace is a tracker. "Making more granular" seems to avoid other Dynatrace scripts here.  
Addressed `dynatrace.com` in https://github.com/easylist/easylist/commit/02974dab4cf6e5e496a0ad0d462322fb9ca5e233

But Visiting the site has no issues here.

ghaydtner:
What "dynatrace-" tried to block is https://dynatrace-ag1.kp.org/jstag/managed/. This is an instance of Dynatrace hosted by Kaiser Permanente itself. "dynatrace-" in general therefore doesn't block Dynatrace.

@ryanbr The current rule breaks the WordPress part of the Dynatrace website, www.dynatrace.com/news as "dynatrace-" blocks CSS and JS from the WordPress site:
image
See for example https://www.dynatrace.com/news/customers/. Does that work for you?

arminru:
Hi @ryanbr! This still breaks the search and repos in [github.com/dynatrace-oss](https://github.com/dynatrace-oss). When typing into the org page search bar or viewing repos Github uses XHR requests which are still blocked by that rule. Would it be possible to refine it a bit further?

image

If I got the commit message in 1b9cf1f right, you were trying to block a host that contained `dynatrace-` in the hostname, so maybe looking only in the host part (`//dynatrace-` instead of `/dynatrace-`) would be sufficient.

PR #8618, opened 2021-08-10
A: magic.wizards.com

Yuki2718:
wizards

There's also cookie banner as you see (FB Annoyance enabled in the SS), but since it requires a complex rule I skipped in this PR, plz fix on your end.


spirillen:
You should take a view on this script, from a fast review it looks like tracking

`https://cdn.gsght.com/twizards.js`

Yuki2718:
I'm aware that the tracking pixel is called from that script. If the site in question is sensitive (this is an online real-time game site involving money), I'm not willing to risk blocking initiator as it's more likely to cause trouble, unless I myself use the site daily and thus am sure it doesn't cause trouble.

PR #8631, opened 2021-08-11
A: https://www1.kissanimes.cc/watch-anime/the-idolmster-cinderella-gi…

Yuki2718:
…rls-spin-off


ryanbr:
Not seeing "recipefy.net/rssfeed.php" is this needed? and should it be in Social? @Yuki2718

PR #8671, opened 2021-08-16
A: edition.cnn.com

Yuki2718:
Not sure whether `||cnn.com/ad/` is still in use.


ryanbr:
Reverted, broke http://go.cnn.com

Yuki2718:
@ryanbr No need to totally remove the blocking rule, `@@||cdn.cnn.com/ads/adfuel/ais/*/cnngo-ais.min.js$domain=go.cnn.com` in AG Base works fine.

PR #8748, opened 2021-08-19
P: reklama.tochka.com [rule fix]

nikitindv:
Hi guys!
I've found missing $ sign in $domain specifier according to syntax [rules](https://help.eyeo.com/adblockplus/how-to-write-filters), so actually, this didn't work for [8665](https://github.com/easylist/easylist/pull/8665).


Khrin:
The dollar sign isn't needed for every filter option.

PR #8869, opened 2021-08-27
Drop too short filters in FOP.py

mjethani:
Based on a request from @ryanbr 

If a filter contains less than three characters and is _not_ a comment, it should be dropped.

Input:

```
a
ab
abc
```

Output:

```
abc
```

### DIFF
--- a/FOP.py
+++ b/FOP.py
@@ -216,6 +216,9 @@ def fopsort (filename):
                         filterlines = elementlines = 0
                     outputfile.write("{line}\n".format(line = line))
                 else:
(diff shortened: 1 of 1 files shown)


ryanbr:
Okay, lets do this for 2 chars or less then

mjethani:
> Okay, lets do this for 2 chars or less then

It's already two characters or less. You can merge it as it is.

This means `a` and `ab` will be removed, but `abc`, `abcd`, and so on will not be removed.

PR #8875, opened 2021-08-27
A: https://www.youtube.com/embed/BlwrNtbndR4?enablejsapi=1

smed79:


Khrin:
Was already listed in EasyPrivacy, but recently has been removed because for some users caused issues to YT's watch history.

PR #8880, opened 2021-08-27
Hide specific AMP elements instead of .amp-unresolved

jridgewell:
AMP adds `.amp-unresolved` to every element while its associated JS file is being downloaded. When the JS file is downloaded, the element is "upgraded" and we remove the `.amp-unresolved` class.

I believe easylist is taking advantage of this to hide the elements that have their JS blocked. However, this causes a lot of Cumulative Layout Shift issues, because elements are popping into view when the JS downloads. Eg, all of our `<amp-img>` (which aren't blocked by easylist) start as `display: none` and then once they're upgraded they become visible, causing shifts.

This is a big issue for us, as one of our core goals is for elements to never shift the pages layout. We go through a lot work to make sure every AMP element on the page has the correct size as early as possible.

Worse yet, AMP is working on a new element upgrade system that defers expensive work until it's necessary. We've essentially implemented something just like lazy loading images, but it'll work for every AMP element on the page. This depends on knowing exactly where the element is on the page, so that we can defer the upgrade until it's near viewport. Because easylist is hiding every AMP element on the page, our `IntersectionObserver` will never fire. This means we'll never upgrade the element (and we'll never remove the `.amp-unresolved` class because that happens _after_ upgrade). So every AMP element just stays hidden permanently. See https://github.com/ampproject/amphtml/pull/35845.

So instead hiding elements indirectly through `.amp-unresolved`, I've gone through and hidden the specific element that's associated with a blocked JS file.

| List                                   | Blocked JS                                                              | Hidden Elements                                                          |
|----------------------------------------|-------------------------------------------------------------------------|--------------------------------------------------------------------------|
| fanboy_notifications_general_block.txt | - `/amp-user-notification-` <br> - `/amp-web-push-`                     | - `##amp-user-notification` <br> - `##amp-web-push-widget`               |
| fanboy_annoyance_general_block.txt     | - `/amp-apester-`                                                       | - `##amp-apester-media` (**new**)                                        |
| fanboy_social_general_block.txt        | - `/amp-social-`                                                        | - `##amp-social-share`                                                   |
| easylist_cookie_general_block.txt      | - `/amp-consent-`                                                       | - `##amp-consent`                                                        |
| easylist_general_block.txt             | - `/amp-ad-` <br> - `/amp-auto-ads-` <br> - `/amp-sticky-ad-` (**new**) | - `##amp-ad` <br> - `##amp-custom-ad` <br> - `##amp-sticky-ad` (**new**) |

(This effectively reverts https://github.com/easylist/easylist/commit/f1e088a3f00)

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -14396,7 +14396,6 @@
 ##.amp-ad
 ##.amp-ads
 ##.amp-ads-container
-##.amp-unresolved
 ##.ampFlyAdd
 ##.amsSparkleAdWrapper
 ##.anc_ads_show
@@ -17393,7 +17392,6 @@
 ##.hwg-row-ad
 ##.hyad
 ##.hype_adrotate_widget
-##.i-amphtml-unresolved
 ##.i360ad
 ##.iAdserver
 ##.iRx9
@@ -22854,6 +22852,7 @@
 ##.ampad
 ##.spotim-amp-list-ad
 ##AMP-AD
+##amp-sticky-ad
 ! Dealnews (Hearst Newspapers)
 ##div[class$="dealnews"] > .dealnews
 ! Genric mobile element
--- a/fanboy-addon/fanboy_annoyance_general_hide.txt
+++ b/fanboy-addon/fanboy_annoyance_general_hide.txt
@@ -3698,3 +3698,5 @@
 ###searchform > #gb > #gbw .gb_Mc[aria-label="promo"] > .gb_ea
 ! Google default homepage nag
 ###searchform > #gb > #gbw .gb_Vc[aria-label="promo"] > .gb_ga
+! Amp
+##amp-apester-media
--- a/fanboy-addon/fanboy_notifications_general_hide.txt
+++ b/fanboy-addon/fanboy_notifications_general_hide.txt
@@ -150,6 +150,7 @@
 ##.window-push
 ##.wordpress-fire-push-popup
 ##.ys-push
+##amp-user-notification
 ##amp-web-push-widget
 ##div[class^="container_NotificationsBanner_"]
 ! Mobile Notifications


easylist/easylist_general_block.txt
```diff
@@ -3375,6 +3375,7 @@
 /ame-dfp/*
 /amp-ad-
 /amp-auto-ads-
+/amp-sticky-ad-
```
Yuki2718, line 3378: This is not needed by another rule `-sticky-ad-`.

jridgewell:
```suggestion
```

jridgewell:
Is there anything else I can do to get this merged?

ryanbr:
amp-unresolved is still being hit. `https://www.axios.com/cair-lawsuit-muslim-board-b33a27b6-cb55-483c-809c-1a0f909ae172.html`

amp-unresolved

jridgewell:
Inspecting the page, these appear to be all `<amp-ad>` elements. I think they're being caught as `.amp-unresolved` because the rule appears earlier in the list (the `##.amp-ad` won't fire, the `##amp-ad` one will but it's on line 22807)

ryanbr:
I'm not comfortable making any wholesale changes as of right now. .amp-unresolved shouldn't be hit at all (if were unused or dead).

jridgewell:
What can I do to make you more comfortable? I can provide a list of every AMP element, so we can review which should have a specific rule to forbid?

Note that `.amp-unresolved` is only added by our JS library, pages never include the element by default. The class is only added to our AMP elements by the library when it initializes, and the library removes it when the associated JS downloads. So we can be reasonably sure that only the elements you select will need to be hidden.

ryanbr:
No actual harm with .amp-unresolved? It doesn't break any websites currently. I'm approaching this more cautiously given the element is still being hit and reasons why it was added to begin with. Filters are only removed when filters aren't being hit, aka "Dead filters". 

Also the other commit changes requested in this PR.  What exactly is this patch trying to do. Limit the actual patch to what is needed and why its needed rather than sorting/putting into sections. It seems to complicate what the goal of this PR.

jridgewell:
> No actual harm with .amp-unresolved?  

Hiding `.amp-unresolved` causes layout shift for users, because AMP adds `.amp-unresolved` to all AMP elements then removes it when the element's JS downloads. Layout shift is generally a bad thing, which users disliking pages that shift content while they're reading it. Google Search believe this is important enough of a metric that they are adding it to the page's Search Rank. So layout shift annoys users, and lowers the page's ranking.

> What exactly is this patch trying to do.

We're trying to remove `.amp-unresolved.` From what I can understand, https://github.com/easylist/easylist/commit/f1e088a3f00 added it because certain elements we never resolving because the element's JS was blocked. This is an indirect approach, it's better overall to just hide the specific element that was blocked from loading.

So, I went through and found all rules that block AMP JS, and hide the specific elements. I'll undo the reordering.

jridgewell:
As a demo, try loading up https://output.jsbin.com/wiferin/quiet. With adblock off, I see very little (if any) layout shift. With adblock on, the `<h1>` tag shifts from being at the very top of the page to being under the images. This is a negative impact we're trying to avoid.

jridgewell:
We're still interested in getting this merged. We can observe real impact on layout shift metrics caused by this. Everything hidden by `##.amp-unresolved` can be better hidden as `##amp-foo` tag element, so this should have no negative impact once merged.

PR #8887, opened 2021-08-28
P: verkkouutiset.fi (big slider image won't load)

peace2000:
Sample link: https://www.verkkouutiset.fi/petteri-orpon-oikeiston-iskunkestavyys-testissa/

Big slider image won't load, added a fix.


ryanbr:
Remove the "script,third-party," bits not really needed @peace2000

peace2000:
Removed. Yeah they aren't needed, just thinking about performance :D Though probably won't matter in this case. @ryanbr

ryanbr:
Less unused switches the better @peace2000

Issue #8833, opened 2021-08-25
www.dailymotion.com


johansatge:
### List the website(s) you're having issues:

```
https://jsfiddle.net/61kz8g04/1/
```

### What happens?

When embedding a [Dailymotion](https://www.dailymotion.com/) player on a webpage, the whole player is blocked if `?ads_params=` is included in its URL.

Example:

```
https://www.dailymotion.com/embed/video/x82dy9e?ads_params=myAdsParams
```

<img width="200px" src="https://user-images.githubusercontent.com/5699295/130743784-72182b1b-250e-4203-a851-0eb128ae028c.png">

### List Subscriptions you're using:

<img width="200px" src="https://user-images.githubusercontent.com/5699295/130743420-13cb672e-cd8a-4e4a-b9eb-b48e0db3ecca.png">

### Your settings

- OS/version: macOS 11.4
- Browser/version: Firefox 91.0
- Adblock Extension/version: uBlock Origin 1.37.2

### Other details:

Suggested solution: allow usage of the `ads_params` query string with the Dailymotion player, as detailed in the PR #8387.

Thanks!

ryanbr:
Other than a testcase, is there working example 

This could apply to any other filter?  

`https://www.dailymotion.com/embed/video/x82dy9e?ad_size=myAdsParams`
`https://www.dailymotion.com/embed/video/x82dy9e?ad_width=myAdsParams`
`https://www.dailymotion.com/embed/video/x82dy9e?ad_ids=myAdsParams`

johansatge:
Hey @ryanbr,
`ads_params` is the only concerned filter in my opinion, as it's the only query string we officially support in the Dailymotion player.

The other query strings listed in `easylist_general_block.txt` are not supported in the player, as far as I can tell (so having `https://www.dailymotion.com/embed/video/x82dy9e?ad_size=myAdsParams` for instance, would be an integration error)

ryanbr:
Is there an example where it's broken? and (other than the url pasted) where it states `ads_params` is needed, like docs?

johansatge:
Here is an example: https://www.leparisien.fr/video/video-30-kmh-a-paris-les-conducteurs-mitiges-sur-la-mesure-30-08-2021-NOYMC2VTBJCO3M4VLOTJE6Z2KY.php

<img width="200px" src="https://user-images.githubusercontent.com/5699295/131456429-3ae416ab-39fe-4a8c-938d-8e49cbfe54bf.png">

[...]

PR #8925, opened 2021-09-01
Fix Dailymotion allowlist rules

johansatge:
Hello,

This follows issue #8833 and PR #8387

As described in the issue, the suggested fix didn't work for some reason, and was reverted earlier today in 19ef9ee (@ryanbr)

The new allowlist rule, however, is less permissive that the original one (removed initially 2 weeks ago in 9f00d05
```shell
# Previous rule
@@||dailymotion.com/embed/$subdocument,domain=~gum-gum-streaming.co|~gum-gum-streaming.com|~gum-gum-streaming.tv|~mavanimes.co|~streaming-vostfr.com
# Current urle
@@||dailymotion.com/embed/video/*?ads_params=$subdocument
```

Indeed the Dailymotion player can be embedded with the `/embed/video` path, but also `/embed/playlist`.

We are also working on a new player product, currently in private beta, that will be hosted on `geo.dailymotion.com/*.html`.

For thoses reasons, we suggest going with more permissive rules, like the ones suggested in the current PR:

```
@@||dailymotion.com/embed/$subdocument
@@||dailymotion.com/*.html$subdocument
```

Thanks!


ryanbr:
These are too generic. 

`@@||dailymotion.com/embed/video/*?ads_params=$subdocument` maybe `@@||dailymotion.com/embed/?ads_params=$subdocument` ?  But I need examples where its broken

johansatge:
Hello, thank you for the feedback.

Here is a less generic set of suggested rules:

```
@@||dailymotion.com/embed$subdocument
@@||dailymotion.com/player.html$subdocument
@@||dailymotion.com/player/*.html$subdocument
```

Examples of corresponding Dailymotion player URLs:

```shell 
# Regular Dailymotion player iframe
# Documentation: https://developer.dailymotion.com/player/
www.dailymotion.com/embed
www.dailymotion.com/embed/video/x3rdtfy
www.dailymotion.com/embed/playlist/x6bmjt
# Future Dailymotion player iframe, currently in beta
# Documentation: https://dailymotion.github.io/player-embed/
geo.dailymotion.com/player.html?video=x3rdtfy
geo.dailymotion.com/player/x123.html?video=x3rdtfy
```

It's important for us to keep the player URLs in the allowlist without `ads_params`, because [the easylist rules are bundled into Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1189143), and used there to detect and block heavy advertising components.

The Dailymotion player should be excluded from that heavy ads mechanism, with or without `ads_params`, as the player itself is not an ad (and the ads and trackers that the player may generate are already blocked correctly in `easylist_specific_block.txt`, `easyprivacy_trackingservers.txt` and others)

johansatge:
Hello! here is an exemple to confirm the heavy ads issue described above (tested with Chrome 92):

```
https://www.cosmopolitan.fr/after-3-hero-fiennes-tiffin-se-confie-sur-sa-phobie-inattendue,2053264.asp
```


In this screenshot, the Dailymotion player, that is using our new `https://geo.dailymotion.com/player/xgi8.html` URL format, is incorrectly flagged as an ad by Chrome, probably because the allowlist only allows the old iframe format for now. (`https://dailymotion.com/embed/`)

_(Note Chrome doesn't flag the iframe as an ad every time, it may be hard to reproduce)_

Sandydoucet:
Hello, 

Any news on the previous request regarding our new url format (https://geo.dailymotion.com/player/xgi8.html) to add in the allowlist please ?

ryanbr:
Not likely. too generic.

Sandydoucet:
Hi @ryanbr,

Thanks for your feedback on this. 
On previous comment, there's  URL proposal that could be less generic as mentionned by @johansatge + an page example here : 

https://github.com/easylist/easylist/pull/8925#issuecomment-912646219. 

Sandy

johansatge:
Understood, thx for the feedback! Let's close this pull request for now then.

PR #8947, opened 2021-09-02
A: espoo.fi (generic cookie hide)

peace2000:
https://www.espoo.fi/en

The site has changed so cleaned old entries at the same time.


ryanbr:
##[aria-labelledby=cookie-dialog] -> ##[aria-labelledby="cookie-dialog"] @peace2000

peace2000:
Ok I put quotes back to that rule. I left then originally out intentionally for character trimming / optimization reasons. Quotes are not needed at the value part when it contains only characters withing range `a-z`, `A-Z`, `0-9`, `-`, `_` (with exception that the value cannot start with a number or with `-`.

I recently made this optimization for the Finnish list as a suggestion made by @DandelionSprout https://github.com/finnish-easylist-addition/finnish-easylist-addition/pull/329

@ryanbr

PR #8952, opened 2021-09-02
D: pushinpage.com

spirillen:
I suggest we removes `||pushinpage.com^$third-party` as already listed in the easylist_adservers.txt

And the $domain.tld is anyhow only a API frontend

Relates to:
  - https://github.com/easylist/easylist/pull/7552
  - https://mypdns.org/my-privacy-dns/matrix/-/issues/2610


ryanbr:
Fine to leave it

PR #8972, opened 2021-09-03
Anchor javascript: protocol

mjethani:
It's a good idea to anchor `javascript:` if the intention is to match the beginning of the URL.


ryanbr:
https://github.com/easylist/easylist/commit/387ecca888f74c51275a3cad309d8b504b9e7ab3

PR #8981, opened 2021-09-04
P: kauppahalli24.fi (EP breaks content)

peace2000:
https://www.kauppahalli24.fi/ content won't load, added a fix. Problem clearly visible.


ryanbr:
Maybe `@@||inpref.s3.amazonaws.com/sites/$script,domain=kauppahalli24.fi` instead?

peace2000:
@ryanbr Done.

PR #8983, opened 2021-09-05
A: https://www.sogou.com/

shenzhiming88:
image


peace2000:
What is that element? Some kind of "download our app"?

shenzhiming88:
@peace2000 Yes. Well, I don't think it should be blocked

PR #9002, opened 2021-09-06
M: add to Amazon TLDs

Yuki2718:
For a gigantic international site like Amazon, I think EL should should cover all major locale. I confirmed some of these filters are effective on JP locale.


Khrin:
@Yuki2718 It's OK, just can I ask you to add here also a few more domains, where Amazon is also available in English? (namely ae/nl/sg) Then this PR would be ready to be approved.

Yuki2718:
@Khrin Done

PR #9005, opened 2021-09-07
M: fix Matomo Marketplace HeatmapSessionRecording plugin

Findus23:
fixes https://github.com/uBlockOrigin/uBlock-issues/issues/1708

This allows no additional tracking, but makes sure that the HeatmapSessionRecording plugin can be bought here successfully:
https://plugins.matomo.org/HeatmapSessionRecording

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -2178,7 +2178,7 @@
 /headupstats.gif?
 /healthanalytics.js
 /heatmap_log.js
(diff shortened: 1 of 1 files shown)


spirillen:
I'm really against this, as it will open up for "supporting" spyware (https://en.wikipedia.org/wiki/Spyware)....

A more righteous rule would be

```css
/HeatmapSessionRecording/*$important
```

ryanbr:
No harm to be honest, it's limit to only these domains

PR #9037, opened 2021-09-09
P: https://support.nec-lavie.jp/driverlist?searchType=1 (fixes https:…

Yuki2718:
…//github.com/AdguardTeam/AdguardFilters/issues/93984#issuecomment-914734436)
https://github.com/AdguardTeam/AdguardFilters/issues/93984#issuecomment-914734436
Another option is to remove `||confirmit.com^$third-party` from EP.


stm-x:
In my case, I needed to add this rule.
`@@||adobedtm.com^*/satelliteLib-$script,domain=nec-lavie.jp`
Probably because AdGuard doesn't block adobedtm.com, but easyprivacy does.

PR #9071, opened 2021-09-12
A: https://zhidao.baidu.com/question/310094914193578644

shenzhiming88:
This is the share button
image


ryanbr:
`fanboy_social_general_hide.txt:##.share-area`  we already have this as a generic.

shenzhiming88:
I found that it is allowed in `CHN: AdGuard Chinese`
`zhidao.baidu.com#@#.share-area`

PR #9089, opened 2021-09-12
A: https://www.jianshu.com/p/ae0293657a2b

shenzhiming88:
Ref: https://github.com/AdguardTeam/AdguardFilters/pull/89571
The page will collapse the content and when the button to expand the content is clicked, the page will promote its own products.
These rules can automatically expand all content, thus blocking self-promotion
Screenshot_2021-09-12-23-47-23-986_org mozilla firefox

### DIFF
--- a/fanboy-addon/fanboy_annoyance_international.txt
+++ b/fanboy-addon/fanboy_annoyance_international.txt
@@ -333,6 +333,8 @@ baidu.com##.app-side-banner
 sputniknews.cn##.arrowup
 wanda.cn##.backtop
 worldjournal.com##.btn-top
+jianshu.com##.call-app-btn
+jianshu.com##.collapse-tips
 qidian.com##.focus-img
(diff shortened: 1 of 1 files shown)


ryanbr:
`jianshu.com#$#.collapse-free-content { overflow: auto !important; height: auto !important; }` adds yet another scrollbar. It should be tweaked. Also put into fanboy_annoyance_specific_uBO.txt

peace2000:
Only uBO and Adguard support style injection, ABP doesn't. Will have to wait for that: https://gitlab.com/eyeo/adblockplus/abc/adblockpluscore/-/issues/210

shenzhiming88:
> Only uBO and Adguard support style injection, ABP doesn't. Will have to wait for that: https://gitlab.com/eyeo/adblockplus/abc/adblockpluscore/-/issues/210
> 
> You'll have to add your rules to uBO list only. (If those cosmetic rules cause issues on ABP without style injection).

Yes, so I gave feedback on this issue at uBO. I will add these rules when ABP is supported

PR #9150, opened 2021-09-17
P: vr.fi (login dialog blocked)

peace2000:
Cookie dialog hiding rule hides also the login dialog. The issue is that the cookie dialog and login dialog have the exact same name:

`##div[data-testid="modal"]`

And using pseudo selectors such as `:nth-of-type()` won't help here because those elements change their position, depending if you have previously accepted cookies for this site or not. Also opening the login dialog changes things.

So I took a different approach. I'll let `##div[data-testid="modal"]` to hide the dialog but I added another rule: `##div[data-testid="modal"]:not(:has(div[data-testid="cookie-consent-modal"])):style(display: block !important)` to force the (login) dialog to go visible, if the dialog doesn't contain `cookie-consent-modal` as a descendant. The latter rule is indeed a procedural but it doesn't matter, the login dialog will show up very quickly anyway. And there won't be a blink of that cookie dialog at all as it will be hid via standard css rule.

Test link: https://www.vr.fi/en/train-tickets/multi-ticket

Click "Log in"
image

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -87,8 +87,8 @@ bundesanzeiger.de###cc
 interestingengineering.com##.modal-backdrop
 bundesanzeiger.de###cc > #cc_banner
 paihdelinkki.fi###sliding-popup.sliding-popup-bottom
-vr.fi##div[data-testid="modal"]
 aixfoam.*##[id="cc-notification"]
+vr.fi##body > script[src] + div[data-testid="modal"]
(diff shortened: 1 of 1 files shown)


peace2000:
In case someone reads this PR later, these are all the current filters in use for vr.fi:

`vr.fi##+js(aeld, wheel, preventDefault)`
`vr.fi##body:style(overflow: auto !important;)`
`vr.fi##div[data-testid="modal"]`
`vr.fi##div[data-testid="modal"]:not(:has(div[data-testid="cookie-consent-modal"])):style(display: block !important)`

PR #9173, opened 2021-09-20
A: https://www.tumblr.com/

### LINKED ISSUE #9095: www.tumblr.com

`https://www.tumblr.com/dashboard`

### What happens?

No meaningful cosmetic filtering; basically a repeat of #7700

- EasyList 202109130439

### Your settings

- OS/version: macOS 11.5.2 (20G95)
- Browser/version: Firefox 93.0b4 (64-bit)
- Adblock Extension/version: uBlock Origin 1.37.2

### Other details:

Classnames need to be updated to target the elements; previous cosmetic filters can be assumed to be entirely broken unless they avoid targeting via classnames entirely. I've written a script [here](https://gist.github.com/AprilSylph/275bf9ba829fd5f3bdfb632f549b81da) for producing up-to-date rules. Current output:

```
www.tumblr.com##.j3_kY
www.tumblr.com##.JD7Cz
www.tumblr.com##.VdRef
www.tumblr.com##.WFhoV
```

I'm not completely sure if there's any performance improvement for any blocker to use `www.tumblr.com` rather than `tumblr.com` but I would imagine so. (Any cosmetic filtering targeting `www.tumblr.com` can be assumed to be useless for all other subdomains, since they are exclusively either internal domains which serve only raw files or individual Tumblr blogs which do not use the same web app as the main site.)

AprilSylph:
hides ad containers; fixes #9095; removes outdated rules

Screen Shot 2021-09-20 at 11 36 36

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -1938,6 +1938,7 @@ lasvegassun.com##.varWrapper
 ktm2day.com##.vc_wp_text
 www.tumblr.com##.j3_kY
 www.tumblr.com##.JD7Cz
+www.tumblr.com##.VdRef
 www.tumblr.com##.WFhoV
 ggrecon.com##.venatus-block
 fox10phoenix.com##.vendor-unit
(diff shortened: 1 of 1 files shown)


AprilSylph:
@ryanbr review please? people are complaining to me about ads because EasyList has not added these filters yet and uAssets/AdGuard's classname-free solutions do not cover everything.

AprilSylph/XKit-Rewritten#285

PR #9179, opened 2021-09-20
P: iltalehti.fi (anti adblock nag, whitelist to be undetected)

peace2000:
`https://www.iltalehti.fi/iltv-paivarinta/a/efc6d8b4-0806-4696-a6b3-1b652c7d9964`

If filter `##.ad-placement` is in use, the site will detect adblockers. Whitelisting this element makes adblocking undetected. Whitelisting that element won't show any ads.

kuva

`Tämä sisältö on sinulle maksuton. Maksuttomuus on mahdollista ilmoitustuloilla, joten siksi tämän videon katsominen edellyttää Adblocker-ohjelman poistamista.`

Translation:
`
This content is free of charge for you. Freeness is possible with notification revenue, which is why watching this video requires removing an adblocker program.`

### DIFF
--- a/easylist/easylist_allowlist_general_hide.txt
+++ b/easylist/easylist_allowlist_general_hide.txt
@@ -39,6 +39,7 @@ flat-ads.com,job.inshokuten.com#@#.ad-content
 transparencyreport.google.com#@#.ad-icon
 flat-ads.com,lastpass.com#@#.ad-img
 guloggratis.dk#@#.ad-links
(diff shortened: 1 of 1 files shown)


peace2000:
Hi, @Khrin  why this was refused? This just whitelists a cosmetic element which is used to check adblocking status. Easylist already has similar entries:

https://github.com/easylist/easylist/blob/master/easylist/easylist_allowlist_general_hide.txt

kuva

PR #9205, opened 2021-09-23
P: https://www.sui.com/tally/new.do

shenzhiming88:
These elements should not be blocked
image


shenzhiming88:
@Khrin

Khrin:
What's written on this bar? What's its purpose? I can't access to this page, a login is required.

shenzhiming88:
> What's written on this bar? What's its purpose? I can't access to this page, a login is required.

They are action buttons, such as select all and arrange, etc.

PR #9270, opened 2021-09-29
A: ylasatakunta.fi (cookie specific hide uBO)

peace2000:
https://www.ylasatakunta.fi/

Specific hiding rule for uBO, due to blinking issue with generic filter.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -22,6 +22,7 @@ schulze-immobilien.de##*:style(filter: none !important)
 gov.lv##body:style(overflow: auto !important;)
 !
 backmarket.de##._3NAusrrr
+ylasatakunta.fi##.ag_cookie_banner
 vuokraovi.com##.alma-data-policy-banner
 bing.com##.bnp_cookie_banner
 elisaviihde.fi##.ea-cookie-disclaimer


spirillen:
Evening CookieMonster aka @peace2000 wouldn't it be possible to make this one global

```css
##.ag_cookie_banner
```

peace2000:
Evening. It's already a global filter but it blinks occasionally for this site in uBO. Sometimes you'll have to add a specific filter for uBO to avoid blinking issues with elements. That's because uBO doesn't inject all many thousands of generic filters unconditionally on each site like ABP does. It's an important performance optimization. Though it has a drawback: elements occasionally blink in some websites if they don't have a specific filter and only have a generic.

spirillen:
Hey thanks for that info... that might explain why my rules sometimes flicker for mypdns...

peace2000:
Very likely. Official source for that info: https://github.com/gorhill/uBlock/wiki/Does-uBlock-block-ads-or-just-hide-them%3F#cosmetic-filters

peace2000:
Also it's possible to make uBO to inject a cosmetic filter unconditionally by adding an asterisk before it: https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#specific-generic

Though that should nout be overused :D

spirillen:
Super thanks for the info @peace2000 very useful tricks

peace2000:
Great that it was useful, I struggled with the same issue for a long time before I was told about this... :D

PR #9297, opened 2021-10-03
A: theverge.com (cookie banner hide for non video articles)

peace2000:
In this site, cookies have to be accepted in order to see embedded videos. But as not all articles have embedded videos, it's not needed to whitelist the whole site, only those articles that actually have videos (one browsing the site might not always read articles that have videos, those articles are a minority.).

I know this filter blinks sometimes but it will be fixed when browsers start to support CSS4 (and native `:has`). But even a small blink is better than not to hide that big annoying banner at all.

Article having an embedded video (banner won't be hid):
https://www.theverge.com/2018/8/21/17761424/

Article not having an embedded video: (banner will be hid)
https://www.theverge.com/22705148/windows-11-upgrade-beta-release-preview-impressions


peace2000:
Fyi @thetomaso you have sent issue reports concerning this site before. So I'm notifying :)

Issue #9274, opened 2021-09-30
/reporting/metrics blocked - too generic


samfitz66:
A core part of my teams SAAS product (not available to the public as it's a B2B product) is broken for any users with ublock origin enabled due to the blocking of the /reporting/metrics path (https://github.com/easylist/easylist/blame/642655f94666745692f6ecc693b0391e66cb3677/easyprivacy/easyprivacy_general.txt#L3540) as part of this recent change - https://github.com/easylist/easylist/issues/9233.

Please consider a more specific URL path to block for the citi.com issue, as /reporting/metrics is too generic and is likely going to block other key functionality of other sites in addition to our own.

lcaille:
Hi,

We also got this issue in one of our B2B products as well.

ryanbr:
Closing, filter is still good. Nothing has broken. If we had any replicated issues maybe would be reviewed.

samfitz66:
> ng has broken. If we had any replicated issues maybe would be reviewed.

@ryanbr We have issues in our B2B app as a result of this, it's not in the open web as its a SaaS product, how do you suggested we fix it? I've tried to help fix it by raising https://github.com/easylist/easylist/pull/9333. Happy to do a call and screenshare to show you the issue if you would like evidence.

Khrin:
If it's really an issue, I believe you can whitelist yourself your domain.

Issue #9233, opened 2021-09-27
New tracker: contents2.00110.citi.com


yokoffing:
### List the website(s) you're having issues:

`https://www.citi.com/`

[OISD Domain Blocklist](https://www.reddit.com/r/oisd_blocklist/comments/pwkgw1/oisd_domain_blocklist/?sort=new) in my NextDNS logs show `contents2.00110.citi.com` as a first-party tracker. I think it's a third-party tracker masking as a first-party request.

If I'm understanding it correctly: `wup-edc5180c.us.v2.we-stats.com` becomes `contents2.00110.citi.com`.

### Your settings

- OS/version: 
- Browser/version: 
- Adblock Extension/version: 

### Other details:

Proposed filter: `||contents2.00110.citi.com^`

PR #9333, opened 2021-10-07
Improve specificity of citi.com tracker block

samfitz66:
Improves the specificity of the block issued for https://github.com/easylist/easylist/issues/9233, while taking into account concerns raised in https://github.com/easylist/easylist/issues/9274

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -1552,7 +1552,7 @@
 /Diagnostics?hit=
 /didna-pixel-
 /diffuser.js
-/digital//reporting/metrics
+/digital/reporting/metrics
 /digital-data-enrichment.
 /DigitalAnalytics.umd.
 /digitalice_tag.js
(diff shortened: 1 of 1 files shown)


ryanbr:
Closing, filter is still good. Nothing has broken. If we had any replicated issues maybe would be reviewed.

iam-py-test:
> Improves the specificity of the block issued for #9233, while taking into account concerns raised in #9274

Provide a url which is broken so people can investigate...

samfitz66:
I can't provide a URL for you to test as our affected URL is part of our SaaS product, but here's a screenshot:

image

Would really appreciate if you could merge this to make the block more specific. Of course we could change the naming of our internal endpoints, but this will be considerable effort, so would really help us if you could change on your end instead, as otherwise we need to tell all our clients to disable ublock in order for our app to work (which doesn't benefit either of us). Thank you 🙏

iam-py-test:
Is there a way to create a test account and send the credentials to the Easylist team?
(I don’t have any access to this repo and don’t make decisions here either)

samfitz66:
Can do if necessary, but they can see from the screenshot what would need to be changed to fix it and I believe this patch would work, it would fix the issue this is causing us without affecting the original Citi block.

@ryanbr - let me know if you would like me to set you up with a test account or have a demo to show this issue, or if the above screenshot is sufficient.

PR #9338, opened 2021-10-08
A: [NSFW] https://www.youporngay.com/

CbumDino:
Filter for hiding ads at [youporngay](https://www.youporngay.com/). 

Proposed filter - adding the domain to an existing filter: `##[src^="bLob:"]`

<details open>
<summary>NSFW screenshots</summary>
Screenshot 2021-10-08 at 10 32 27
Screenshot 2021-10-08 at 10 32 38
</details>


Yuki2718:
Maybe all YP sites can be added? `you-porn.com,youporn.com,youpornru.com`

PR #9348, opened 2021-10-09
A: kotipizza.fi (cookie banner hide)

peace2000:
https://www.kotipizza.fi/tuotteet/rullat/kebabrulla

Note, the banner is only visible on the first page load. If you refresh the page, it will always disappear (need to clear cache to see it again).

kuva

### DIFF
--- a/easylist_cookie/easylist_cookie_international_specific_hide.txt
+++ b/easylist_cookie/easylist_cookie_international_specific_hide.txt
@@ -663,6 +663,7 @@ ifolor.fi##.dialogBanner--align-bottom
 timma.fi##.modal-backdrop.in
 lauttakyla.fi##.modalElement
 lidl.fi,pelit.fi,telia.fi##.notification
+kotipizza.fi##.notification-bottom-margin.notification-warning
 helmet.fi##.notifier_warning
 suomi24.fi##.s24_cc_banner-wrapper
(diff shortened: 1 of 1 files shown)


ryanbr:
Probably cause issues @spirillen

peace2000:
I want to make sure that not all possible "warnings" on this site get blocked. They could show "warning" banners at the top, which could be other than 🍪 banners.

And as ryanbr stated I also think that a generic filter for "warnings" would cause false positives.

@shenzhiming88 @spirillen

peace2000:
I btw ate that stuff yesterday that is shown in the picture 😄

peace2000:
@spirillen Ok :)

Though you should use both forms:

`##.notification-warning`
`##.notifier-wrapper`

and

`##^.notification-warning`
`##^.notifier-wrapper`

If you use generic filters, because not every time elements appearing on websites are pre-listed on pages source code. Sometimes they are generated afterwards the page has been loaded.

If you want to use ##^ as a generic filter, it should only be used as a way to "strengthen" normal cosmetic filters.

spirillen:
Hey @peace2000 , @ryanbr , @shenzhiming88 

This far I haven't  seen any issues by blaclisting according to @peace2000 suggestion in https://github.com/easylist/easylist/pull/9348#issuecomment-939477278

PR #8317, opened 2021-07-07
P: pelaaja.fi (link can't be clicked, Easyprivacy)

peace2000:
Original issue report: https://github.com/finnish-easylist-addition/finnish-easylist-addition/issues/306

PR #9413, opened 2021-10-16
P: pelaaja.fi (link can't be clicked, Easyprivacy)

peace2000:
Reference https://github.com/easylist/easylist/pull/8317

https://pelaaja.fi/

Even if that uBO specific fix is not added, EP still needs this whitelist rule anyway to unbreak opening of the link in the yellow bar:

kuva

@ryanbr


peace2000:
@ryanbr Done

PR #9415, opened 2021-10-16
A: maxgaming.fi (specific cookie hide for uBO)

peace2000:
https://www.maxgaming.fi/

Specific filter to overcome blinking issues.


spirillen:
> smile
> 
> I was looking to buy some cable adapters... this is an IT stuff store, not a newspage. stuck_out_tongue

Oh, you missunderstod the news, it was a metaphor for where you might find them.

Never bind to Finland, but here the is a newsstand with tabloid at any bakery :wink: and this was your fourth or fifth cookie today :wink:

PR #9470, opened 2021-10-23
A: wolt.com (cookie banner hide)

peace2000:
https://wolt.com/

Not worth adding as a generic filter: https://publicwww.com/websites/%22ConsentsBannerOverlay%22/


krystian3w:
```
###app div[class*="GdprConsentsModal"]
###app div[class*="ConsentsBanner"]
```

PR #9502, opened 2021-10-27
 A: bosch-home.* (cookie list, uBO)

peace2000:
https://www.bosch-home.fi and all it's other numerous top level domains.

Rule `bosch-home.*##.cookielaw-modal-layer` is needed at least in the Finnish site, probably for others as well.

`bosch-home.*##[class="o-cookielaw"]` is needed at least here: https://www.bosch-home.com/?clear=1#/Togglebox=region-toggle-4/ (probably for other TLD's as well)

`bosch-home.*##+js(rc, pinned|cookielaw-blur-background, body, stay)` will remove blur effect and overflow hidden properties from the body element.

`bosch-home.*##body:style(overflow: auto !important)` (this works faster than +js so it's included as well)

`bosch-home.*##.g-container, .feedback-button:style(filter: none !important)` Is added in order to avoid seeing visible blinking of blurring effect (+js not always fast enough to remove it).
This site and it's TLD's use blur effect on many elements and menus (pics from the Finnish TLD):

kuva

kuva

(Imo, not necessary to apply `:style(filter: none !important)` to each element in these sites, only for those that need it.)

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -1,5 +1,6 @@
 ! uBO Sepecific fixes
 ! :style fixes
+bosch-home.*##.g-container, .feedback-button:style(filter: none !important)
 mtvuutiset.fi###onetrust-consent-sdk:style(display: none)
 mtvuutiset.fi##.container:has(.consent-blocking-embed-message) ~ #onetrust-consent-sdk:style(display: block !important)
 yle.fi###yle-consent-sdk-container:style(display: none)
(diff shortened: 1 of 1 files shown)


peace2000:
@Khrin Many top level domains weren't covered:

https://www.bosch-home.bg/
https://www.bosch-home.ee/
https://www.bosch-home.gr/
https://www.bosch-home.ie/
https://www.bosch-home.lv/
https://www.bosch-home.pt/
https://www.bosch-home.pl/
https://www.bosch-home.si/
https://www.bosch-home.rs/
https://www.bosch-home.com.tr/

instead of adding many domains individually, whyt not to use `bosch-home.*` instead as uBO supports using asterisk? That would be a lot easier to handle.

PR #9517, opened 2021-10-28
A: https://9anime.to/

felippesantos46:
Filter for blocking adserver responsable for redirect popups at [9anime](https://9anime.to/)

Proposed filter: `||kolkwi4tzicraamabilis.com^$popup`

How to reproduce: 
As the webpage won't allow you to open the devtools, the way is to try to play the video using the VideoVard server, after that you will be redirect to the following 
Screenshot 2021-10-28 at 13 14 24
Screenshot 2021-10-28 at 13 06 43


spirillen:
empty commit https://github.com/easylist/easylist/commit/0bb76873b004311a541918b83cf6c602edecd11e

PR #9536, opened 2021-10-31
A: etsy.com - Hide cookie dialog

Nojuuu:
`https://www.etsy.com/`


Nojuuu:
I also think that this file needs better organisation at some point, at the moment it looks a bit messy.

Nojuuu:
Noticed some issues.

PR #9540, opened 2021-10-31
A: etsy.com - Hide cookie dialog

Nojuuu:
`https://www.etsy.com/`

Fixed https://github.com/easylist/easylist/pull/9536. 

The problem was that the page "shrank" slightly each time you expanded the preview image.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -83,6 +83,7 @@ hrkgame.com##.dimmer
 imoradar24.ro##.show.modal-cookies
 uniroyal.pl##.js-cookie-banner > div
 wtk.pl###GDPR-modal
+etsy.com###gdpr-single-choice-overlay
 craftserve.pl###rodo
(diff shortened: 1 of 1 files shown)


peace2000:
@Nojuuu 

EL Cookie's uBO list currently has this rule:
`etsy.com##body,html:style(overflow: auto !important; position: initial !important;)`

If I add this to my custom filters:
`etsy.com###gdpr-single-choice-overlay`

I see no issues at all. Images and even videos seem to work without an issue. https://www.etsy.com/fi-en/listing/751174027/55140-cm-wide-light-grey-linen-curtain?sts=1

Do you know this site better? Can you confirm?

peace2000:
Hmm... it seems to be possible to prevent the shifting of the body element via:

`etsy.com##body:style(padding-right: 0 !important)`

Nojuuu:
Not sure, closed this while back since it didn't seem to get merged and there started to be a lot of changes to the file so I would have had kept rebasing it.

peace2000:
https://github.com/easylist/easylist/pull/12776

Added a fix and it was merged.

PR #9626, opened 2021-11-07
A: [NSFW] https://www.maturetubehere.com/

felippesantos46:
Filters for blocking ads and Popunders + remaining placeholders at [maturetubehere](https://www.maturetubehere.com/)

Proposed filters: 
```
||maturetubehere.com/nb/bac_loa.php
||maturetubehere.com/ai2/*/supc.php
###lotal
###nopp
```

PS: There are couple of filters with /ai2/ , i believe they could go like : /ai2$domain=amateur8.com|maturetubehere.com|sortporn.com|bigtitslust.com|freeporn8.com

PS2: There are some with just /ai/ as well (4x)
Screenshot 2021-11-07 at 08 26 55
Screenshot 2021-11-07 at 08 00 57

<details>
<summary>Screenshot page</summary>

</details>

### DIFF
--- a/easylist_adult/adult_specific_block.txt
+++ b/easylist_adult/adult_specific_block.txt
@@ -136,6 +136,8 @@
 ||madmen.alastonsuomi.com^
 ||mansurfer.com/flash_promo/
 ||matureworld.ws/images/banners/
+||maturetubehere.com/ai2/*/supc.php
+||maturetubehere.com/nb/bac_loa.php
 ||milffox.com/ab-s.js
 ||milffox.com/ai/s/s/js/ssu.
 ||milffox.com/ai/s/s/su.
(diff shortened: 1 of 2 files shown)


Yuki2718:
`nb/bac_loa.php` is a child of `nb/frot_lod.js`, better to block the cause. Also no need to use (slower) wildcard in this case. (`/s/s/supc.php` has been in uBlock filters tho I don't know if EL author want to add the generic filter).

felippesantos46:
Hey @Yuki2718 you are more than right about the first one. We already have `||amateur8.com/nb/frot_lod.js` so i believe we could go with what you purpose: `nb/frot_lod.js` in a more generical way, what do you think ? 
Agreed as well with the second one, a generic filter could be a better solution.

felippesantos46:
@Yuki2718 I've also found this one that differs a little bit from the previous : https://www.porngem.com/

Screenshot 2021-11-08 at 15 10 32

On this matter is it not better to block the whole `/ai2` rather than `/s/s/supc.php` ?

Thank you in advance !

Yuki2718:
> what do you think ?

I don't fully know EL authors' filter preference, it's thier decision so just wait for their response.

> On this matter is it not better to block

On AdGuard or uBlock Origin all of them are covered by
```
/s/js/ssu.v2.js?v=
/s/su.php?t=
```

PR #9629, opened 2021-11-07
A: [NSFW] https://txxx.com/

felippesantos46:
Filter for blocking popunders at [txxx](https://txxx.com/)

Proposed filter: `||txxx.com/diarrhoeatablets/`

Screenshot: 
Screenshot 2021-11-07 at 09 14 58


spirillen:
Seems to work for me

PR #9641, opened 2021-11-08
A: stackexchange.com (cookie specific hide uBO)

peace2000:
Specific hiding filter for uBO to overcome blinking issues.

Sample domains:
https://android.stackexchange.com/
https://coffee.stackexchange.com/
https://apple.stackexchange.com/


ryanbr:
How does it make it ubO specific @peace2000 ?

peace2000:
`##.js-consent-banner` blinks occasionally for stackexchange.com based sites so a specific filter is needed for uBO to fix that issue. ABP injects all generics unconditionally so it doesn't need this but as uBO doesn't inject generics unconditionally, it sometimes needs specific filters.

PR #9663, opened 2021-11-10
M: [NSFW] (Fix) https://freeadultcomix.com/

felippesantos46:
Filter for fix broken webpage + adding adserver responsible for redirect popups

Proposed filters: 
M: `||freeadultcomix.com` to `||freeadultcomix.com^$third-party`
A: `||hentaimotel.com^$third-party`

Screenshots:
Screenshot 2021-11-10 at 10 12 35

Screenshot 2021-11-10 at 10 24 27

### DIFF
--- a/easylist_adult/adult_adservers.txt
+++ b/easylist_adult/adult_adservers.txt
@@ -106,6 +106,7 @@
 ||fncnet1.com^$third-party
 ||forkjork12.com^$third-party
 ||freakads.com^$third-party
+||freeadultcomix.com^$third-party
 ||freevideos24h.com^$third-party
 ||freewebfonts.org^$third-party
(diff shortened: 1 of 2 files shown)


easylist_adult/adult_adservers.txt
```diff
@@ -127,6 +128,7 @@
 ||getiton.com^$third-party
 ||ggwcash.com^$third-party
 ||golderotica.com^$third-party
+||hentaimotel.com$third-party
```
spirillen, line 131: It would be better with `a.hentaimotel.com` otherwise you might end up blocking things you wouldn't block. and this domain is just yet another dumb WP template: https://mypdns.org/my-privacy-dns/porn-records/-/issues/1496

PR #9705, opened 2021-11-13
A: (nsfw) book.dmm.co.jp

Yuki2718:
`||cn.dap.dmm.co.jp^` is blocked in JPF (not pure tracking server) but it'll be worth their tracking pixel blocked in EP as well.

<details>
<summary>Screenshot</summary>

dmm

</details>

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -2316,6 +2316,7 @@
 /imp?*&sz=
 /imp?*&u=
 /imp?imgid=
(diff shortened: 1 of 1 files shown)


spirillen:
Didn't see this rule in the [Tor browser](https://www.torproject.org/download/), but found this annoyance(?)

Banner below "Product"

```css
book.dmm.co.jp##.l-areaDetailBanner
```

Yuki2718:
> but found this annoyance(?)
> 

I don't see that element, maybe need VPN?

PR #9733, opened 2021-11-15
A: ruohonjuuri.fi (stats server)

peace2000:
https://www.ruohonjuuri.fi/


spirillen:
Why not go all in?? The one or the other would in both cases serve the same purpose, while my suggestion while handle any potential futuristic name changes.

```
||userneeds.com^$third-party
```

peace2000:
True, changed.

@spirillen

PR #9785, opened 2021-11-20
A: bloomberg.com (generic cookie block)

peace2000:
https://www.bloomberg.com/europe

Also in https://kotaku.com/ - better to make this generic as there are probably other sites using this same dialog as well.


peace2000:
Not a good call. Videos take ages to load if cookies not accepted. Sample page: https://www.bloomberg.com/news/videos/2021-11-19/reaction-from-kenosha-after-rittenhouse-found-not-guilty-video

PR #9805, opened 2021-11-22
A: [NSFW] https://fapopedia.net

felippesantos46:
Block adserver related to redirect popups at [fapopedia](https://fapopedia.net)

Proposed filter: `||creptdeservedprofanity.com^$third-party`

Screenshots:
Screenshot 2021-11-22 at 16 28 26


Yuki2718:
This belongs to a known revolving adserver class so should be the section in easylist_adserver.txt without `$third-party`.

PR #9815, opened 2021-11-23
A: (ouedkniss.com) #9813

smed79:


easyprivacy/easyprivacy_specific_international.txt
```diff
@@ -105,6 +105,8 @@
 ||liberation.fr/newsite/js/cmp/
 ||ma-petite-recette.fr/visites
 ||neko-san.fr/stats
+||ouedkniss.com/static/alexa.js
+||ouedkniss.com/static/pixel.js
```
spirillen, line 109: Why did you not make there more generic as

```css
/static/alexa.js
/static/pixel.js
```

I believe I would have done that.

PR #9820, opened 2021-11-23
A: kuntsari.fi + other sites (generic cookie block)

peace2000:
https://www.kuntsari.fi/

Used in some Finnish sites: https://publicwww.com/websites/cdn.gravito.net%2Fcmp%2F/


spirillen:
You gets me here... I don't see any cookie on the site `kuntsari.fi`

image

peace2000:
There's a cosmetic filter that hides the dialog but script blocking is better option (it also overcomes blinking issues with the generic cosmetic filter).

spirillen:
agree

PR #9859, opened 2021-11-26
A: https://www.huya.com/102491

DeepChirp:
These are the elements that interfere with the user's normal viewing of the video.
image


shenzhiming88:
@ryanbr

PR #9864, opened 2021-11-26
A: vpd.fi (specific cookie hide for uBO)

peace2000:
https://www.vpd.fi/seagate-game-drive-ps4-ulkoinen-kiintolevy-2-tb.html

Specific filter needed to overcome blinking issues with the generic filter.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -34,6 +34,7 @@ uusisuomi.fi,vuokraovi.com##.alma-data-policy-banner
 bing.com##.bnp_cookie_banner
 calvinklein.*##.ck-modal--cookieModalMain
 elisaviihde.fi##.ea-cookie-disclaimer
+vpd.fi##.ec-gtm-cookie-directive
 bbc.com##.fc-consent-root
 elisa.fi##.react-navi-ea-cookie-disclaimer
 pringles.com##.truste_box_overlay


peace2000:
@Khrin This is still needed as the generic filter is too slow on uBO on this this site. Specific filter fixes the issue that the cookie banner element flashes once before it's hidden.

Here's s screenshot of that banner:

kuva

Khrin:
There's no real need for duplicates.

peace2000:
@Khrin 
On uBo, there is a need sometimes to add a specific filter because the generic filter doesn't work quickly enough and the element quickly flashes once before it's hidden. That's because uBo doesn't inject unconditionally all thousands of generic cosmetics filters on each page, but checks the site first if there's a need for for a specific filter. This is good for performance, but occasionally due to this, the element will appear shortly before it's hid. In comparison, ABP always injects all thousands of cosmetics in each page.

https://github.com/gorhill/uBlock/wiki/Does-uBlock-Origin-block-ads-or-just-hide-them%3F

```
Just like ABP, uBO will hide DOM elements on a web page which match cosmetic filters. uBO uses a [different method](https://github.com/gorhill/uBlock/wiki/Cosmetic-filtering-in-uBlock:-version-0.4.0.0-update)](url) than other big-name blockers to hide the DOM elements though.

Different classes of cosmetic filters are applied differently:

    - Specific cosmetic filters: injected before page's root DOM is loaded
        No flickering
    - Generic cosmetic filters: injected after page's root DOM is loaded
        These are cacheable
        Since these cosmetic filters are applied after page load, DOM elements may be hidden after they are rendered
            Notice the emphasized "may": generally, you won't see this: the slower the computer, the higher the likelihood
            Solution for when the behavior occurs and is an issue for a user: to create the appropriate specific cosmetic filter.
   - Cached generic cosmetic filters: injected before page's root DOM is loaded
        No flickering
```

ryanbr has previously accepted all my PR's in scenarios in these situations:
https://github.com/easylist/easylist/pull/11686
https://github.com/easylist/easylist/pull/9448

PR #9863, opened 2021-11-26
A: reevoo.com (generic cookie hide)

peace2000:
https://www.reevoo.com/shopping

PR #9867, opened 2021-11-27
Append to #9863

spirillen:
Appending to https://github.com/easylist/easylist/pull/9863#issuecomment-980471513


Khrin:
Too much generic.

PR #9897, opened 2021-11-30
A: [NSFW] pictoa.com

felippesantos46:
Block popunders (exoclick)

Proposed filter: `||pictoa.com/38.js`

Screenshot:
Screenshot 2021-11-30 at 10 24 50


spirillen:
There is still a onclick popunder redirect....

PR #9900, opened 2021-11-30
P: https://www.mustijamirri.fi/

Nojuuu:
Hides cookie dialog/background overlay and fixes scrolling.


Nojuuu:
@peace2000  @:ing you in case  you want to add temporary fix

Nojuuu:
I made a small change because this seems to be the actual behaviour of the site after you accept cookies.

peace2000:
Added a temporary fix.

peace2000:
For ABP it's needed to add whitelistings, uBO can handle.

Nojuuu:
Had to remake: https://github.com/easylist/easylist/pull/9918

PR #9901, opened 2021-11-30
A: (nsfw) hqbabes.com

Yuki2718:


Yuki2718:
Hmm, on my end the rule is no more useful.

spirillen:
That was a short lived rule...

One new rule we can us use is, to at least remove the link to external sources.

```css
hqbabes.com##li[id*="trades"] 
hqbabes.com##[href^="/tt/"] # All links starting with the /tt/ is links to 3rd party sites
```

These rules are far from perfect and should be taken as initiators for further rule development.

PR #9918, opened 2021-12-01
P: https://www.mustijamirri.fi/

Nojuuu:
Hides cookie dialog/background overlay and fixes scrolling on uBO.

Whitelists it for ABP because the scrolling cannot be fixed there.


ryanbr:
What is with the `,html` change to the filters?

Nojuuu:
Probably messed something up when trying to fix conflicts?

Nojuuu:
I checked and it seems that those changes exist in the file at the moment, so I didn't make them myself.

Nojuuu:
Navbar is supposed to follow when you scroll down with the fix in https://github.com/easylist/easylist/commit/6fa9152f3c87817fd30510b17d675e870c3be829#diff-f13f723ecf5d69ef289811a7dd2eced2a49e6fff6408ad10eb5e8ff51abc134f it does not.

Thats why `body,html:style(height: auto !important; overflow: auto !important)`

Also reason for using `##[class^="privacy_headsup"]` was that it targets both background overlay and parent element of the dialog itself with just one filter and that it wouldn't matter for uBO users if those allow filters are included in other files for other blockers.

PR #9707, opened 2021-11-13
A: businessinsider.com (generic cookie block)

peace2000:
https://www.businessinsider.com/federal-court-affirms-block-on-biden-vaccine-rules-says-overreach-2021-11?r=US&IR=T

Videos work at the sample site properly, there's a video at the end of the article.

PR #9931, opened 2021-12-03
A: tori.fi (specific cookie block)

peace2000:
https://www.tori.fi/


easylist_cookie/easylist_cookie_specific_block.txt
```diff
@@ -6,6 +6,7 @@
 ||asus.com/js/*/alert-info.js
 ||cmp.infopro-digital.com^
 ||cmp.osano.com^
+||cmp.tori.fi^$domain=tori.fi
```
spirillen, line 9: I dont see why you are making this rule this complex.

```
cmp.tori.fi.    300     IN      CNAME   cdn-1047.privacy-mgmt.com
```

I do block all of `privacy-mgmt.com` I do therefor not need this one, so why not just add it as

`||cmp.tori.fi^` ?

peace2000:
`||cmp.tori.fi` is only needed for `tori.fi`, probably it won't make a big difference but I think it performs better when it's made domain specific instead of generic. Imo it's good practice to make it generic only if it's needed.
 
ryanbr warned against making `privacy-mgmt.com` generic:

https://github.com/easylist/easylist/pull/9707

spirillen:
Yeah some think it is problematic to block this in a generic form, I happens have a other view on this topic, but that we all know by now. :wink: 

That is also why I'm structuring it as one issue = one domain so that users how where to look for comment about a blacklisted record, for them self to determine if they will allow the *Ware on there network. It is to my opinion not up to anybody else to control.

Never forget the [F.A.Q.](https://mypdns.org/my-privacy-dns/matrix#faq)

# FAQ
**Q**: Your lists have broken my website, by blocking a third party
domain!

**A**: We have absolutely not broken your website by blocking any
third-party domain.
It's you and only you who has submitted bad code and mis-written your
website to behave badly. Be happy we haven't blocked your website
already for leaking privacy to third party.

**Q**: Why don't you whitelist example.com?

**A**:

1) It's up to individual users (i.e you) to create and maintain their own
   whitelist.

2) By whitelisting bad domain X because website Y uses bad domain X,
   we're deceiving our users who believe they're protected against bad
   domain X.

3) Through this repository's issue board, we're trying our best to give
   people the right information for them to decide whether or not they
   should allow access to some domains, despite them being marked as bad."

peace2000:
Fixed conflicts

easylist_cookie/easylist_cookie_specific_block.txt
krystian3w: Chromium may rare use `privacy-mgmt.com` also completly no support CNAME.

krystian3w:
Conflicts back.

PR #9938, opened 2021-12-05
A: https://www.prosenectute.ch

salim-b:
Question: Why don't we just add `###cookie` to [`easylist_cookie_general_hide.txt`](https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_general_hide.txt)? Would it cause too many false positives?


Khrin:
@salim-b Yes

PR #9987, opened 2021-12-09
A: animebee.to

felippesantos46:
Hide popup messages at https://animebee.to/

Screenshot 2021-12-09 at 06 09 57


ryanbr:
Another way will be implemented.

felippesantos46:
Hey @ryanbr, can you provide this other way ? I have some of those cases in ES and PT Lists. 
Thank you in advance.

PR #10014, opened 2021-12-11
A: kicks.fi (cookie banner hide)

peace2000:
https://www.kicks.fi/


Khrin:
It doesn't seems to be necessary anymore.

PR #10036, opened 2021-12-13
globsads.com Appendix to #10034

spirillen:
Appending to https://github.com/easylist/easylist/pull/10034

External related:
- https://mypdns.org/my-privacy-dns/matrix/-/issues/3865


spirillen:
@ryanbr why make a second PR when you merge my code without altering it? First seems like it is to avoid giving me the credit ( I hope this is not the case)

PR #10042, opened 2021-12-14
A: https://www.watchseries1.video/

felippesantos46:
Block adservers at https://www.watchseries1.video/

Screenshots:
Screenshot 2021-12-14 at 06 01 21
Screenshot 2021-12-14 at 06 00 59


felippesantos46:
There is also some other popunders, solved by: `watchseries1.video#$#abort-current-inline-script document.querySelectorAll popMagic`

PR #10083, opened 2021-12-16
A: https://animedao.to/

felippesantos46:
Block adservers responsible for redirect popups

Screenshots:
Screenshot 2021-12-16 at 11 59 53
Screenshot 2021-12-16 at 11 55 28


Khrin:
`||mradx.net^` is not an Adserver (Probably is also used as CDN)

Also the first filter needs to be corrected

felippesantos46:
Hey @Khrin, I believe the second filter came by mistake on this "bundle"
The first one was corrected as well. 

Thank you !

PR #10135, opened 2021-12-17
A: https://animasu.vip/

felippesantos46:
Block adserver responsible for redirect popups at https://animasu.vip/

**Tested as mobile**

Screenshot:


felippesantos46:
**Duplicated** 
#https://github.com/easylist/easylist/pull/10134

PR #10217, opened 2021-12-22
A: https://watchsouthpark.tv/

felippesantos46:
Block adserver at https://watchsouthpark.tv/

Screenshot:
Screenshot 2021-12-22 at 14 54 50

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -10527,6 +10527,7 @@
 ||prevailedballotcaution.com^
 ||previouslyawokestale.com^
 ||previsibilisais.website^
+||prevotch.com^
 ||preyjustly.com^
 ||prfctmney.com^
 ||prfdesk.pro^


felippesantos46:
Hey again @spirillen ! I've tried to use my own template, but no idea if right 😅, should i just address the issue or also open a PR in here ? 
Thank you in advance

spirillen:
Perfecto :confetti_ball: only one little tiny thing

image

only the domain... (and then you can take it with the blockquetes here as requested by the easylist team 

https://github.com/easylist/easylist/blob/2b5e3cec861fadfc6e2665926c2aaa3bc32532ae/issue_template.md?plain=1#L17

image

The reason we should not use protocols is to avoid any misconceptions as a domain can be attached to any protocol and it can be several and changed over time.

> or also open a PR in here 

No need I take care of than when it is added to the dns server for the RPZ hosting. The issue is fine, thanks. and now I can save about 3 minutes per issue :partying_face: :tada:

PR #10242, opened 2021-12-23
A: https://ytmp3.plus/

felippesantos46:
Hide fake buttons at https://ytmp3.plus/

Screenshot:
Screenshot 2021-12-23 at 10 22 50

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -651,7 +651,6 @@
 ###aboveplayerad
 ###abovepostads
 ###acAdContainer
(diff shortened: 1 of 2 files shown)


Yuki2718:
That's too risky and a bait for detection. `||ytmp3.plus^$subdocument` seems fine.

felippesantos46:
> That's too risky and a bait for detection. `||ytmp3.plus^$subdocument` seems fine.

Agreed, i didn't realize the Iframe before 😅, Thank you @Yuki2718 !

PR #10281, opened 2021-12-25
A: (nsfw) buondua.com (fixes https://github.com/AdguardTeam/AdguardFi…

Yuki2718:
…lters/issues/104456)
https://github.com/AdguardTeam/AdguardFilters/issues/104456
Ad is reproduced only on Chromium.


spirillen:
I still see some ad placeholders

spirillen:
Found a back to top

`##.bottom-hidden-bar`

Yuki2718:
I don't see that request. Anyway `||buondua.com/hit/?` for EP.

PR #10293, opened 2021-12-27
A: https://dramaindo.xyz/

felippesantos46:
Hide banner at https://dramaindo.xyz/

Screenshot:
Screenshot 2021-12-27 at 10 16 03


spirillen:
Why not make it `dramaindo.xyz###teaser` so when they learn to count to 4? :smile: 

Also I found this on the right side, not entirely sure what it is, but css title indicates some third party "services"

`##.textwidget`

image

PR #10327, opened 2021-12-29
A: https://www.exprssnews.com/

felippesantos46:
Hide side bannners at https://www.exprssnews.com/

Screenshot:

Screenshot 2021-12-29 at 14 44 01


Yuki2718:
I tend to suggest reinstate `_300_250_`. The removal of this has been causeing false negative:
https://github.com/AdguardTeam/AdguardFilters/commit/6f3b17486560873e401a3dadcd930a6ff0fce98f
https://github.com/AdguardTeam/AdguardFilters/issues/104064#issuecomment-999512221

spirillen:
Sounds like a good idea @Yuki2718 I second that one. That would be a good generic filter.

PR #10334, opened 2021-12-30
Reduce overblocking

salim-b:
Moves a too unspecific hiding rule from general to specific file, fixing i.a. the manifesto signing on https://aktionvierviertel.ch/manifest/

### DIFF
--- a/easylist_cookie/easylist_cookie_allowlist_general_hide.txt
+++ b/easylist_cookie/easylist_cookie_allowlist_general_hide.txt
@@ -228,7 +228,6 @@ e-dym.pl,earthsky.org#@#.cp-popup
 earthsky.org#@#.cp-popup-container:not(body):not(html)
 seguro.elespanol.com#@#.didomi-notice-banner
 myrobotcenter.at,myrobotcenter.co.uk,myrobotcenter.de#@#.ec-gtm-cookie-directive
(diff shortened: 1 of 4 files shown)


Yuki2718:
Such a PR would require at least a few real examples of breakage.

salim-b:
Signing the manifesto on https://aktionvierviertel.ch/manifest/ is a real example (link target of the repsective button is https://aktionvierviertel.ch/manifest/#elementor-action%3Aaction%3Dpopup%3Aopen%26settings%3DeyJpZCI6IjI2MDciLCJ0b2dnbGUiOmZhbHNlfQ%3D%3D )

Yuki2718:
I mean just one example will not be enough to convert a generic filter.

salim-b:
I see. I just figured from the CSS selector `##.elementor-popup-modal.dialog-widget` that it is probably used for non-cookie popups on various other sites, too, since *Elementor* is simply an [open-source (WordPress) website builder](https://github.com/elementor/elementor/).

If you decide to not merge this PR, I'll submit an exception rule for https://aktionvierviertel.ch instead. Just let me know.

Yuki2718:
> 
> If you decide to not merge this PR, 

To be clear, I'm not an EL maintainer so has no right to decide. Just saying.

Issue #10303, opened 2021-12-28
asahi-kasei.com: cookie


hirorpt:
### URL address of the web page

`https://www.asahi-kasei.com/`

### Category

cookie

### Describe the issue

~~~
asahi-kasei.com###box-cookie-01
~~~

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

![Screenshot](https://user-images.githubusercontent.com/82331005/147547984-f5bd093b-e277-4327-92fd-6f17a024808d.png)

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.40.2
Firefox: 95
filterset (summary): 
  network: 79794
  cosmetic: 59503
  scriptlet: 16430
  html: 612
listset (total-discarded, last updated): 
  added: 
    fanboy-cookiemonster: 20912-14, now
  default: 
    ublock-filters: 30652-37, now
    ublock-badware: 3635-1, now
    ublock-privacy: 181-0, now
    ublock-abuse: 81-0, now
    ublock-unbreak: 1721-0, now
    easylist: 62308-571, now
    easyprivacy: 26558-483, now
    urlhaus-1: 7867-0, now
    plowe-0: 3679-3, now
filterset (user): [empty]
modifiedUserSettings: [none]
modifiedHiddenSettings: [none]
supportStats: 
  launchToReadiness: 446 ms
popupPanel: 
  blocked: 2
```
</details>

PR #10367, opened 2022-01-03
Address asahi-kasei.com cookie consent

iam-py-test:
This PR fixes https://github.com/easylist/easylist/issues/10303
I used a specific rule as I wanted to avoid breakage, however, I can switch to a generic one if needed


iam-py-test:
Thanks @ryanbr

PR #10381, opened 2022-01-04
P: Fix cnn[.]com

iam-py-test:
This PR fixes https://github.com/uBlockOrigin/uAssets/issues/11194 as recommended by @mapx-


iam-py-test:
Already fixed

PR #10409, opened 2022-01-06
A: seatgeek.com (cookie hide for uBO)

Nojuuu:
https://seatgeek.com/

Hides cookie dialog, background overlay and fixes scrolling.


peace2000:
I would create an issue report about this and suggest filters to be added because the rule `##body:style(overflow: auto !important;)` gets constantly edited, causing then issues to existing pull requests that try to edit the same line.

@Nojuuu

ryanbr:
rebase @Nojuuu

Nojuuu:
@ryanbr done

PR #10443, opened 2022-01-10
Remove InMoment.com Domains

githubib:
By blocking the javascript on the sites, the surveys are being broken and the users cannot submit their surveys to let us know about their experiences.  The sites listed are closed to the public and are invite only.  Those invitees are therefore unable to participate and have a diminished experience.  InMoment doesn't use scripts to collect any personal information.  Thank you.

Survey Landing Page (Without reference to EasyPrivacy)
image

Survey Landing Page (With EasyPrivacy list enabled)
image

The domains in Question (https://github.com/easylist/easylist/search?q=inmoment.com)


githubib:
This is not necessary.  Issue resolved with bug request.  Thank you.

PR #10504, opened 2022-01-17
M: https://m.dict.cc/

salim-b:
Block left-over ad containers on mobile site.


Khrin:
@salim-b I believe these rules can be added as generic filters in https://github.com/easylist/easylist/blob/master/easylist/easylist_general_hide.txt

salim-b:
@Khrin Ok, see https://github.com/easylist/easylist/pull/10506

PR #10306, opened 2021-12-28
block gelbooru header ad

Znunu:
This banner is blocked atm, but leaves an empty space behind. This should take care of that. I hope edited the correct file. I just looked for existing gelbooru entries
image

PR #10540, opened 2022-01-21
hide realbooru ad on gelbooru

Znunu:
I tried coming up with a more general rule, but couldn't get it to work. Fortunately this seems like the only ad left,

image

On a sidenote;
i see that other PRs are following some convention w.r.t. title (e.h. "A:"). I cant find where it says what this convention exactly is


Znunu:
thank you ♡( ◡‿◡ )

gelbooru:
https://github.com/easylist/easylist/pull/10306

I question this commit as it was previously explained to me that our store and any self promotion is allowed. Does this not fall under this provision?

PR #10549, opened 2022-01-21
A: https://9anime.se/

felippesantos46:
Hide banner at https://9anime.se/

Screenshot:
Screenshot 2022-01-21 at 13 51 54


Khrin:
In all likelihood, this is self-promotion. I wouldn't block it for now.

felippesantos46:
> In all likelihood, this is self-promotion. I wouldn't block it for now.

Oki, thank you for the review !

PR #10633, opened 2022-01-27
A: [NSFW] https://www.cambro.tv/

felippesantos46:
Block banners at https://www.cambro.tv/

Screenshot:
Screenshot 2022-01-27 at 12 18 30


spirillen (commented):
You might checkup with this issue for `domains=`

You might find more to add

PR #10683, opened 2022-01-31
Restore Angular Material dialog backdrop

PowerKiKi:
Because the selector is too wide, it will break legit behavior of all apps
that use Angular Material and their dialog systems. This include standard
modal dialogs, but also datepickers, auto-completion and so on.

With this selector, it would make it impossible to close the dialog by clicking
outside of it.

This is testable on the library official documentation: https://material.angular.io/components/dialog/overview. Clicking on "Pick one" will open a dialog, and clicking anywhere outside of the dialog should close it.

image

### DIFF
--- a/easylist_cookie/easylist_cookie_general_hide.txt
+++ b/easylist_cookie/easylist_cookie_general_hide.txt
@@ -7785,7 +7785,6 @@
 ##.ccwrap
 ##.cczcook_banner-wrapper
 ##.cd__cookie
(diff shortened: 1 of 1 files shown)


PowerKiKi:
This would fixes https://github.com/easylist/easylist/issues/10841

stissot (approved):
Problem confirmed and PR fixes it as expected

PowerKiKi:
Was fixed in 5940ebe9b4f0ac8cdcc71adec28dbf0290a38930

PR #10685, opened 2022-01-31
A: https://123moviesme.online/

felippesantos46:
Block adserver at https://123moviesme.online/

Screenshot:
Screenshot 2022-01-31 at 16 37 28


BottledSoda:
@felippesantos46 I'm curious, is there any particular reason you're using Adblock Plus?

PR #10825, opened 2022-02-09
A: https://www.tumblr.com/dashboard

AprilSylph:
Screenshot 2022-02-09 at 09-42-02 Tumblr


ryanbr:
What is with tripadvisor addition included in this tumblr patch?

AprilSylph:
@ryanbr That was a merge commit, the full diff vs master is only one line

PR #10876, opened 2022-02-13
P: https://mubi.com

chaptergy:
The currently used `appear-enter-done` class is used in a lot of places in mubi, and everything with this animation is removed when this filter list is active. It is unclear whether the automatically generated classname I replaced it with changes over time, but it is better to see the cookie banner then instead of not being able to start watching a movie.

Fixes https://github.com/easylist/easylist/issues/10877


Yuki2718 (approved):

Yuki2718:
Can confirm false positive. This is actually a P (= Problem) commit.

chaptergy:
Thanks for the info, force pushed, to change the commit message as well.

PR #10897, opened 2022-02-17
A: [NSFW] https://mangadna.com/

felippesantos46:
Block adserver at https://mangadna.com/

Screenshot:
Screenshot 2022-02-17 at 11 12 17


spirillen:
Thanks for both domains

PR #10918, opened 2022-02-18
A: [NSFW] https://motherless.com/

felippesantos46:
Block hosted adserver at https://motherless.com/

Screenshot:
Screenshot 2022-02-18 at 12 01 10


Yuki2718:
It's in EP.

Khrin:
@felippesantos46 That's a tracker and is already in EasyPrivacy.

PR #11047, opened 2022-02-25
A: https://www.macstadium.com

salim-b:
Is there a good reason for `##.cookie-banner` not being part of generic hide list?


ryanbr:
`##.cookie-banner:not(body):not(html)` @salim-b

PR #11114, opened 2022-03-02
A: https://4stream.gg/

felippesantos46:
Block adserver at https://4stream.gg/

Screenshot:
Screenshot 2022-03-02 at 13 12 19


bogachenko:
_I can't find the commit I need, so I'm writing here._

@Khrin `$popup,third-party,domain=...|4stream.gg|...` This rule breaks the site `http://www.fromhots.com/` it is impossible to click on the streaming video links

PR #11130, opened 2022-03-04
A: https://www.detik.com/

felippesantos46:
Hide placeholder at https://www.detik.com/

Screenshot:
Screenshot 2022-03-04 at 10 41 39


Khrin:
@felippesantos46 I think this filter can be added as generic one.

felippesantos46:
> @felippesantos46 I think this filter can be added as generic one.

Done 👍 !
Thank you !

PR #11188, opened 2022-03-08
A: http://watchhouseonline.net/

felippesantos46:
Block adserver at http://watchhouseonline.net/

Screenshot:
Screenshot 2022-03-08 at 10 52 50


Khrin:
Looks the same as https://github.com/easylist/easylist/pull/11193?

PR #11235, opened 2022-03-10
M: Removed coin-hive.com

smed79:

PR #11238, opened 2022-03-11
A: #11235

smed79:


ryanbr:
Okay, reasons for keeping it now its not a coin miner? @smed79

smed79:
added to avoid the redirection !!

PR #11242, opened 2022-03-11
A: #11235 (coin-hive.com)

smed79:


ryanbr:
Any reason to keep it? @smed79

smed79:
Avoid loading the script on sites who self host the script, ex `https://coins4days.com/lib/coinhive.min.js`

PR #11298, opened 2022-03-15
A: bguiok.xyz

pattyland:
Published only scam, including on https://alicemarissa.bguiok.xyz


Khrin:
Emoty site and anyway probably it shouldn't be blocked by EasyList.

PR #11300, opened 2022-03-15
A: lovelyglrls.com

pattyland:
Published only scam, including on https://jbvng.lovelyglrls.com


Khrin:
Show only `Site is under construction, please visit later` and anyway probably it shouldn't be blocked by EasyList.

PR #11308, opened 2022-03-16
A: https://animebee.to/

felippesantos46:
Block redirects + Inside banners at https://animebee.to/

Screenshot:

Screenshot 2022-03-16 at 07 01 27
Screenshot 2022-03-16 at 06 59 05


Yuki2718:
`##.xad-wrapper` is for iframe, not `animebee.to`

PR #11416, opened 2022-03-23
A: sfanytime.com (cookie banner hide)

peace2000:
https://www.sfanytime.com

They change element names - this old one doesn't work: https://github.com/easylist/easylist/pull/7296/

That's why I suggested a procedural filter.


Khrin:
@peace2000 I haven't tested the filter, but anyway it should go in https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_specific_ABP.txt

peace2000:
@Khrin but will uBO load filters from that list? This filter works in ABP and in uBO / AG as well.

Is it necessary to write a separate filter for each even though one filter would work for all?

PR #11510, opened 2022-03-29
A: (nsfw) amateur-vids.eu

smed79:


spirillen (changes requested):
Removes the player as well

image

Yuki2718:
Yes, the rule hides the whole player.

PR #11516, opened 2022-03-31
A: https://streamingcommunity.icu/

felippesantos46:
Block image popup at streamingcommunity.icu

Screenshot:
Screenshot 2022-03-31 at 10 51 45


felippesantos46:
> This should stay in EL Italy.

True, my bad !

PR #11546, opened 2022-04-02
A: playpilot.com (cookie banner block)

peace2000:
https://www.playpilot.com


ryanbr:
no need for the $script or $domain bit @peace2000

peace2000:
@ryanbr Yeah it's not technically "needed" but it should perform better as that rule only activates on a specific domain and it only checks script-based requests.

But, I removed them as asked.

PR #11553, opened 2022-04-03
A: https://www.androidauthority.com/piracy-opinion-3145442/

ajayyy:
Example at https://www.androidauthority.com/oneplus-10-pro-review-android-authority-3145591/

image

image

Progress status was also blocked because it is now in the middle of the page

image


Khrin:

ajayyy:
Would it make sense under annoyances?

PR #11657, opened 2022-04-14
A: https://www1.solarmovies.movie/

felippesantos46:
Block image popups at https://www1.solarmovies.movie/

Screenshots:
Screenshot 2022-04-14 at 08 23 35


Yuki2718:
Is the site alive? No such server found on my end.

PR #11666, opened 2022-04-14
A:republicworld.com

mediumkreation:
User reported issue:
https://github.com/mediumkreation/IndianList/issues/5002


Khrin:
@mediumkreation I think that `###floatAd` can be added as generic filter.

mediumkreation:
@Khrin Updated as suggested.

PR #11707, opened 2022-04-20
A: http://www.cakrawalanews.co.id/

smed79:
https://github.com/ABPindo/indonesianadblockrules/pull/318


gmarcher:
it's specifically for Indonesian, it should be fixed in the regional filter, right?

Khrin:
I believe so, this cannot stay in EL.

PR #11741, opened 2022-04-21
A: Hides banner from realgearonline.com

sabinahschmitz:
Screenshot 2022-04-21 at 10 20 29

Hides top & bottom banner src="http://www.adpeepshosted.com/adpeeps.php"


Khrin:
it doesnt seems to be necessary? at least for me...

ryanbr:
Yeah, can confirm, its not needed

PR #11743, opened 2022-04-21
A: Hide ad with adserver repairalmostanything.com

sabinahschmitz:
Screenshot 2022-04-21 at 10 54 41

- hide top banner (href="https://www.ipnoid.com/")


Khrin:
Looks like more a tracking service to me.

Issue #10645, opened 2022-01-28
unsplash.com - links are incorrectly blocked/removed


OliverJAsh:
### List the website(s) you're having issues:

`unsplash.com`

### What happens?

(I work at Unsplash.)

Since https://github.com/easylist/easylist/commit/7badbbf1f78ef19775097017fbbd79a69e0e8514#diff-d95e1ad0368a5fb3bacb89fc7f209b9fa33a9974d9d2cf4b9a98903b7aa356a2, the EasyList filter blocks `unsplash.com##.eziW_`. /cc @ryanbr

`.eziW_` is our `resetAnchor` class which is a very generic class used almost everywhere across the site for styling links. As a result, many links are blocked/removed across the site.

Some examples:

![image](https://user-images.githubusercontent.com/921609/150767719-71b49802-1e51-464b-a25d-a9b76604c7c2.png)

![image](https://user-images.githubusercontent.com/921609/150767725-973351cf-2343-40e9-9edb-81a1fb919559.png)

![image](https://user-images.githubusercontent.com/921609/151344320-1ab4ca61-2f2d-4d7b-8318-acd483ebecd5.png)

![image](https://user-images.githubusercontent.com/921609/151345651-67a8350f-d331-4392-95cd-d54ccef2a84a.png)

![image](https://user-images.githubusercontent.com/921609/151345667-d3aef916-c1b3-4d21-9573-3c34aeb062d2.png)

### Your settings

- OS/version: 
- Browser/version: 
- Adblock Extension/version:

OliverJAsh:
@ryanbr What was the intention of your change in https://github.com/easylist/easylist/commit/7badbbf1f78ef19775097017fbbd79a69e0e8514 to block classes `HjpWP` and `eziW_`? I work at Unsplash and I would be happy to provide a hook/selector for the specific element you're trying to target.

I know we show an ad on `/t/wallpapers` but I don't think it's necessary to block these classes for that because it's already blocked by the filter for `a[href*="adzerk.net"]`:

https://github.com/easylist/easylist/blob/f142038998d7426fb4e9ecc50f55c26402674838/easylist/easylist_general_hide.txt#L22538

OliverJAsh:
Fixed by https://github.com/easylist/easylist/commit/79714187fd585eb39148204e6e0841a41fabf93f by the looks of it

OliverJAsh:
[...]

PR #11744, opened 2022-04-21
Remove bad filter for unsplash.com

OliverJAsh:
Fixes https://github.com/easylist/easylist/issues/10645

cc @Edivania

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -1476,7 +1476,6 @@ mobafire.com##.exhibit-vertical
 streamingmedia.com##.expand
 yourbittorrent.com##.extneed
 thefinancialexpress-bd.com##.extraLink
-unsplash.com##.eziW_
 formula1.com##.f1-dfp-banner-wrapper
 audiophilereview.com##.f7e65-midcontent
 audiophilereview.com##.f7e65-sidebar-ad-widget


OliverJAsh:
@ryanbr Thanks! Is there any way we can prevent https://github.com/easylist/easylist/issues/10645 from happening again? This is the second time it's happened this year.

edivaniadc:
> @ryanbr Thanks! Is there any way we can prevent #10645 from happening again? This is the second time it's happened this year.

@OliverJAsh and @ryanbr as removing `unsplash.com##.eziW_` ([previous case](https://github.com/easylist/easylist/commit/7badbbf1f78ef19775097017fbbd79a69e0e8514#diff-d95e1ad0368a5fb3bacb89fc7f209b9fa33a9974d9d2cf4b9a98903b7aa356a2)) the `Ad label` is back, could you please test whether one of below custom filters would block other styling links?
`unsplash.com##.LBDpZ.eziW_`
`unsplash.com##.RAaYj`

Just wanna make sure it won't be another problematic filter so it could be a way to prevent future issues in trying to hide this Ad label. Thanks in advance.

al2
al1

OliverJAsh:
@edivaniadc Let me take a look. If there isn't already a class that targets that item specifically then I can add one for you. Does it have to be a class or can it be any selector?

edivaniadc:
> @edivaniadc Let me take a look. If there isn't already a class that targets that item specifically then I can add one for you. Does it have to be a class or can it be any selector e.g. an attribute?

Thanks @OliverJAsh it could be either a class or a selector.
The most important is to be something more specific so it would not block any part of the content.

OliverJAsh:
I've just deployed a change that adds an attribute called `data-ad` to the container element. This CSS would hide the entire ad, including the ad label and any padding beneath.

```css
[data-ad] {
    display: none;
}
```

edivaniadc:
> data-ad
Thank you so much!
Just waiting for the changes to go live so I can test again.
unp

OliverJAsh:
Try now, I just purged the cache.

edivaniadc:
> Try now, I just purged the cache.
@OliverJAsh is still the same - Ad label: `.LBDpZ .eziW_ `  > `.RAaYj` 
https://unsplash.com/t/history
https://unsplash.com/t/current-events

OliverJAsh:
Hmm, this is what I'm seeing. Try clearing your browser cache?

image

image

edivaniadc:
@OliverJAsh just [added it](https://github.com/easylist/easylist/pull/11767). Thanks a lot.

PR #11763, opened 2022-04-22
A: https://www.manga-raw.club/

felippesantos46:
Hide placeholder at https://www.manga-raw.club/

Screenshots:
Screenshot 2022-04-22 at 11 01 35


Khrin:
it doesn't seems worthwhile to be added

PR #11809, opened 2022-04-25
A: (dacast.com) #11808

smed79:


smed79:
@Khrin @ryanbr Don't you think it's a good generic as :

https://github.com/easylist/easylist/blob/675ad5ec8ec743312027ffe8bd0a62667908c2e5/fanboy-addon/fanboy_social_general_hide.txt#L10897

PR #11821, opened 2022-04-26
A: Hide banner on minecraft.fandom.com

sabinahschmitz:
Screenshot 2022-04-26 at 10 08 44
Hiding banner with ##.top-leaderboard


Yuki2718:
Already hidden by `##.top-ads-container`

PR #11850, opened 2022-04-28
M: Ads on techadvisor.com

sabinahschmitz:
Screenshot 2022-04-28 at 12 13 03

Ads spotted on https://www.techadvisor.com/how-to/windows/record-pc-sound-3800679/ - 
Ad ID seems to have changed.


edivaniadc:
It looks like those are cv ads but @Khrin @sabinahschmitz can you please check if `.pkcTGn` would hide for you?
thanks

PR #11887, opened 2022-05-02
A: [NSFW] https://supjav.com/

felippesantos46:
Block popunders at https://supjav.com/

Screenshot:
Screenshot 2022-05-02 at 10 29 13


Yuki2718:
It's better to block `||support.streamjav.top^` (already blocked in AG Base) which is a CNAME-disguised adserver of AdSpyglass (` 0i0i0i0.com`). `/pBUKvw1.js` part can change easily.

PR #11898, opened 2022-05-03
Add regex101.com analytics filter

migueldemoura:
The domain `analytics.regex101.com` hosts a Plausible Analytics instance
from where analytics scripts are loaded from and metrics sent to.


Khrin:
I see it already blocked by `/plausible.js`

ryanbr:
Yeah,, not needed. already blocked @migueldemoura

migueldemoura:
Indeed, we don't need this one, thanks!

PR #11899, opened 2022-05-03
Create rickroll.txt

moongazer07:
added rickroll links


moongazer07:
add rickroll link list so rickrolls are blocked
note: needs formatting so only the link itself gets blocked not the entire website for example youtube

Khrin:
Surely your efforts would be appreciated, but this isn't the correct place for such a list.

PR #11918, opened 2022-05-06
P: https://www.google.*/maps/* (Brave-specific CNAME issue)

Yuki2718:
On Google map click bottom-left Layers => More and Globe view will be partly broken:

map

map1

This seems to be a CNAME issue by Brave's uncloak, adding `@@||gstaticadssl.l.google.com^$domain=google.co.jp|google.com` to custom filter fixes it on my end. The responsible rule is `||gstaticadssl.l.google.com^$~font,third-party,~xmlhttprequest` and `third-party` doesn't help due to cname uncloaking:

cname

On uBO + Firefox it doesn't happen thx to `@@||google.*^$cname` in Unbreak, but apparently Brave can't interpret `$cname`. @pes10k is there a plan to support the modifier?

### DIFF
--- a/easylist/easylist_thirdparty.txt
+++ b/easylist/easylist_thirdparty.txt
@@ -212,7 +212,7 @@
 ||groupon.com/javascripts/common/affiliate_widget/
 ||grscty.com/images/banner/
 ||gsniper.com/images/
-||gstaticadssl.l.google.com^$~font,third-party,~xmlhttprequest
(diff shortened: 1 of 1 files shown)


Yuki2718:
The rule came from unreliable arguments in https://www.reddit.com/r/uBlockOrigin/comments/8lyb5n/dealing_with_googleadapislgooglecom_and/ and has been causing only issues like https://github.com/uBlockOrigin/uAssets/issues/7071 without any confirmed goodies, hence removed.

PR #11948, opened 2022-05-08
A: intercom.io

adam891213:
This is an annoying chat widget that appears in the lower right corner of the screen.
a

Usage example:
https://sprint.ly/
https://www.snapeda.com/
https://www.solved.fi/
https://www.streak.com/


ryanbr:
While annoying, I'd avoid blocking chats because it could be useful for some people. A method of contacting the site without going through emailing them

PR #12078, opened 2022-05-20
A: https://www.razon.com.mx/

edivaniadc:
Hide left ad overlay and ad close button on [razon.com.mx](https://www.razon.com.mx/)
_Did not add `##.stickyad-close` as it would hide the close button only._

Reference https://github.com/easylist/easylist/commit/49ecc17

razon


ryanbr:
Non-english @edivaniadc

PR #12094, opened 2022-05-23
A: Hide ads on instacart.com

sabinahschmitz:
Screenshot 2022-05-23 at 12 04 12


ryanbr:
Should be in `easylist_specific_hide_abp.txt` @sabinahschmitz

PR #12105, opened 2022-05-24
A: Hide item on respawnfirst.com

sabinahschmitz:
Screenshot 2022-05-24 at 12 52 57

VPN to US, EL + ABP
Go to https://respawnfirst.com/best-amd-rx-6950-xt-graphics-cards/
There is small element on the left ([id="post-consent-ui"]). It is unclickable and covers the content.


ryanbr:
Not an advert? Is fixed in ELC and Annoyances

PR #12185, opened 2022-06-01
M: https://www.wowbiz.ro/laurentiu-reghecampf-meci-spectaculos-al-dezvaluirilor-duminica-aceasta-la-40-de-intrebari-cu-denise-rifai-20225521

Branzeturi:
(video)


Khrin:
Not needed.

PR #12283, opened 2022-06-12
Update easyprivacy_trackingservers.txt

amit-o:
coralogix.com is not a tracking company - we're a systems observability company, we don't have a user tracking product

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -764,7 +764,6 @@
 ||cookieless-data.com^$third-party
 ||cooladata.com^$third-party
 ||copperegg.com^$third-party
-||coralogix.com^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
Came from: https://github.com/easylist/easylist/commit/9103d70f7f3f18e4e12aff7502d701bd5870156e

Want examples of `coralogix.com` use before removing. @amit-o

amit-o:
Hey @ryanbr 
You would use it a centralized place to send the logs of your applications to, so when something is down/not working you'd have an easier time checking services status. You can also set alerts, etc.  
We're in the same business space as datadog (disregarding their Real User product), sumologic, logz.io, etc.

ryanbr:
Away from my PC, but if I recall we block many datadog urls.

ryanbr:
Closing, given we're blocking similar `datadog` urls, no reason to allow coralogix

PR #12304, opened 2022-06-15
A: https://www.rock103rocks.com/

edivaniadc:
Banner ads on the following domains:
https://www.rock103rocks.com/
http://95x.com/
https://www.957thevibe.com/

I did not add only `.widget_simpleimage `for 957thevibe.com because it would also hide the **957thevibe banner** that links to:
https://www.957thevibe.com/95-7-the-vibe-all-access


edivaniadc:
No longer needed. It would hide cumulus self-promo

Issue #12294, opened 2022-06-14
Outbrain ads on buzzfeed.com


edivaniadc:
### List the website(s) you're having issues:
https://www.buzzfeed.com/jp/ryosukekamba/lisa4?ref=hpsplash

### What happens?
Outbrain ads on https://www.buzzfeed.com/jp/ryosukekamba/lisa4?ref=hpsplash

_Although page is in Japanese would you please consider blocking these ads in EasyList?
 website also have the English version https://www.buzzfeed.com/_

Scroll down and search for `.recirclist_recircList__` 
<img width="1279" alt="buzz" src="https://user-images.githubusercontent.com/57706597/173629463-ecd34d60-b2a7-4bb9-99b0-d6f7577c770f.png">
All posts have https://traffic.outbrain.com/network/redir? link and then redirect to either buzzfeed.com or other links such as: https://www.descente.co.jp/media/editors_picks/buzzfeed/20783/?bfseed=sodisco&distro_platform=outbrain

### List Subscriptions you're using:
ABP filters + EasyList

### Your settings

- OS/version: macOS
- Browser/version: Chrome 102
- Adblock Extension/version: ABP 3.14

### Suggested filters:
`buzzfeed.com##section[class^="recirclist_recircList__"]`
`||odb.outbrain.com^$script,domain=buzzfeed.com`

PR #12306, opened 2022-06-15
P: https://www.buzzfeed.com/jp/ryosukekamba/lisa4?ref=hpsplash

edivaniadc:
Noticed that filter: `##section[class^="recirclist_recircList__"]` added on [commit](https://github.com/easylist/easylist/commit/6bd03eb) blocks the section called **もっと読む** below Outbrain widget (with ABP disabled) on https://www.buzzfeed.com/jp/ryosukekamba/lisa4?ref=hpsplash

Removed the rule as it could be affecting other sites as well but please let me know if an exception for this particular domain would be better.

Steps to reproduce: 
ABP Lists:  ABP filters + EasyList
VPN location: Japan and US

bzf1

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -24042,7 +24042,6 @@
 ##a[onmousedown^="this.href='https://paid.outbrain.com/network/redir?"][target="_blank"]
 ##a[onmousedown^="this.href='https://paid.outbrain.com/network/redir?"][target="_blank"] + .ob_source
 ##a[target="_blank"][onmousedown="this.href^='http://paid.outbrain.com/network/redir?"]
-##section[class^="recirclist_recircList__"]
 ! Taboola
 ###block-boxes-taboola
 ###component-taboola-below-article-feed


edivaniadc:
@ryanbr when you have the chance, could you please have a look at this issue? thanks

ryanbr:
Whats the actual issue? its addressing https://github.com/easylist/easylist/issues/12294 using the same classes to counter outbrain ads on other sites @edivaniadc

edivaniadc:
> Whats the actual issue? its addressing #12294 using the same classes to counter outbrain ads on other sites @edivaniadc


@ryanbr it turned out that filter: `##section[class^="recirclist_recircList__"]` is also blocking the section called もっと読む (below Outbrain widget) which is actually part of the website content. 
I could search items by item with: `recircitem_stacked__ ` or the whole section with `recirclist_recircList__` 
You can test with Japan or US VPN location.
bz82

So basically `||outbrain.com^$domain=bgr.com|buzzfeed.com|tvline.com` would be enough to block the Outbrain ads and removing `##section[class^="recirclist_recircList__"]` or adding an exception to its domain, would allow the もっと読む section.

PR #12315, opened 2022-06-16
A: https://www.techpowerup.com/

smed79:


ryanbr:
They'll just counter it, we'll just leave it

PR #12342, opened 2022-06-20
A: right banner ad

sabinahschmitz:
Screenshot 2022-06-20 at 12 56 17


Khrin:
Looks like self-promotion

PR #12395, opened 2022-06-24
A: https://digg.com

edivaniadc:
Sponsored ads throughout the page

dg

Other filter suggestion: 
`digg.com#?#.story-flex:-abp-has(div:-abp-contains(DIGG PICKS | SPONSORED))`


monzta:
@edivaniadc instead if `.story-flex` I would recommend selecting `article.fp-vertical-story` like in https://github.com/easylist/easylist/commit/4f31d7a This will also remove the placeholders of the elements

edivaniadc:
> @edivaniadc instead of `.story-flex` I would recommend selecting `article.fp-vertical-story` like in [4f31d7a](https://github.com/easylist/easylist/commit/4f31d7a) This will also remove the placeholders of the elements

@monzta thanks for the suggestion. Here is the update https://github.com/easylist/easylist/pull/12395/commits/c1cdf768b2a7d43decf78963720ec0cdd7f5faeb

PR #12397, opened 2022-06-24
P: https://wevpn.com/billing/order.php

edivaniadc:
EasyList rule: `||postaffiliatepro.com^$popup,third-party` prevents **wevpn.com** links from opening on third-parties.

Sample URL 1 - https://www.vpncompare.co.uk/wevpn-review/ 
click on `Visit WeVPN` buttons
blocked link: https://wevpn.com/billing/order.php?aid=twstdfngrs
vpncompare

Sample URL 2 - https://www.youtube.com/watch?v=2AUfYIbuz50
video description > click on https://www.wevpn.com/aff/techlore
blocked link: https://wevpn.com/billing/order.php?aid=techlore
yt


ryanbr:
I would expect we block some affliate popups, if we create an exception here then why not for other sites. I think it should be left alone @edivaniadc

PR #12490, opened 2022-07-05
A: foreca.fi / telsu.fi + others (cookie script block)

peace2000:
https://www.auto1.fi/

https://www.blogit.fi/

https://www.foreca.fi/

https://ilmainensanakirja.fi/

https://www.suomi24.fi/viihde

https://www.telsu.fi/

https://www.testeri.fi/

`||cmp.quantcast.com` not safe as a generic, only safe as domain specific.

### DIFF
--- a/easylist_cookie/easylist_cookie_international_specific_block.txt
+++ b/easylist_cookie/easylist_cookie_international_specific_block.txt
@@ -72,6 +72,7 @@
 ||browser-consent-front.coco.s-cloud.fi/js/$script
 ||cdn.gravito.net/cmp/$script
 ||cmp.tori.fi^
+||cmp.quantcast.com^$script,domain=auto1.fi|blogit.fi|foreca.fi|ilmainensanakirja.fi|suomi24.fi|telsu.fi|testeri.fi
 ||hsy.fi^*/hsycookie.js
 ||sanoma.fi^*/sccm-b.js
 ||sanoma.fi^*/sccm-c.js


Nojuuu:
@peace2000 

`https://bbs.io-tech.fi/threads/mozilla-firefox-keskustelu.1277/page-42#post-11018371`

related?

Khrin:
Confirmed: https://github.com/easylist/easylist/commit/e6ae618b393086d9076d79d02f2914a815d4a1f6

I do wonder if it's causing problems elsewhere too...

peace2000:
It seems that Firefox own tracking protection also causes issues:

kuva

peace2000:
> @peace2000
> 
> `https://bbs.io-tech.fi/threads/mozilla-firefox-keskustelu.1277/page-42#post-11018371`
> 
> related? I can reproduce that scrolling issue but it happens very randomly and when it does there seems to be `overflow: hidden !important;` in `body`.

When I open Foreca in incognito mode and then refresh the page, I can repro broken scrolling.

Though **foreca.fi##body:style(overflow: auto !important)** seems to fix it.

Are there any other issues than blocked scrolling?

When I tested this page I didn't happen to test with subpages that are that long to actually have a scrolling bar...

@Nojuuu

peace2000:
Hmm maybe it's indeed that stupid surveys that causes the issue after all...

Nojuuu:
> Have you @Nojuuu noticed any other issues than blocked scrolling when the cookie script is blocked? If not, an uBO based fix could be easily made.

Everything should work fine.

peace2000:
Yes, that't the cause of this issue after all, not blocking cmp scripts.

peace2000:
https://github.com/easylist/easylist/issues/12507

PR #12599, opened 2022-07-15
D: OpenTable promoted result

danielireson:
**Source: I work on the product**

We noticed promoted results weren't being shown in search results for users whos subscribe to EasyList. We believe this is a mistake because promoted results are often still the most relevant search result (they're promoted for a reason). This rule was originally added in https://github.com/easylist/easylist/commit/a46931b5df6a5e6dde178af12445efe2bc479fbd.

The default order for search results reflects a number of factors including table availability, restaurant location, search criteria, restaurant reviews and ratings, and restaurant payments.

To give an example of where this would be incorrect, with the rule applied the first search result in the screenshot below is hidden (the result with the promoted tag), however it really shouldn't be. I'm searching for "Indian" in London at 5:30pm for two people and the promoted result is ranked higher in search results because it matches this search criteria (i.e. nearby, cuisine match, good rating, availability at requested time).

image


ryanbr:
I'm not convinced this needs to be reverted, since its ads mixed into search results @danielireson

danielireson:
Hi Ryan, thanks for considering the change. Our argument is that it shouldn't be considered as an advert; restaurants can't pay to guarantee the first search result like you would with a traditional PPC advert going to the highest bidder. In this case, restaurants pay for the potential of a higher search result ranking only when other search criteria is met (such as time or location). The first search result is the one most relevant to the search parameters so hiding it isn't in the user's interests. When we apply the promoted tag it's because one factor of the search ranking is restaurant payments, but it's never the only factor that's considered.

ryanbr:
If it's not considered an advert, why is it `Werbeanzeige` labeled on these links. We're doing similar blocks on amazon.com (and others)

danielireson:
That was considered the best translation from `Promoted` but in any case, it's just a label and the label doesn't affect the search ranking model. Although it's hard to see which category our block falls under from those specified in the readme, I can see similar blocks on sponsored results on Amazon, and so in that case I don't think I'm going to be able to convince you that this is different. I will close this PR based on that, thanks for reviewing my change.

Issue #12507, opened 2022-07-07
Foreca.fi - scrolling gets blocked


peace2000:
### List the website(s) you're having issues:

`https://www.foreca.fi/meteorologilta/o0wnyh6r`

### What happens?

Scrolling goes broken suddenly and there are loading circles as if the page tries to load something:

![kuva](https://user-images.githubusercontent.com/17256841/177817031-b772a601-3df8-408c-a9bf-adf93137c7f5.png)

It's a survey that tries to load.

### List Subscriptions you're using:

Easyprivacy

### Your settings

- OS/version: Win 10
- Browser/version: Firefox 102
- Adblock Extension/version: uBO 1.43.1b3

### Other details:

Open the page always in incognito mode, and then refresh the page once and the issue should appear.

This seems to fix it:
`@@||userneeds.com^$domain=foreca.fi`

After the fix, the survey appears:

![kuva](https://user-images.githubusercontent.com/17256841/177817661-111e9197-4019-4319-b1b9-61102a60b10a.png)

And when I close the survey, scrolling restores. (I first thought the issue was caused by blocking cmp scripts but it was an EP issue after all).

Nojuuu:
@peace2000 
```
||digitalfeedback.euro.confirmit.com/api/digitalfeedback/loader/prod/overlay$xmlhttprequest,domain=foreca.fi
```
Does adding that prevent it completely? I'm not 100℅ sure yet.

peace2000:
> Not seeing it here, is there specific scripts needed? http://webstatustest.dk/ for example had `https://cdn.userneeds.com/unt-lite.js` `https://cdn.userneeds.com/int.js?int=f2lKnHvNgQDb` were blocked (though didn't cause issues)

@ryanbr 

Actually, when checking this more carefully, you'll only need to whitelist the frame so that it won't get blocked.

These requests from userneeds are visible in the logger:

![kuva](https://user-images.githubusercontent.com/17256841/178049663-92927125-8f3a-47d6-8aee-ca29ee35c710.png)

So just this rule is enough:

`@@||survey.userneeds.com^$subdocument,domain=foreca.fi`

It's a bit more narrowed than more initial suggestion and seems to be enough.

Another option is to use

`||digitalfeedback.euro.confirmit.com/api/digitalfeedback/loader/prod/overlay$xmlhttprequest,domain=foreca.fi`

As suggested by @Nojuuu. It does work as well. No more scrolling getting blocked.

[...]

PR #12694, opened 2022-07-25
A: foreca.com (generic survey block, annoyance)

peace2000:
Referring to this message: https://github.com/easylist/easylist/issues/12507#issuecomment-1179505506

EP should not block a survey (it's fixed in EP now), but Annoyance list should block it. Blocking the survey this way won't cause issues.

### DIFF
--- a/fanboy-addon/fanboy_notifications_thirdparty.txt
+++ b/fanboy-addon/fanboy_notifications_thirdparty.txt
@@ -6,6 +6,7 @@
 ||centrnotify.com^$third-party
 ||checke.biz^$third-party
 ||checkpost.club^$third-party
+||confirmit.com^$third-party
 ||conforama-push.com^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
`fanboy-addon/fanboy_annoyance_thirdparty.txt:||confirmit.com^$third-party`  ?

peace2000:
I guess that would be better.

confirmit.com is indeed in wide usage: https://publicwww.com/websites/%22confirmit.com%22/

ryanbr:
Or maybe move it too notifications third-party?

peace2000:
Edited my PR.

ryanbr:
remove it from one list/txt file, move it to the other..

peace2000:
I moved it to fanboy_notifications_thirdparty.txt, isn't that the correct one?

peace2000:
kuva

?

PR #12731, opened 2022-07-28
Add Roblox phone verification warnings

JustWorkBruhh564563:
They are obtrusive, annoying and very distracting. If you havent for example verified your phone, on the home page there is an alert that takes 30% of the webpage asking you to verify your phone number.
Also verifying your phone has no benefits, it's grants you access to nothing. Even worse they make your account vulnerable to SIM jacking, which is a very common attack to compromise high value accounts (commonly referred as "beaming" in the account compromising/hacking community).

And even on top of that, when theres a data breach regarding account info, phone numbers are always included in it. They are not encrypted in any way, all past and current phone numberes are stored forever, and anyone who gains access to the OKTA (the account managment service they use) panel can look at everybodys phone numbers. Even worse, half of the sites users are under 13. It's probably not good if more kids get doxxed because their phone numbers got leaked.
As such I believe these should be added to the privacy filters as it's such a bad privacy risk too.


ryanbr:
screenshot of issue? Also is it bad for security?

PR #12740, opened 2022-07-28
Remove audioeye non-analytics endpoints

mgius-ae:
The two domains and specific JS bootstrapping code provide accessibility/WCAG compliance services to websites.  These javascript files contain the code necessary to provide these services.  These domains only serve static files necessary to do this work.

The cookiestorage html page supports websites that prefer to leave these services opt-in for users. 

air and analytics domains (already found in this list, left intact) handle all of the reporting and analysis.

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -254,8 +254,6 @@
 ||audience.newscgp.com^
 ||audienceapi.newsdiscover.com.au^$third-party
 ||audienceinsights.net^$third-party
-||audioeye.com/ae.js
-||audioeye.com/frame/cookieStorage.html
 ||audit.303br.net^
 ||audit.median.hu^
 ||aurora-d3.herokuapp.com^
@@ -2188,9 +2186,7 @@
 ||wp.com/i/mu.gif$image
 ||wp.com/wp-content/js/bilmur.min.js
 ||wren.condenastdigital.com^
-||ws.audioeye.com^
 ||ws.sharethis.com^$script
-||wsmcdn.audioeye.com^
 ||wtbevents.pricespider.com^
 ||x.babe.today^
 ||x.disq.us^


mgius-ae:
Audioeye previous was added to and then removed from this list.

https://forums.lanik.us/viewtopic.php?p=128216-audioeye-com#p128216

https://github.com/easylist/easylist/commit/258b3d92fa351b921f7eeb86dd619407b2d5e9b5#diff-c7eb20b53d44b1cf3515d142ad3880d4a03b985ef48f3e7807ac07d73b8e2cca

ryanbr:
I have my doubts. `https://wsv3cdn.audioeye.com/frame/cookieStorage.html?build=prod&pscb=`

https://pastebin.com/jtYrZ2gD

I have little trust for audioeye

mgius-ae:
If we leave the cookieStorage block in it would prevent opt-in only sites from working, and I believe would prevent config changes in the visual toolkit from persisting. I can restore the block for this endpoint if it is a concern.

The other three block lines, however, prevent our customer's sites from working as expected.  By leaving the block in place there is an onus on privacy conscious users with accessibility needs to make a choice between privacy and access.  These users would need the technical know-how to understand why sites are not working as intended and how to whitelist AudioEye on their local machines.  I think this is a poor tradeoff given that we restrict all reporting and analytics to specific domains that are easily (and currently) blocked.

jkalberer:
@ryanbr I can't disagree with your concerns since there is definitely the potential for abuse. 

The cookie code in our scripts is only used for storing user settings for our accessibility tool. Users can turn on different modes in the tool that will be applied to sites when loaded (bigger font, tinting colors, etc). These settings are very important for users of our tools and it is ideal if they can persist across different domains.

The code itself is very old at this point and needs to be cleaned up and modernized. We need to come up with a strategy that works with Apple's cookie policies as well as making sure that we're doing things right for privacy in general.

I'll work with the team to come up with a plan around these cookies that tries to solve our use-cases without passing cookies via iframes. 

We'd like to be transparent with you all and unblock the things that are necessary for our services to run which gives users a good experience. Other things like telemetry can remain in your block list.

Anyways, we'll be back once we have an update and we can take a second pass at this.

Cheers!

PR #12762, opened 2022-07-30
A: fazer.fi (join as a member notification)

peace2000:
https://www.fazer.fi/

Pop-up that asks to join as a member to receive benefits. (Not a newsletter so put it to the notifications list).

kuva


peace2000:
@ryanbr This was closed without explanation? Was that an accident or was that on purpose?

Khrin:
You do have a very broad definition of "notification"...

PR #12807, opened 2022-08-03
P: is.fi (video doesn't work)

peace2000:
https://www.is.fi/taloussanomat/art-2000008982700.html

kuva


Nojuuu:
Looks like theres some tracking requests after that is whitelisted

`||gatling.nelonenmedia.fi/hitcounter`
`||sanoma.endpoint.finnpanel.fi/fp/ping.gif`

PR #12814, opened 2022-08-04
Hidden ads using the css filter

AirSayy:
Ads were hidden using the CSS filter on the cnbc website.


ryanbr:
thumBlock is far too generic. so no @AirSayy

Issue #5540, opened 2020-06-12
topplant.pl


krystian3w:
URL:

`http://topplant.pl/`

## what happend:

Empty pop-up with **locked scroll** after use Fanboy Social Blocking List

### Please test without list "EasyList Cookie" becasue have a generic network filters:

> https://github.com/easylist/easylist/blob/f42fa496daf70a505c26230a882a4bd131219faf/easylist_cookie/easylist_cookie_general_block.txt#L824-L825

## screen:

![obraz](https://user-images.githubusercontent.com/35370833/84534816-e9058800-acea-11ea-9a31-8e93fed90c12.png) |
--- |

## what we can to do...:

If we must unlock scroll that needed is duplicate that lines https://github.com/easylist/easylist/blob/f42fa496daf70a505c26230a882a4bd131219faf/easylist_cookie/easylist_cookie_general_block.txt#L824-L825 in  Fanboy Social Blocking List project.

DandelionSprout:
Why not simply remove the overlay completely with `##.popmake-overlay` instead?

krystian3w:
Locked scroll, simple `##.popmake-overlay` no recover scroll:

krystian3w:
We with this recover scrolling:

```adb
topplant.pl##.pum-open:style(overflow: auto !important;)
topplant.pl##html.pum-open body > [aria-hidden]:style(padding-right: 0 !important;)
```

---

krystian3w:
https://github.com/MajkiIT/polish-ads-filter/pull/7468

PR #12849, opened 2022-08-07
A: aamulehti.fi + others (consent dialog hide, uBO)

peace2000:
All these sites are owned by Sanoma Corporation. On these sites, cookies have to be accepted in order to see embedded content.

This set of filters will unlock the scrolling for each page (via `:style`), and hide the consent dialog only, if the page doesn't have any embedded content.

There are two types of sites on this PR: news websites and magazine/forum based.
For news websites, this rule is being used:
`##body:not(:has(.embed-cookie-consent-icon)) > div[id^="sp_message_container"]`

For magazine/forum based:
`##body:not(:has(.consent_required_iframe)) > div[id^="sp_message_container"]`

Note!
This rule should not be used this time:
`##+js(rc, sp-message-open, html, stay)`

Because it suffers from a glitch: even if the consent dialog is hid, it still tries to disable scrolling, leading to a small, horizontal "bounce" effect, which is annoying. User styles have a precedence over site scripts, which is why they work better against scripts.

Some sample links (the upper one doesn't have embedded stuff (dialog will be hid), the lower one does have embedded stuff (dialog won't be hid)):

https://www.aamulehti.fi/tampere/art-2000008989602.html
https://www.aamulehti.fi/kotimaa/art-2000008984691.html

https://www.etlehti.fi/artikkeli/ihmiset/kohta-elakkeelle-nelja-ihmista-kertoo-milta-tuntuu-jattaa-tyotakki-naulaan
https://www.etlehti.fi/artikkeli/vapaa-aika/katso-suomen-vanhin-varielokuva-talta-naytti-helsingin-kauppatori-vuonna-1936

https://www.gloria.fi/artikkeli/sisusta/malli-vilma-bergenheim-ja-huippukiekkoilija-sean-bergenheim-sisustivat
https://www.gloria.fi/artikkeli/huippumallit_musiikkivideoilla_kymmenen_parasta

https://www.hs.fi/ulkomaat/art-2000008989805.html
https://www.hs.fi/urheilu/art-2000008988784.html

https://www.hyvaterveys.fi/artikkeli/mieli/laulaja-laura-voutilainen-teen-sita-mista-innostun-ja-joskus-vauhtia-liikaa
https://www.hyvaterveys.fi/artikkeli/vanhemmuus/kansa-sekosi-aaro-vauvan-kuvasta-teilla-varsin-hyvantuulisen-nakoinen

https://www.is.fi/kotimaa/art-2000008990209.html
https://www.is.fi/politiikka/art-2000008974427.html

https://www.tiede.fi/artikkeli/uutiset/masennuksen-serotoniinihypoteesi-saa-kritiikkia
https://www.tiede.fi/artikkeli/uutiset/halvaantunut-pystyi-kirjoittamaan-ajatuksen-voimalla

https://www.vauva.fi/keskustelu/4596081/ketju/arsyttavin_fiktiivinen_hahmo_ikina
https://www.vauva.fi/comment/42579908#comment-42579908

Edit: Note!
`is.fi` has a separate weather subpage in which the same `:style` that works for others, doesn't work there:
https://www.is.fi/supersaa/

Managed to make a separate blocking filter for that subpage solely:
`||cdn.privacy-mgmt.com^*supersaa$xhr,domain=is.fi`
That weather subpage doesn't have embedded content so it's ok to block GDPR request.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -2042,3 +2042,9 @@ autohero.com##body,html:style(overflow: auto !important; position: initial !impo
 ! tallink.com
 tallink.com###cms-web-consent-root
 tallink.com##.cws-modal-open:style(overflow: auto !important; pointer-events: auto !important;)
+! Sanoma Corporation
+aamulehti.fi,etlehti.fi,gloria.fi,hs.fi,hyvaterveys.fi,is.fi,jamsanseutu.fi,janakkalansanomat.fi,kankaanpaanseutu.fi,kmvlehti.fi,kodinkuvalehti.fi,merikarvialehti.fi,nokianuutiset.fi,rannikkoseutu.fi,satakunnankansa.fi,soppa365.fi,suurkeuruu.fi,sydansatakunta.fi,tiede.fi,tyrvaansanomat.fi,valkeakoskensanomat.fi,vauva.fi##html.sp-message-open:style(overflow: auto !important; width: initial !important)
+aamulehti.fi,etlehti.fi,gloria.fi,hs.fi,hyvaterveys.fi,is.fi,jamsanseutu.fi,janakkalansanomat.fi,kankaanpaanseutu.fi,kmvlehti.fi,kodinkuvalehti.fi,merikarvialehti.fi,nokianuutiset.fi,rannikkoseutu.fi,satakunnankansa.fi,soppa365.fi,suurkeuruu.fi,sydansatakunta.fi,tiede.fi,tyrvaansanomat.fi,valkeakoskensanomat.fi,vauva.fi##html.sp-message-open > body:style(position: unset !important; margin-top: 0 !important)
+aamulehti.fi,hs.fi,is.fi,jamsanseutu.fi,janakkalansanomat.fi,kankaanpaanseutu.fi,kmvlehti.fi,merikarvialehti.fi,nokianuutiset.fi,rannikkoseutu.fi,satakunnankansa.fi,suurkeuruu.fi,sydansatakunta.fi,tyrvaansanomat.fi,valkeakoskensanomat.fi##body:not(:has(.embed-cookie-consent-icon)) > div[id^="sp_message_container"]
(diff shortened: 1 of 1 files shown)


Nojuuu:
I noticed that you also need:
`##html.sp-message-open body:style(margin-top: unset !important)`

otherwise you can't sometimes scroll because of it.

image

krystian3w:
Also maybe 0 instead unset if have same behaviour?

peace2000:
> I noticed that you also need: `##html.sp-message-open body:style(margin-top: unset !important)`
> 
> otherwise you can't sometimes scroll back to the top because of it.
> 
> image

True, I can repro. It seems that if you scroll the page down and then refresh (F5), that issue appears. Fixed now.

Dunno if it matter whether it's `unset` or `0`, but as `0` is the original value, I used it.

krystian3w:
I think pc world family may use same library for pop-up.

krystian3w:
But any time can implement better block load embedded posts and videos.

Nojuuu:
You can just block the script if there's no issues with embeds, these cookie popups are now appearing because original filter was made site specific.

krystian3w:
BTW these rewrite still need filter to generic block stock settings of pop-up maker (pum): #5540

Pum can be used to show any text not only cookies, gdpr & ccpa.

ryanbr:
We have exisitng sp-message filters, https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_specific_uBO.txt#L1939

@peace2000   can it not be added here?

peace2000:
@ryanbr No, because hiding that dialog for these domains is conditional: it's only hid if the article doesn't have embedded content. It's a custom solution for this site cluster only, and can't be merged with existing filters.

Also using `##+js(rc, sp-message-open, html, stay)` is not option, because it doesn't work well with conditional cosmetics: it causes a bounce effect glitch. `:style` is the only thing that works here because user styles have a precedence over site scripts.

peace2000:
And btw concerning those rules you pointed...

If hiding sp message works without and issue (embedded content works), then it should handled through blocking the script: https://github.com/easylist/easylist/blob/da862639b56eaeafa199c0016bee9a8edf17a93b/easylist_cookie/easylist_cookie_specific_block.txt#L77

peace2000:
Linking just in case: https://github.com/easylist/easylist/pull/12988

PR #12901, opened 2022-08-11
A: vpd.fi (generic cookie block)

peace2000:
https://www.vpd.fi/


ryanbr:
Causes too many false positives.

peace2000:
Okay, but make site specific?

PR #12935, opened 2022-08-13
Add znsrc.com to list of thirdparty trackers

MarkSRobinson:
znsrc.com is a tracker used by gem.com.  The whole domain appears to be nothing but tracking with no legitimate website behind it.


ryanbr:
Site where its hosted? @MarkSRobinson

MarkSRobinson:
@ryanbr I'm not sure I understand your question.  The site where `znsrc.com` is hosted?  Where gem is hosted?

ryanbr:
We're not going to add any random server without an example link.

PR #12936, opened 2022-08-14
P: / A: cmore.fi (hide GDPR dialog without breakages)

peace2000:
On https://www.cmore.fi/ you'll have to accept cookies in order to use customer service chat.

kuva

"Hyväksy Toimivuus evästeasetuksista käyttääksesi chattia." Translates to "Accept Functionality on cookie settings in order to use the chat".

This filter set hides the dialog generally, but when the user clicks the button to go to settings:
kuva
The settings dialog will open.

The trick is that when the button is clicked, it also removes `style` attribute from the `.onetrust-pc-dark-filter`, thus negating the hiding filters.


peace2000:
Wait, forgot to add some filters...

PR #12952, opened 2022-08-15
Update easylist_general_hide.txt

AirSayy:
img
Added an  ,,EasyList" filter to the list.


ryanbr:
Too generic, also causing false positives.

AirSayy:
> Too generic, also causing false positives.

> I suspected...thank you..I'll be more careful next time.

PR #12967, opened 2022-08-16
A: tiktok.com (verify element)

peace2000:
When browsing https://www.tiktok.com/@cnetdotcom?language=en the page a while, it shows a needless verify element. I don't see any good reason for it / it doesn't state it's purpose or reason. When I solve it, clear cookies and open the page again, it appears again. No need to do anything peculiar, just browse the page a short while.

kuva


Khrin:
I don't think that's a smart idea to block their captcha, regardless of how much is annoying or useless in this case...

peace2000:
Ok it seems that this causes issues - additional videos won't load so I'll close this. Noticed this just now...

Issue #12973, opened 2022-08-17
Alter or remove Koko Analytics filter


dannyvankooten:
Hello,

000296b3f67137aee4b86c8d739e52670e21b6d6 / #12864 added a filter to block all requests containing `koko-analytics`, with the aim of disabling analytics tracking by websites that run the [Koko Analytics plugin for WordPress](https://www.kokoanalytics.com/), of which I am the author.

While it could be argued that this plugin should not be blocked by EasyPrivacy as it does not track users in any way, it just counts visits and views to certain URL's, does not store anything visitor related (it wouldn't scale at all if it did), instead I would like to propose to change the filter so it only  blocks the tracking script.

My reason for this is that I have been getting a lot of questions from plugin users which are also running an ad-blocker, asking me why their dashboard suddenly broke. It's because EasyPrivacy is blocking the request to the JavaScript file which spins up the application showing their statistics in their website backend (WP Admin). 

I would really like to have my plugin working for users with an ad-blocker enabled without them having to whitelist their own domains because every user that is using Koko Analytics instead of Google Analytics or what have you is a win for the internet, imo. 

I'll add a PR which addresses this shortly hereafter.

Thank you for your consideration!

dannyvankooten:
You can view this issue in action but also what Koko Analytics is tracking / showing by visiting the demo dashboard for kokoanalytics.com here, by the way: https://www.kokoanalytics.com/?go-to-demo

(Disable your ad-blocker if you see a blank page, in case that was not clear. :sweat_smile:)

dannyvankooten:
Thank you for looking into this @ryanbr, but 81d0220498533535c360c5d0e16f7b17852800e6 doesn't really work because Koko Analytics is a WordPress plugin which people install on their own websites (domains). The dashboard on kokoanalytics.com is just an example, but it could live on any domain. 

The proper fix would be to only block the tracking script and not any file that lives in a subdirectory called `koko-analytics`.

arnelap:
If you want to block just the tracking script as Danny suggested, and allow the backend of Koko to keep working for those using Easyist, the correct script to block is: **/koko-analytics-collect.php**

Thanks!!


<rant>
Please note that Koko doesn't infringe on end user privacy in any way.

Koko doesn't track individual users in any way, all it does is count the total amount of visits to a Wordpress post or page. It does not keep track of IP addresses and doesn't assign any ID to individual users. 

[...]

PR #12974, opened 2022-08-17
 Change koko-analytics filter to only block tracking script 

dannyvankooten:
The current filter breaks the entire Koko Analytics plugin for website owners, because the JavaScript file for showing their analytics dashboard breaks too. With this change, only the tracking script is blocked while the showing of statistics keeps working for website owners also running an ad-blocker w/ EasyPrivacy.

Also see #12973


Yuki2718:
I'm not an EL maintainer but remove the last `*` as it's not needed and causes slower processing.

PR #12976, opened 2022-08-17
A: ek.fi (specific cookie hide)

Nojuuu:
https://ek.fi/ajankohtaista/uutiset/logistiikka-kuntoon-kaupungeissa-tata-yritykset-tarvitsevat/


peace2000:
Is it ok to block the script?

`||ek.fi/wp-content/plugins/webtoffee-gdpr-cookie-consent/public/js/cookie-law-info-public.js$script,domain=ek.fi`

Could also be mage generic:

`-gdpr-cookie-consent/`

Nojuuu:
I tried that and blocking that just moves it to the bottom of the page for me 

kuva

peace2000:
That's even better then seeing a visible banner at the mid page :D

Maybe that site needs a cosmetic then. But still `-cookie-consent/` would be a good generic imo.

PR #12997, opened 2022-08-19
A: https://www.thejournal.ie/

edivaniadc:
Image backgrounds ad on thejournal.ie
 
The links is blocked by EL rule: `||doubleclick.net^$third-party` but image background remains unblocked.

tjie


ryanbr:
||thejournal.ie/media/hpto/$image  should be enough @edivaniadc

edivaniadc:
> ||thejournal.ie/media/hpto/$image should be enough @edivaniadc

@ryanbr thanks for the suggestion. 
I left the `cdn.` to make it more specific and because they seem to be using the same link for a while; just changing the folder path and image format: https://github.com/easylist/easylist/commit/e86eb06 
but if you want I can update filter to only `thejournal.ie`

PR #12998, opened 2022-08-19
P: tui.* (Fix scrolling)

Nojuuu:
`https://www.tui.dk/`
`https://www.tui.no/`
`https://www.tui.se/`


peace2000:
Maybe change:
`5asec.pt,tui.dk,tui.fi,tui.no,tui.se##+js(rc, no-scroll, , stay)`
->
`5asec.pt,tui.dk,tui.fi,tui.no,tui.se##+js(rc, no-scroll, body, stay)`

Even if selector is optional, it should perform better if selector if specified:

https://github.com/uBlockOrigin/uBlock-issues/discussions/2195

PR #13034, opened 2022-08-22
Fix boxzilla filters

wavygecko:
Fixes boxzilla overlay selector.

e.g. `https://thebaffler.com/salvos/dystopia-for-realists-oshea`


krystian3w:
May need similar update like in AdGuard Annoyances: #15330

PR #13039, opened 2022-08-23
A: https://crashchance.com tracker

0o5rlubm:
According to the website itself, that has an explanation page: "This domain is used by digital publishers to control access to copyrighted content in accordance with the Digital Millennium Copyright Act and understand how visitors are accessing their copyrighted content."

Seen in the wild in https://www.presse-citron.net/


Khrin:
Definitely not a tracker.

PR #13040, opened 2022-08-23
A: https://lemonde.fr annoying section

0o5rlubm:
remove "service" section present on the bottom of articles, ie. https://www.lemonde.fr/politique/article/2022/08/22/darmanin-veut-un-encadrement-militaire-pour-les-mineurs-delinquants-dans-les-territoires-d-outre-mer_6138702_823448.html


0o5rlubm:
Hi  @Khrin , could you give me a hint about what is wrong with this filter?

PR #13048, opened 2022-08-23
A: https://doctolib.fr tracking

0o5rlubm:


0o5rlubm:
I would have also added ` ||api.doctolib.fr/nr_insert_* ` but it seems to block logins

PR #13070, opened 2022-08-24
A: fortune.com

Yuki2718:


Yuki2718:
@Khrin Apparently `fortune.com##.homepage`  is enough, pls change if you prefer that.

PR #13055, opened 2022-08-24
A: nordiskfilm.com (cookie hide for uBO)

Nojuuu:
https://www.nordiskfilm.com/news-0

Issue #12861, opened 2022-08-08
Character trimming suggestion


peace2000:
Since the release of Adblock Plus 3.13 https://blog.adblockplus.org/releases/adblock-plus-313-for-chrome-firefox-microsoft-edge-and-opera ABP has supported using `:has()` and `:has-text()` (Which are supported by uBO and AG as well) as alternatives to `:-abp-has()` and `:-abp-contains()`.

If those `-abp-`prefixed filters would be changed to shorter `:has()` and `:has-text()` that would decrease character count of filters. (If change is made, `?` must still be retained between `##`.)

ryanbr:
I'll wait on Eyeo first to confirm

peace2000:
Ok, though it's mentioned on that link:

![image](https://user-images.githubusercontent.com/17256841/183624089-cd842e0a-1d14-4016-a06c-5aa64332cb8b.png)

krystian3w:
This break compatibility with users stay on ABP lower 3.0 e.g. in Firefox PreQuantum.

peace2000:
Imo users of very old versions should update and not stay in very old versions.

Versions of ABP that are lower than 3, don't even support the usage of regular expressions in `:-abp-contains` filters: https://blog.adblockplus.org/releases/adblock-plus-303-for-chrome-firefox-and-opera-rolled-out-progressively

And I doubt that Easylist will withhold from using regexp with `:-abp-contains` (or with `:has-text`).

![image](https://user-images.githubusercontent.com/17256841/184298617-eeb3e72a-66d3-4d7b-985f-200bea73b5a0.png)

Those highlighted entries aren't supported under ABP 3.

But anyway, it's up to Easylist authors to decide how legacy versions they want to support.

krystian3w:
Header decided:

https://github.com/easylist/easylist/blob/7bc43f6140692850278e1c39a78920818507a8d2/easylist.template#L1

E.g. SlimBoat and Slimjet in own AdBlocker support up to 3.4, so list with ID 3.5 generate error download.

peace2000:
https://github.com/easylist/easylist/commit/33ca7f0

PR #13083, opened 2022-08-25
M: filter adjusting, performance reasons

peace2000:
See my comment here: https://github.com/easylist/easylist/pull/13055#issuecomment-1227212174

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -2148,7 +2148,7 @@ topannonces.fr##.cmpcontainer
 topannonces.fr##body:style(overflow: auto !important;)
 ||topannonces.fr/assets/js/cmp.min.js
 ! storyhouseegmon / egmont.com
-storyhouseegmont.dk,storyhouseegmont.no,storyhouseegmont.se,egmont.com,nordiskfilm.com##+js(rc, cookie-bar, div, stay)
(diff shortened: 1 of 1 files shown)


Yuki2718:
In the similar vein it is better to rewrite filters like `#?#div:-abp-has` if possible.

peace2000:
@Yuki2718  what do you mean?

Yuki2718:
> what do you mean?
`#?#div:-abp-has` => procedural filter js code evaluates all the div on the page, if the numbers are big it has the same issue as you pointed out.

peace2000:
Well, a quick way to literally get rid off `#?#div:-abp-has` entries could be achieved by doing this: https://github.com/easylist/easylist/issues/12861, lol :D

Yeah, got what you mean.

Are there many such entries in EL?

One easy (maybe) way could be to make the filter a bit more specific.

Like:

`div:-abp-has(.adcrap)`

->

`.article-container > div:-abp-has(.adcrap)`

Then it would only check divs under article-container and ignore all the other divs in the page.

But, that would probably require a separate issue :P

Also, doesn't native `:has` perform better anyway? It's about to land soon.

peace2000:
Well, made one PR: https://github.com/easylist/easylist/pull/13087

This is not small project to check all of those.. :D

Yuki2718:
> Are there many such entries in EL?

No, only 3 including the one you opened PR.

> Also, doesn't native :has perform better anyway? It's about to land soon.

Only on Chrome? This is a bit of my concern as probably I have to rewrite some :upward with :has, once uBO starts to take this as native selector.

peace2000:
Firefox soon as well: https://caniuse.com/css-has :) they have experimental support now.

PR #13096, opened 2022-08-26
Add separator character mark and content type to JS filter rules

salim-b:
Currently, the rule `-twitter.js` is way too broad and breaks this site: https://www.republik.ch/2022/08/26/ukrainer-finden-arbeit-in-der-schweiz-neue-studie-zur-gletscherschmelze-und-schwere-vorwuerfe-gegen-twitter

While at it, I just narrowed down all similar rules ending in `.js` to avoid similar issues like the above one in the future.


Khrin:
Better to just remove the filter. Also, not sure if all filters were meant to block only scripts.

PR #13111, opened 2022-08-27
P: tweakers.net (can't open GDPR if needed)

### LINKED ISSUE #12285: tweakers.net - cookiebar blocking prevents users from enabling 3rd-party embeds
`https://tweakers.net/nieuws/197852/portal-achtige-puzzelgame-the-entropy-centre-komt-in-2022-uit.html`

### What happens?
When I click on 'Open cookievoorkeuren' nothing happens:
image

What **should** happen is that this should open the cookie-consent bar which will enable the user to give consent for embedding e.g. YouTube:
image

EasyList
Easylist Cookie List

### Your settings
- OS/version: Windows 7/10/11
- Browser/version: Firefox 101, Chrome 102
- Adblock Extension/version: AdblockPlus 3.14

### Other details:
The culprit is the recent addition of the following rule to the Easylist Cookie List in https://github.com/easylist/easylist/issues/12281:
`tweakers.net##.koekie_bar`

Now obviously I am a representative of the webite in question. I am mainly reporting this issue to raise awareness to the negative consequences that blocking cookie-consent bars have for users. We understand that these cookie-bars (and walls) can be an annoyance to users, but not only are they mandatory for websites because of EU-regulations (GDPR), they also give users the possibility to decide for themselves what they deem acceptable in terms of the trade-off between functionality (e.g. watching YouTube embeds) and privacy.

We are very proud to be able to say that our website is by default completely free of any third-party tracking, and our cookie-consent bar is - in our view - a prime example of an unbiased selection tool. We do hope that you understand that in this case you are taking away this choice from users without giving them any alternative option. This also often results in false bugreports adressed to us.

peace2000:
I think that making this small change fixes #12285 . The banner will be hid but it's now possible to click it open in case the user wants.

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_hide.txt
+++ b/easylist_cookie/easylist_cookie_specific_hide.txt
@@ -1956,7 +1956,7 @@ msn.com##.cookieBanner
 ajinomoto.co.jp,meteoblue.com,wsj.com##.gdpr-message
 github.com##.hx_cookie-banner
 newsnow.co.uk##.js-cookie-msg
-tweakers.net##.koekie_bar
+tweakers.net##.koekie_bar:not(.tall)
 ivi.tv##.lowest-teaser
 mgtv.com##.m-agreement
 symbolab.com##.nl-cookiepolicy


krystian3w:
https://github.com/easylist/easylist/issues/14762

krystian3w:
Converted to `set-cookie`?

Page shouldn't unlock load 3-rd party content until GDPR will be accepted.

peace2000:
@krystian3w  https://github.com/easylist/easylist/pull/16469

PR #13120, opened 2022-08-27
Trackers update

0o5rlubm:
some bot protection stuff?


ryanbr:
Removed filter, not worth fighting. they kept changing too often

PR #13121, opened 2022-08-27
A: https://quickbooks.intuit.com/ tracking

0o5rlubm:


ryanbr:
`liveperson` is used for Chat @0o5rlubm

0o5rlubm:
> `liveperson` is used for Chat @0o5rlubm

i was able to go to the last step of the calendar booking flow, that is part of chatbot, with this tracking domain in the blacklist. This domain does not look to be a blocker.
i should have mentioned this sorry

ryanbr:
Not worth the potential FP's.

PR #13137, opened 2022-08-29
A: https://www.marketwatch.com/

edivaniadc:
Hide **Dianomi** sticky ad on https://www.marketwatch.com/
**Note: ad visible on ABP with ACCEPTABLE ADS list enabled**

m94

m95


ryanbr:
Shouldn't this be fixed in Acceptable Ads list? @edivaniadc

edivaniadc:
> Shouldn't this be fixed in Acceptable Ads list? @edivaniadc
@ryanbr maybe, but would it be possible to fix that in EL directly please? so users wouldn't get this ad at all.

ryanbr:
Lets fix this in Acceptable Ads list,  since the dependency for the issue solely depends on this list to be enabled. If the issue showed in standard EL, then it would be reviewed

PR #13163, opened 2022-09-01
A: https://www.doordash.com/search/store/burger/?event_type=search

edivaniadc:
Hide ads on https://www.doordash.com/search/store/burger/?event_type=search
VPN set to US or CA

_Note: I've tried also with ` data-anchor-id="StoreCard"` but it didn't work._

12d


ryanbr:
Reversed, breaks add to cart.

PR #13191, opened 2022-09-04
A: https://www.snbforums.com/

smed79:

PR #13199, opened 2022-09-05
M: (snbforums.com) Revert #13191 in favor to 12ec8ee

smed79:


ryanbr:
was removed, no need to revert.  other filter considered 'bad'

PR #13245, opened 2022-09-08
Update easylist_general_hide.txt

AirSayy:
Found some element filters on www.gsmarena.com.
Screenshot (32)


ryanbr:
Check `https://github.com/easylist/easylist/blob/master/easylist/easylist_general_hide.txt` they are all covered

PR #13278, opened 2022-09-09
A: https://api.met.no/

smed79:


ryanbr:
Too generic, imo we should trim some of these. could result in breakages.

PR #13352, opened 2022-09-14
A: (Fix) https://www.manager-magazin.de/

felippesantos46:
Fix newsletter being block by EL Privacy at https://www.manager-magazin.de/

Screenshot 2022-09-14 at 18 10 34


ryanbr:
Bad for privacy @felippesantos46

PR #13359, opened 2022-09-15
A: https://www.instagram.com/

felippesantos46:
Hide Paid sponsorship content.

To reproduce:
1 - Enter at the Instagram profile (need to be following some pages - Champions league official page usually releases a lot of ads)  
2 - Scroll and have CTRL + F ( Paid Partnership )  
3 - you will find something like that: 
instapaidpartnership
paidpartnership

4- Two of those filters are necessary, otherwise requests will be made and the paid partnerships will be appearing and disappearing.


sugarfreesweetener:
https://github.com/uBlockOrigin/uAssets/issues/14912

@Khrin

vistageek:
@Khrin I don't think this request should have been added. This is removing/blanking valid posts that are not necessarily "advertisements" (even the screenshot above shows a non-advertising post, just information about a football match and scores with small "Presented by Spotify" text at the top).

Easylist already does a good job at blocking actual advert posts on Instagram, but this is over the top blocking non-advertisement posts from people I actually follow on Instagram.

PR #13362, opened 2022-09-15
A: mangas-raw.com

Yuki2718:


Khrin:
Isn't this a bit too mouch generic? Not a risk of false positives?

Yuki2718:
> Isn't this a bit too mouch generic? Not a risk of false positives?

I agree, but this is the simplest rule. Alternatives require at least two rules e.g. `##.entry-content > center` + `##.kln`

Yuki2718:
Updated

PR #13454, opened 2022-09-21
Remove wikipedia banners

wrt54g:
Removes the funding and sound of human knowledge banners of the Wikipedia site.


kotx:
Why was this closed?

PR #13471, opened 2022-09-22
M: Update

felippesantos46:


Yuki2718:
EL has `||predictiondexchange.com^` so `||predictiondexchange.com^$third-party` is redundant.

PR #13488, opened 2022-09-24
A: zendesk tracking

0o5rlubm:
seen at https://www.okpal.com/, then click on bottom right bubble


ryanbr:
Could this break the chat app?

0o5rlubm:
> Could this break the chat app?

i've tested and it doesn't appear to break the usage of the chat system.

ryanbr:
I'm not going to take the chance.

PR #13509, opened 2022-09-26
A: http://www.freeglobes.net/

smed79:


ryanbr:
Safe to move to easyprivacy trackerservers list itself @smed79

PR #13571, opened 2022-09-30
A: satshop.fi + others (GDPR script)

peace2000:
https://www.satshop.fi/

Amasty GDPR script is used by some sites: https://publicwww.com/websites/%22Amasty_GdprFrontendUi%22/

But better make it specific instead of generic. I once added it as a generic https://github.com/easylist/easylist/pull/7051 but it was later removed (can't find it on ELC anymore), maybe it caused breakages as a generic filter.


ryanbr:
A generic for one domain? Is there more cases we cab add to this?

peace2000:
Added more domains but also noticed I had to make changes as those sites store normal cookie related functions under `/Amasty_GdprFrontendUi/` so can't block everything as it causes breakages. Changed this so that it only blocks the modal dialog. I'll review this more later but this should be safe.

PR #13608, opened 2022-10-04
A: Sponsored Ad carousel on bestbuy.com

sabinahschmitz:
According to the website: "These are ads for products on BestBuy.com. These brands paid for an extra chance to get your attention."

Screenshot 2022-10-04 at 09 34 06


Yuki2718:
No need to use `#?#` for the plain selector.

PR #13625, opened 2022-10-05
A: https://www.luukku.com/ (cookie hide)

Nojuuu:


ryanbr:
Just add the ##.dialog-lightbox-message filters in specific hide @Nojuuu

Nojuuu:
Is it always cookie dialog?

PR #13628, opened 2022-10-05
A: https://www.0564.ua/news

smed79:


Khrin:
Might be too generic, added as specific filter.

PR #13653, opened 2022-10-06
Fix signup flow on posthog.slack.com

xvello:
Hello folks!

[This rule](https://github.com/easylist/easylist/blob/b7e4f1f53deb12808cf086d35f3624755d2c6bad/easyprivacy/easyprivacy_general.txt#L5387) is too greedy and breaks the XHR requests on their Slack instance.

image

While I have experience with cosmetic filters, I don't have much knowledge of the network filter list. My intent here is to whitelist requests to posthog.slack.com from posthog.slack.com, is that what will happen?

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -5384,7 +5384,7 @@
 ://metrika.*/analytics.
 ://mint.*/?js
 ://piwik.$domain=~matomo.org|~piwik.pro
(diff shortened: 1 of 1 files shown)


xvello:
Thanks @ryanbr for removing the rule altogether.

I'm assuming the intent was to block the self-hosted posthog instances, as the software can be self-hosted like mamoto, but it's easy to work around anyway.

PR #13659, opened 2022-10-07
P: [Breakage] https://www.cook.com/

felippesantos46:


ryanbr:
Its a parked, fake site. I expect this to break tbhj, helps no one but advertisers using such a service.

PR #13768, opened 2022-10-15
A: https://app.aclipp.com

0o5rlubm:
also on https://userguiding.com/blog/top-startups/

i was able to create an account and navigate in the app with domain in the blocklist


0o5rlubm:
also as it appears to be a tracking company, perhaps the whole topdomain can be added, as you prefer

PR #13802, opened 2022-10-17
M: Adjust dasboard.unity.com allow list

aitk:
Creates generic fix for dashboard.unity.com for loading assets.

A recent change to the CDN path broke the existing allow list entry.


ryanbr:
This patch is too over reaching.

PR #13803, opened 2022-10-17
A: https://cad.onshape.com/

smed79:


easyprivacy/easyprivacy_thirdparty.txt
```diff
@@ -1235,6 +1235,7 @@
 ||metrics.ctvdigital.net^
 ||metrics.kmsmep.com^
 ||metrics.mdstrm.com^
+||metrics.onshape.com^
```
spirillen, line 1238: You should add `tracking.go.onshape.com`  as well (https://mypdns.org/my-privacy-dns/matrix/-/issues/4122652#hosts-and-pi-hole)

PR #13811, opened 2022-10-18
A: (nsfw) https://it.pornhub.com/view_video.php?viewkey=ph5fe75e2dbf5…

Yuki2718:
…c5 (fixes https://github.com/AdguardTeam/AdguardFilters/issues/132379)
https://github.com/AdguardTeam/AdguardFilters/issues/132379


Khrin:

Yuki2718:
Hmm, okay.

PR #13813, opened 2022-10-18
A: Indeed emails tracking

0o5rlubm:
tracking pixel in emails
Proof: https://pastebin.com/WQdpCH8A

Does it fit in this list? If not, do you have any hint for appropriate repo? Thanks


ryanbr:
Test in gmail is a better fit.

PR #13839, opened 2022-10-19
A: https://www.vistaprint.com

0o5rlubm:
tested that it doesn't break photo album editor


0o5rlubm:
@spirillen among the different separated domains that Vistaprint uses on its pages, there is this one used only for analytics requests with different subdomains (ie. https://api.dcp.vpdcp.com/v1/t), blocking the whole domain doesn't seam to be too risky here (like domain bought by marketing team for its own usage) and site functionalities were working well for me with this blocking rule.

PR #13858, opened 2022-10-21
A: google.com (activity paused nag)

peace2000:
When I set a setting on my Google account not to save my web activity, they have started to show this nag all the time:

image

When I click "got it" and refresh the page, the nag just appears again.

"Learn more" link leads to this page: https://myactivity.google.com/activitycontrols/webandapp?pli=1

So, Google wants me to set my account to save my web activities to a log, otherwise they keep spamming me with that nag :)...

(On my PR I changed `.com` -> `*` because this nag uses different TLD's depending what language version of Google is being used.)


krystian3w:
Doesn't  hide the pop-up "It's been a while since you logged in - do you want to check if your account configuration is still OK"?

PR #13861, opened 2022-10-22
A: (nsfw) manga9.co​

Yuki2718:
Just another adspyglass adserver happened to be 1p. So I think better adservers.txt than specific.txt.


Yuki2718:
Also added narutoon to avoid conflict. All the same.

Khrin:
Better in the specific list anyway

Yuki2718:
All right. BTW manga9 is genarally not NSFW, the title was mistake.

PR #13910, opened 2022-10-25
A: https://www.linforme.com/abonnements

0o5rlubm:
enter a dummy address on https://www.linforme.com/abonnements then you'll have the  stripe form that sends tracking to both domains.

r.stripe.com receives bunch of tracking data in cleartext, whereas m.stripe.com receives encoded tracking stuff. Payment-related data are sent to other endpoints.

I've tested by clocking both domains and was able to successfully make a payment.


ryanbr:
Okay, will remove if we see any issues/false positives.

ryanbr:
@0o5rlubm was changed to $script due too https://github.com/AdguardTeam/AdGuardSDNSFilter/issues/1433 is this still useful? any other sites using r.stripe.com to test?

PR #13924, opened 2022-10-26
Add Jscrambler email tracking pixel host

migueldemoura:
This classic tracking pixel is added to emails, example:
```
<div class="tracking"><img alt="" height="0" src="
https://opens.jscrambler.com/home/index/?token=XXX" width="0"></div>
```

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -5671,6 +5671,7 @@ cgi-bin/counter
 /notifications/beacon/*
 /open.aspx?
 /open.html?$image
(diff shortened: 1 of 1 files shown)


ryanbr:
Which newsletter are these seen on? Also using something like `/opens.jscrambler.com/*$image`  which picks up trackers on webmail like gmail

migueldemoura:
> Which newsletter are these seen on? Also using something like `/opens.jscrambler.com/*$image` which picks up trackers on webmail like gmail

These appear in transactional emails. Swapped to that filter instead, thanks!

PR #13966, opened 2022-10-31
M: no longer needed

Yuki2718:


ryanbr:
Removed both, the `https://www.kireie.com/js/adclick.js` isn't much of an issue

PR #13976, opened 2022-10-31
A: https://www.rfi.fr/fr/

0o5rlubm:
and remove blur background


Khrin:
Not needed?

0o5rlubm:
@Khrin popup is shown on firefox, but not on chrome... tested with empty profiles for both with just ublock installed with default params

krystian3w:
Default doesn't have enabled EasyList Cookies and Fanboy Annoyances.

Then test proves nothing.

PR #14080, opened 2022-11-10
A: https://futemax.app

eporubin:
Then the page (www.futemax.app/) loads first click on the page opens a new tab with ads requested from keptic.com


ryanbr:
ALready added.

PR #14117, opened 2022-11-14
D: https://www.takeaway.com/

jet-georgi-velev:
Hello,

I'm adding this PR on behalf of our Dev department before we have to escalate this.

Two months ago, these were wrongly targeted as ads on our platform while they are not actually advertisements. 
To plead the case - unlike the Google Ads that appear on the top of the Google search that add additional paid search results, the entries marked as `Sponsored` on our platform are already part of the results for the area. Without going into unnecessary details of our internal process, those filtered results, regardless of being labeled or not, will always be there for a certain search as they are physical locations servicing geographical areas and must be present. The are `Sponsored` label doesn't indicate an additional advertisement content like on sites like Instagram, Reddit or Facebook.

By removing them with this filter, the search results for the location that the user is trying to order are visually reduced by 30% to 90%, depending on the location and time, usually leaving them no options and even hiding the places they usually are clients of, thus not actually helping the user get rid of useless content on the page but stripping them from options to actually use the website.

Given this statement, I'm kindly asking you to merge this PR. 

Best regards,
G. Velev


jet-georgi-velev:
One additional note - I tried to find a thread on the forum or a PR discussion regarding the inclusion of those filters and the motivation behind it but I couldn't. 

It would be very helpful to share that information with us, if such is present.

Khrin:
> One additional note - I tried to find a thread on the forum or a PR discussion regarding the inclusion of those filters and the motivation behind it but I couldn't.
> 
> It would be very helpful to share that information with us, if such is present.

https://github.com/easylist/easylist/commit/30b43a1

PR #14135, opened 2022-11-16
M: https://fydeos.io/

Nojuuu:


ryanbr:
Would `fydeos.com,fydeos.io##.z-90000` work? Just to minimise it

Nojuuu:
Seemed to be fine at least for now. 

Also noticed that some paths like `https://fydeos.com/pricing/` still need `##.cookieBox___2OASs`, so I added it back.

PR #14136, opened 2022-11-16
A: Update revolving adservers

smed79:
- https://katb.in/vinavureyel/raw


ryanbr:
Where did these domains come from? seems to be quite a big list.

ryanbr:
```
fanboy@DESKTOP-L731CV9:~$ cat domains.txt | grep INACTIVE | wc -l
499
```

499 Dead domains ^

smed79:
> where did these domains come from?

Most come from known [adnetworks](https://github.com/Ultimate-Hosts-Blacklist/blacklist/blob/master/ip.list) using revolving domains.

> 499 Dead domains

Will be checked again.

PR #14168, opened 2022-11-19
Use common header text for all lists

mgmax:
Before, many lists had different license information and forum/email links.
Now, all this is included from a single template file.

### DIFF
--- a/easylist-cookie-uBO.template
+++ b/easylist-cookie-uBO.template
@@ -2,11 +2,7 @@
 ! Title: Easylist Cookie List (uBO)
 ! Last modified: %timestamp%
 ! Expires: 4 days (update frequency)
-! License: http://creativecommons.org/licenses/by/3.0/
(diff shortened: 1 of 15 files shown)


ryanbr:
Not sure if this is needed?

mgmax:
Two examples:
- For legal reasons, the license info needs to be correct and included with each file.
- The header is the first thing that people see when clicking on such a list e.g. in UBlock. So it is important that the information in this header is correct. For example, for EasyList cookies the header did not contain the Github link and I spent many minutes trying to register to that forum, which still did not work in the end even after disabling all browser extensions.

The alternative would be to copy and paste back and forth between 7 files, so I guess a template is better than that...

Nojuuu:
This is currently not working properly in EasyPrivacy with uBO

image

monzta:
@Nojuuu thanks! this should help https://github.com/easylist/easylist/commit/6114377

PR #14199, opened 2022-11-22
P: https://atcoder.jp/posts/780

Yuki2718:
Links to their books:

atcoder1

atcoder2


ryanbr:
These are amazon ads, should it be allowed. I'm hesitant @Yuki2718

Yuki2718:
> These are amazon ads, should it be allowed. I'm hesitant 

They're their official books and not random ones only to earn ad revenue. Without whitelisting, all the visitor see is strange broken page, which results in whitelisting the whole page, exposing them to more ads and tracker as in `https://twitter.com/gojira_kyopro/status/1594888912630972416`.

Khrin:
@Yuki2718 Why is required such a filter for `assoc-amazon.com`? Why not using `@@||assoc-amazon.com/widgets/cm?$subdocument,domain=atcoder.jp`?

Yuki2718:
@Khrin I will change.

Khrin:
Thanks.

PR #13916, opened 2022-10-26
Remove blocked generic utility classes for Zalando

axyz:
Those selectors are not related to sponsored content. Zalando uses utility classes (e.g. `lRlpGv` is used to set padding). Using those generic selectors will hide a lot of elements that may or may not be sponsored content resulting in a broken experience.

We could eventually maintain the `(span:-abp-contains(/Gesponsert|Gesponsord|Patrocinado|Sponset|Sponsitud|Sponsored|Sponsoreret|Sponsorēts|Sponsorizzato|Sponsorisé|Sponsoroitu|Sponsorowane|Sponsrad|Sponzorirano|Sponzorováno/))`, but not sure how to rewrite it without the hardcoded utility class `Uxq3DH`?

abp-contains may also result in an unwanted behavior here: a sponsored content may have a possibly parent, sibling, or child element that contains the word "sponsored", hiding that element will hide the sponsored flag, but preserve the sponsored content ultimately still showing the sponsored content, but with no way for the user to know that it is in fact sponsored. How are those use cases usually handled?

PR #14217, opened 2022-11-23
remove zalando filters

axyz:
Hi, as a follow up to #13916, I would like to better understand if it would be possible to remove all Zalando's content based filters.

As mentioned in the [readme](https://github.com/easylist/easylist/blob/master/README.md), "first-party" ads should not be targeted. The definition of "first-party" ads, refers to those that do not link or pass-through to any 3rd party site. To my knowledge all the sponsored content present on Zalando website always links to other Zalando pages, without going through external domains.

I want to double check with you if this is correct and if we can proceed removing the existing rules.

Additionally I would like to facilitate and encourage a more active collaboration between Zalando and this project to make sure that both the user preference and the user experience are respected for current and future possible use cases.

In case this won't be possible, I'd like to better understand the guidelines applied and what actions could Zalando eventually take to comply with them.


krystian3w:
Also exist rule "once blocked incorrectly self-promo will be never unlocked".

PR #14235, opened 2022-11-25
A: https://www.wetter.com/

eporubin:
Notification missing on https://www.wetter.com/ when ABP is active caused by EasyPrivacy blocking pushwoosh.com^$third-party


ryanbr:
This is expected @eporubin A notification service that tracks, still tracks. Privacy > notifications.

PR #14341, opened 2022-12-07
A: Update revolving adservers

smed79:


Khrin:
`||sh|ptewauta.net^` needs to be fixed.

ghajini:
also `||or||greemeek.net^`

smed79:
Fixed.

PR #14342, opened 2022-12-07
Use generic filter rule to match `###iqadtile*`

salim-b:
Instead of adding yet another rule for `###iqadtile5` (encountered [here](https://www.golem.de/)), use generic cosmetic filter rule matching any `id` attribute beginning with `iqadtile`.


Khrin:
It's better to stay more specific as is it now, for the time being.

PR #14357, opened 2022-12-09
D: https://www.insider.com/

eporubin:
Please remove this specific filter as it does block the coupons section which seem to be one of the website features. 
image

Please note that the adblock wall on the page seems to have bugs as if you allowlist the website then block it back and refresh it does let you scroll for some time then detects it again.  Kind regards.


ryanbr:
They're coupon ads. @eporubin

PR #14402, opened 2022-12-14
A: Update revolving adservers

smed79:


Khrin:
@smed79 Needs to be rebased.

smed79:
@Khrin fixed.

PR #14403, opened 2022-12-14
A: https://www.similarweb.com/fr/website/easylist.to

smed79:


Khrin:
@smed79 Probably better as specific filter.

PR #14420, opened 2022-12-16
M: video content blocked on autobild.de

eporubin:
Please allowlist this domain as its blocking video content owned by the website 
Thank you 
image


ryanbr:
I'm seeing ads in one screenshot, and no ads on the other. Is this a good fix I have doubts. @eporubin

monzta:
@ryanbr My guess is that in the right screenshot no ad blocker is enabled.

PR #14443, opened 2022-12-19
M: greyhound-data.com

eporubin:
Please add this filter to as the one I have added first and removed now wasn’t working on all pages.  Thanks a lot in advance.

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -1092,7 +1092,6 @@ sciencealert.com##.ad-desktop\:block
 mobilesyrup.com##.ad-goes-here
 dailyuptea.com##.ad-info
 goal.com##.ad-slide_adSlideBanner__cEgCm
-greyhound-data.com##.ad-three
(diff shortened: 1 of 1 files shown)


ryanbr:
Which page is causing issues?

eporubin:
Hi is the same page [greyhound-data.com](https://greyhound-data.com/) is just that on some of the subpages (for example this one `https://greyhound-data.com/knowledge.php` )the initial filter which I submitted  did not work thats why I changed it and now to be applicable on all pages 
image
Thanks a lot for looking into it

eporubin:
Hi @ryanbr  thanks a lot for looking into it, yet could you please elaborate towards your decision  of adding additionally 3 other filters for the same issues than the `greyhound-data.com##.ppts` which I proposed would basically do the same thing. and its at the top of the chain rather than all 3 filters applied which also impact same `<ul class="ppts">` (and the list its composed of only images and played one by one so that it looks like its a gif). I mean the solution applied also work I just want to ensure we can save on some space in the list. 
Thank you in advance.

Issue #9642, opened 2021-11-08
Issues with /tarteaucitron. filter on www.velib-metropole.fr and teleservices.paris.fr


Noob729:
### List the website(s) you're having issues:

`https://www.velib-metropole.fr/en/map`
`https://teleservices.paris.fr/ffaxssl/index.html`

### What happens?

The two pages fail to load when the `/tarteaucitron.` filter is enabled.
Adding an exception in UBO
`@@||www.velib-metropole.fr/tarteaucitron/tarteaucitron.js`
allows the page to load.

Searching for this specific script turns up:
https://tarteaucitron.io/en/

### List Subscriptions you're using:

According to the UBO logger, this particular filter is included in Fanboy's Annoyance List
```
! Checksum: VJzhWgMyUB6vRdE0uVfD0g
! Version: 202110310900
! Title: Fanboy's Annoyance List
! Last modified: 2021-10-31 09:00 UTC
! Expires: 9 days (update frequency)
! Homepage: https://easylist.to/
! License: http://creativecommons.org/licenses/by/3.0/
!
! Please report any unblocked content or problems by email or in our forums
! Email: easylist@protonmail.com
! Forums: https://forums.lanik.us/
```
### Your settings

- OS/version: Linux
- Browser/version: Firefox 94
- Adblock Extension/version: UBO 1.38.6

Noob729:
https://github.com/AmauriC/tarteaucitron.js

Noob729:
Hello @ryanbr 
Thanks for the quick fix.
I was not aware there had been an explicit exception for ameli.fr
I see other exceptions in the list:
```
@@||defense.gouv.fr/lib/tarteaucitron/$~third-party
@@||demarches.interieur.gouv.fr/js/tarteaucitron/
@@||gouvernement.fr^*/tarteaucitron/$~third-party
```
Not sure they match `/tarteaucitron.` though...
What rule are they over-ruling?
Should they be removed?

Noob729:
If I understand correctly, the dot in `/tarteaucitron.` is an explicit dot, not a regex meta-character?

As far as I can tell, many French websites now use the "tarteaucitron" (means lemon pie) script, but only some are broken when the script is blocked.

For the 3 exceptions listed above:
https://www.defense.gouv.fr/bundles/dicoddesign/js/tarteaucitron/tarteaucitron.js
https://www.demarches.interieur.gouv.fr/js/tarteaucitron/tarteaucitron.js
[...]

PR #14458, opened 2022-12-21
A: https://en.geneanet.org/

smed79:
```
https://atrium-paca.fr/
https://www.diplomatie.gouv.fr/en/
https://www.campusfrance.org/
https://en.geneanet.org/
```


smed79:
https://github.com/easylist/easylist/issues/9642

PR #14505, opened 2022-12-27
A: https://www.theverge.com/23496652/paramount-plus-best-shows-2022-n…

0o5rlubm:
…fl-mtv-roundup


ryanbr:
Are they really trackers?

0o5rlubm:
i think so. it sends all tracking data (referrer...) then get backs 3 int values. They be used somewhere, but I don't see any visual change. Finally the [company website](https://concert.io/) doesn't seams to show any useful usage of the product. Merge or close as you want :)

ryanbr:
concert is more of a paywall style company. we already block the lookup tracker.

PR #14616, opened 2023-01-07
A: https://www.france24.com

0o5rlubm:


ryanbr:
This looks messy, why the need for the extra line?

0o5rlubm:
it is necessary to remove the blurring effect, in addition to the modal. It's not possible to use a more strict rule, to still catch only the needed element.
Otherwise we can have a wildcard on all elements, but it may break a style somewhere.. Feel free to edit if you want

PR #14635, opened 2023-01-09
P: halvinsähkösopimus.fi (price watch broken)

peace2000:
https://halvinsähkösopimus.fi/hintavahti/

Technically, this is a newsletter but it's a price watch tool. It's on a separate, dedicated page, thus it should not be blocked. You must willingly click yourself to that page.


ryanbr:
Convert to `xn--halvinshksopimus-1nb04a.fi`

PR #14652, opened 2023-01-10
block /api/log? only for original discovered domain

ds82:
/api/log? was orginally blocked in commit 12fa204, because of a request to hatena.ne.jp/api/log? on the website https://shikabane.hatenadiary.com/entry/2017/03/21/005452

The block is pretty generic even though only this specific website does somehting sketchy in this request.

On the other hand, it can easyly break valid applications using that route in their REST API.


ryanbr:
https://publicwww.com/websites/%22%2Fapi%2Flog%3F%22/  is used elsewhere.

Issue #14672, opened 2023-01-12
Ads (Easylist)


bestplayerbot:
https://github.com/yokoffing/filterlists/issues/49
Ads in these domains
luotphim.fun
filmapik21.fun
98e.fun
time.is
![image](https://user-images.githubusercontent.com/96731476/212052670-5979fe84-ac03-43d3-a227-0191d03596f6.png)

Adblock detected popup
egybist.fun

Khrin:
All sites but one don't even load for me.

PR #14675, opened 2023-01-12
A: https://time.is/

smed79:
https://github.com/easylist/easylist/issues/14672


Khrin:
Looks like self-promotion and not advertising, therefore shouldn't be blocked here.

PR #14683, opened 2023-01-13
A: `itmedia.co.jp`

eporubin:
Please add this filter to hide the ads displayed on the page
image
Thank you


ryanbr:
Make it generic, its non-engish so making it specific isn't good here.

eporubin:
> Make it generic, its non-engish so making it specific isn't good here.

Hi, thanks for looking into it, I have made it generic, sorry we dont have a JP list so i wasnt sure.

PR #14701, opened 2023-01-16
A: https://repack-games.com/

felippesantos46:
Block redirect popups at: https://repack-games.com/
Screenshot 2023-01-16 at 11 50 59


ryanbr:
Will cause false positives @felippesantos46

felippesantos46:
@ryanbr no worries, seems to be working just fine. 
Sorry for the late reply. Closing it.

PR #14711, opened 2023-01-17
A: `soap2day.cc`

eporubin:
Please add the domain to the existing filter it has popping ads on click in another tab. 
Thank you


ryanbr:
Can't simply block the source domain, it'll cause false positives.  @eporubin

eporubin:
Hi, @ryanbr  thank you for looking into it, please take in consideration that this domain `soap2day` has been reported by user several times already and the other domains endings such as `soap2day.ac` and `soap2day.sh` are already in the list and therefore I added the `soap2day.cc` on which our latest complain is. I understand your concern towards the actual domain/adverser but this domains change frequently their ads sources.

eporubin:
Hi, Could I please get an explanation why all the soap2day domains were just removed? we get constant complains from users about this website and from what we experienced over the times its always popping in another tab with ads. And as I was saying tracking down the source domain wont help much they change them frequently and thats min 4 domain. If you came across false positives please exemplify. Thank you 🙏

ryanbr:
Point, removed related filters.

Issue #14723, opened 2023-01-19
Including p.typekit.net in filter list is creating errors on applications


RomanFausek:
Hello!

Can I ask you why are you blacklisting `p.typekit.net`? I saw it being added in this commit https://github.com/easylist/easylist/commit/22a3441fe31e5a5be6d1307026961afe0b2a2b7c

The file from `p.typekit.net` with just the comment inside gets thrown away with assets (fonts) from the network which leads to font fallback in our application. The issue is that for some security policies, once it gets ignored, the rest of the request is thrown away.

In the following screens, you can see that downloading the file from `p.typekit.net` allows the network to continue downloading assets. But the screen with the ad block throws away the rest of the promise (which are the actual font assets).
![Screenshot 2023-01-19 at 10 53 03](https://user-images.githubusercontent.com/8074997/213418534-277a1c5e-8806-401d-a2ee-3951ad8c1836.png)
![Screenshot 2023-01-19 at 10 50 56](https://user-images.githubusercontent.com/8074997/213418558-02a419c9-43df-43a6-b863-8dfdce847a2e.png)

It is apparently not just us because I found out that at uBlock they are already adding whitelisted URLs and this whitelisting approach doesn't make sense to me at all https://github.com/uBlockOrigin/uAssets/blob/master/filters/unbreak.txt#L4865. I understand that you are afraid of some trackers but it might create problems for many websites using ad blocks with your filter list.

Some more details + conversation is here https://github.com/uBlockOrigin/uAssets/issues/16391

garry-ut99:
> RomanFausek : but it might create problems for many websites using ad blocks with your filter list.

https://publicwww.com/websites/%22p.typekit.net%22/

ryanbr:
See details on https://github.com/easylist/easylist/pull/14679

RomanFausek:
@ryanbr This needs a resolution (possibly a revert of #14679) and not just a link to the PR. So I don't agree that it is resolved.

Typekit is a commonly used service and including it in your list makes it unusable for many websites. Many of our customers are already reporting it. The only solution right now is not to use ad block extensions.

garry-ut99:
> stephenhawk8054 : https://github.com/uBlockOrigin/uAssets/issues/16391#issuecomment-1396728719 : `p.typekit.net` I'm not sure about its privacy history. Looks like it's related to Adobe's trackers?

[...]

PR #14759, opened 2023-01-23
M: Remove p.typekit.net as it is breaking websites

AdrienPoupa:
See this issue: https://github.com/easylist/easylist/issues/14723

Fonts can be loaded from `p.typekit.net`, so adding it to this list makes little sense and breaks websites. 

Bungie.net was affected and fixed with an allow list, but this is also breaking robertsspaceindustries.com

Rather than adding each site to the allowlist, I would rather remove it from the list, given the justification given in the [PR](https://github.com/easylist/easylist/commit/22a3441fe31e5a5be6d1307026961afe0b2a2b7c) are wrong: resources are not just in use.typekit.net.

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -1374,7 +1374,6 @@
 ||outbrain.com^*/widgetStatistics.js
 ||p.delivery.net^$third-party
 ||p.placed.com^
-||p.typekit.net^
 ||p.yotpo.com^
 ||p0.com/1x1
 ||paddle.com^*/analytics.js


0o5rlubm:
Hello, I'm the one who submitted this block because everything seems to show that this site is a tracker. Indeed, I have not blocked any typekit, but only the p-domain, which serves only empty files, while the needed resources on the other hand are not blocked. 

I would like to apologize for any inconvenience this may have caused.

However, it seems to me that this rule did not impact the millions of sites that use the tool, but only a small handful. In balance between generalized tracking and a few broken sites, maybe we can whitelist the broken sites while keeping the global block, [like what adguard does](https://github.com/AdguardTeam/AdguardFilters/commit/3f18656243889c03a67fb289f5d5e701e5b13d48)?

0o5rlubm:
> Fonts can be loaded from p.typekit.net

@AdrienPoupa could you give an example link? As far i've seen this domain only sends empty content

ryanbr:
With respect to what exceptions in Easyprivacy, breakage is pretty minimal. Looking at other filters some exceptions range from 1-20. Also the privacy implications, this is an adobe assett and they do not have a good track record for user privacy. 

We'll always review if a bad filter is actually bad, but as it stands this is fine. Breakage isn't really high compared to other Easylist/Easyprivacy exception filters.

AdrienPoupa:
I understand. Would you accept an exception similar to bungie's? https://github.com/easylist/easylist/commit/746465a31b39ef75a46859f6eab6cf11b55070ce

krystian3w:
AdGuard disabled block and allow methods:

https://github.com/AdguardTeam/AdguardFilters/commit/1ed6c7b58024933f017cc85eb77579779311b65f


```adb
! https://github.com/AdguardTeam/AdguardFilters/issues/120891
||p.typekit.net^$stylesheet,removeparam
```

PR #14775, opened 2023-01-24
M: Allow p.typekit.net/p.css for robertsspaceindustries.com

AdrienPoupa:
Given the typekit rule is here to stay (see https://github.com/easylist/easylist/pull/14759), this is adding robertsspaceindustries.com to the allow list for `p.typekit.net/p.css`.

The current impact is the store is not loading at all with Easylist enabled, see details here: https://github.com/easylist/easylist/issues/14723#issuecomment-1400797333


0o5rlubm (approved):

AdrienPoupa:
Thank you!

PR #14774, opened 2023-01-24
M: pandora.tv

piquark6046:
The pandora.tv service will be ended in February 1st, 2023 (KST).
http://www.pandora.tv/company/notice/view/1/1178

 - https://github.com/List-KR/List-KR/issues/572
 - https://github.com/AdguardTeam/AdguardFilters/pull/140855

PR #14833, opened 2023-01-30
M: (nsfw)`tubesweet.com`

eporubin:
Please remove from the list. 
image
User complained the page is broken when blocking ads


Yuki2718:
@eporubin See https://github.com/easylist/easylist/pull/14774#issuecomment-1412244081

PR #14844, opened 2023-01-31
D: `##.quads-location `

eporubin:
Hi, please remove filter from the generic it's breaking this page 'https://subsbr.net/' appears absolutely blank and the users cant access it.


Yuki2718:
Removing the rule by jsut signle false positive instead of `#@#`? I know the rule is mostly redundant by other rules but hope this does not cause false negative.

Yuki2718:
An example of FN: `https://www.freesteamkeys.com/`. Maybe `###quads-ad8` or `##.quads-ad8` can be added.

PR #14848, opened 2023-01-31
A:oneindia.com

mediumkreation:
https://github.com/mediumkreation/IndianList/issues/6692


ryanbr:
`oneindia.com#?#.one-news-links ul li:-abp-has(> div:only-child)` in specific_abp

PR #14857, opened 2023-02-01
A:timesnownews.com

mediumkreation:


ryanbr:
`timesnownews.com#?#div:-abp-has(> .bggrayAd)` needs to be _abp list

mediumkreation:
> `timesnownews.com#?#div:-abp-has(> .bggrayAd)` needs to be _abp list

Modified.

PR #14875, opened 2023-02-02
A: `hornoxe.com`

eporubin:
Hi, could you please ad this filter to hide this links to social media platforms on the page. 
image

Thanks a lot in advance


ryanbr:
`hornoxe.com##.custom-html-widget`  should be enough

eporubin:
Hi, @ryanbr Thanks for highlighting it, I corrected the filter as you suggested.

PR #14923, opened 2023-02-08
A: https://plus.nhk.jp/watch/ch/g1

evans811:
This change adds a new filter to Easylist's specific hide list to hide the large sign-in overlay on the NHK Plus player.

For reviewers: Please note that NHK Plus is only available in Japan and may require a VPN to access from outside the country. It also only supports Microsoft Edge, Google Chrome (or other Chromium-based browsers), and Safari, and requires DRM to be enabled for playback.


evans811:
Below is a screenshot of the NHK Plus interface. There is a large modal at the top of the player asking the user to register. This modal can be removed by blocking `.hls-player_message`.

Yuki2718:
Not ad

PR #14944, opened 2023-02-11
A: https://www.amazon.fr/gp/buy/spc/handlers/display.html

0o5rlubm:
I suggest following to hide the annoying big Subscribe to Amazon prime banner. Note that numbers of customers have fallen in the trap of this page, I know few ones that subscribed to Prime during checkout without having noticed it.

I've tried to block only this one and not all other Prime links on the page that use same kind of class, that explain the 3 selectors.

Even if it sounds good to me it would be nice to have a second tester due to the impact :-)

Before
Screenshot 2023-02-11 at 01-18-01 Processus de paiement Amazon com

After
Screenshot 2023-02-11 at 01-18-18 Processus de paiement Amazon com


Khrin:

Issue #15037, opened 2023-02-21
[Fanboy's Social] Add Shareon to the list


kytta:
### List the website(s) you're having issues:

<https://shareon.js.org/>

### What happens?

I am a developer of the [Shareon](https://github.com/kytta/shareon) share buttons. I envision those to be as ethical as possible, which is why I also would like them to actually be blocked by the ‘social’ ad blockers.

Right now, when I use the Fanboy's Social Blocking List (uBO edition), only some buttons supported by Shareon disappear. Others continue to function. Although my buttons aren't sending any data anywhere, I think it would make sense to block the element as well as the init script for them.

### List Subscriptions you're using:

- uBlock built-in filters
- EasyList
- EasyPrivacy
- Actually Legitimate URL Shortener Tool (by DandelionSprout)
- Online Malicious URL Blocklist (urlhaus)
- Anti-Facebook
- **Fanboy's Social**
- Peter Lowe’s Ad and tracking server list
- EasyList Germany
- StevenBlack/hosts

### Your settings

- OS/version: macOS Monterey version 12.6.2 (21G320)
- Browser/version: LibreWolf 110.0-1 (based on Firefox 110)
- Adblock Extension/version: uBlock Origin 1.47.0

### Other details:

If you want to block Shareon, you'd need to block:

- the `.shareon` CSS selector
- `shareon.*.js` which loads the script to initialize the buttons
- `shareon.min.css` which loads the styles for the buttons

[See the docs](https://shareon.js.org/#install) for more info

krystian3w:
Something was added secretly: https://github.com/easylist/easylist/commit/b7d4259325bd4f1846c2146095c2c12515132a0a

ryanbr:
so secret it was referenced in https://github.com/easylist/easylist/pull/15038

PR #15038, opened 2023-02-21
Add Shareon

kytta:
Closes https://github.com/easylist/easylist/issues/15037

### DIFF
--- a/fanboy-addon/fanboy_social_general_block.txt
+++ b/fanboy-addon/fanboy_social_general_block.txt
@@ -1515,6 +1515,7 @@
 /shareit.
 /shareit_
 /sharelike.
+/shareon.
 /ShareOnMySpace/*
 /sharepage-
 /sharePost/*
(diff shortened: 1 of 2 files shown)


ryanbr:
This would create false positives on 1st party content

kytta:
> This would create false positives on 1st party content

Yeah I thought that too, I just took example of the other list entries. How would one make something like this possible? Or is the name too generic for this?

kytta:
> `##button[ng-click^=shareOn]`

I want to point out that this entry has nothing to do with my package, as it's not an Angular component :)

---

> `/shareon.` has the slight drawback that it would also match `https:/`**`/shareon.`**`js.org`, but there are ways to narrow it down, from the common (`/shareon.$third-party`) to the rather rare (`/[a-z0-9]\/shareon\./`)

Aw shucks, you're right! Well, one could do a list of all possible scripts:

```
/shareon.es.js
/shareon.iife.js
/shareon.umd.js
/shareon.min.css
```

krystian3w:
Also works:

```adb
/shareon.$script,stylesheet
```

due homepage don't use first party scripts and css files yet.

krystian3w:
https://github.com/easylist/easylist/issues/15037

PR #15111, opened 2023-03-02
A: block web crypto miner

XfedeX:


ryanbr:
Won't just block these domains without working examples.

ryanbr:
https://publicwww.com/websites/%22%2Fwebmnr.min.js%22/

PR #15114, opened 2023-03-02
M: Remove datadoghq.eu

yoannmoinet:
We're working on using this domain to serve our assets.
So it's not testable right now, but it did block us when we tried to move forward with that.

I don't understand why it's been added in the first place as we don't do tracking from our top domain names to avoid this kind of situation.

[Commit](https://github.com/easylist/easylist/commit/e38b423ecfa0ccba7e8e5eae40fe4fb8a54cf336) where it's been added looks odd as it doesn't state the same domain name in the commit message.
Maybe it was a mistake?

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -923,7 +923,6 @@
 ||datacoral.io^$third-party
 ||datacygnal.io^$third-party
 ||datadoghq-browser-agent.com^$third-party
-||datadoghq.eu^$third-party
 ||datadsk.com^$third-party
 ||datafeedfile.com^$third-party
 ||datafront.co^$third-party


ryanbr:
>I don't understand why it's been added in the first place as we don't do tracking from our top domain names to avoid this kind of situation.


`browser-intake-datadoghq.eu/api/v2/logs?`

ryanbr:
Datadog _is_ a tracking company, why should we care? why should I care its blocked.

yoannmoinet:
We've just noticed that this rule broke our website.

I'm just trying to find a solution to that, I'm really not trying to trick you into allowing a tracker.

Can we maybe create an exception so it's at least not blocked for `datadoghq.com` domains?
Like if we're on `datadoghq.com` do not block `datadoghq.eu`.
Not sure if possible or not.

Khrin:
This request is totally unacceptable.

PR #4212, opened 2019-10-28
P: Removed getrewardful.com (fix #4170)

smed79:

PR #15120, opened 2023-03-04
P: Removed partnerstack.com

nchudleigh:
### Filter affected:
Here is the commit partnerstack.com was added to the adserver list in https://github.com/easylist/easylist/commit/b27cfaba9441d860ac8351bbeef505a6af5b778c

### 1st/3rd-party sites affected:

PartnerStack.com is our primary domain which hosts the marketing website, and our applications. 
Our customers using ad blockers have been contacting support because a warning screen is displayed for the application.

All of partnerstack applications and sites, this is just the most trafficed subset, there are other sites affected as well.

- partnerstack.com
- app.partnerstack.com
- dash.partnerstack.com
- market.partnerstack.com
- *.partnerstack.com (all of our customers hosted program pages e.g. https://mondaycom.partnerstack.com/)

### How is it broken?
Users presented with adblock warning screen before visiting any web property.

### Description of why it should be removed:
👋 Hi, I run PartnerStack — the company this domain belongs to.

I believe the rule was added because of our tracking snippet at https://js.partnerstack.com/v1 

Fair enough to add js.partnerstack.com domain with a third party script rule but after reviewing your policies I don't think the current rule is fair. It is causing severe stress on our business and users (many of whom make their living on our site)

The affiliate links for PartnerStack application are hosted on grsm.io and partnerlinks.io. These domains are listed in the easylinks_adservers block list alongside partnerstack.com. While it would be more accurate to put them on a tracking list, this is acceptable given the easylist policy.

I took the liberty of finding a similar case, linked below- Rewardful provides the same services as PartnerStack and also had their domain added in error.
https://github.com/easylist/easylist/pull/4212

Thank you for looking into this, and for the continued work on Easylist, I have been a happy uBO user for many many years.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -13590,7 +13590,6 @@
 ||partion-ricism.xyz^
 ||partnerlinks.io^
 ||partnershipknackcloset.com^
-||partnerstack.com^
 ||partoukfa.com^
 ||partsnoises.com^
 ||partyguarantee.com^


nchudleigh:
@Yuki2718 Thanks for the the commits! Anything else I can do to help this along? 

Our support team is having a very difficult time fielding all the questions & tickets from our customers. Every single one of our properties (docs, landingpage, apps, etc) is affected.

Yuki2718:
@nchudleigh I unblocked only your main site. Does the third-party rule still break your applications, IOW they're hosted on your customer sites as third-party? Can you provide an example we can check?

nchudleigh:
Yep that's legitimate blocking, though we do have other cases of third party scripts that are not tracking related.

Would it be possible to narrow the rule be just js.partnerstack.com?

Yuki2718:
Possible but need confirmation that the current rule actually causes trouble for user.

nchudleigh:
For sure, I am just educating myself on the rulesets and what exactly they do. I am new to the inner workings of these lists and rules. 

Also want to clarify, we are talking about the commits you made to the adguard and ublockorigin repos?

Thanks again for helping out with this.
This is really impacting our users/customers and as a result our support team.
So on their behalf, thank you for your attention to our case.

nchudleigh:
Here is an example implementation of our sign up referral
https://s2sreviews.com/referral

Here are the docs for the feature.
https://docs.partnerstack.com/docs/referral-iframe

nchudleigh:
@ryanbr @Khrin I updated the description to match the dispute template. Apologies, I didn't notice it in the repo previously.

Khrin:
Obviously this request can't be accepted.

Yuki2718:
@Khrin What about their site itself i.e. `$third-party`?

PR #15130, opened 2023-03-06
https://st1.zoom.us/zoom.ico

yellitout:
A proposal to block Zoom's favicon basically being [a symbol of russian invasion](https://en.wikipedia.org/wiki/Z_(military_symbol)).

Using the letter "z" as a favicon is inappropriate and insensitive, particularly in light of the tragic reality that this very same symbol is being used by Russians to kill thousands of innocent people. It goes without saying that such a use of the letter "z" is deeply offensive and deeply problematic. We should be mindful of our symbols and their potential impact on others, particularly in the context of such violence and suffering.


Khrin:
@yellitout If you feel the need for this filter, you can add it as custom filter. The EasyList filters list serve to block advertisements.

PR #15147, opened 2023-03-08
A: https://www.skynewsarabia.com/

smed79:
- https://github.com/easylist/listear/issues/29


Khrin:
@smed79 Are you sure that needs to be blocked?

smed79:
for me Yes. 
Because the ribbon is persistent and back even if you close it.

PR #15172, opened 2023-03-09
M: Update

eporubin:
Hi, 👋  Amazon complaint in regards to the existing filters and provide the following explanation: `The analytics resources are exclusively used for 1st party performance – how fast the amazon site loads, which content is seen on the site and how customers interact with the site, error tracking. No 3rd party data sharing, cookies or cross-site user tracking.` Could you please remove or would you rather want them allow-listed? please let me know.  Thank you in advance.


Khrin:
I think the actual filters can stay.

PR #15197, opened 2023-03-12
A: https://fr.delsey.com

0o5rlubm:
Elevar is a tracking company for shop websites
blocking it does not seams to block delsey.com order process


ryanbr:
The it $third-party and move to the tracking servers list.

0o5rlubm:
Done :smile_cat:

PR #15282, opened 2023-03-23
M: Fix for `www.insiderintelligence.com`

eporubin:
Hi, 👋  could you please add the allowlisting filter to the list.  Easy privacy breaks the search engine (top bar one ☝️ ) on this website `www.insiderintelligence.com`. This filter fixes the issue. Thank you in advance.


spirillen (commented):
Doing this would opens op a huge gap for tracking....

PR #15343, opened 2023-03-29
[nsfw] - fancentro.com

miller-cntr:


ryanbr:
Its a fake site, so no.

SkyCloudDev:
`fancentro.com` is not a fake site, its a legitimate competitor to sites like `onlyfans.com`, could you please remove it from the list so that sites with written links to fancentro can display them rather than being hidden.

PR #15417, opened 2023-04-03
Remove !#include uBO Fanboy notifications

peace2000:
Because of this entry, the whole notification list is now unloadable in uBO, as easylist-downloads.adblockplus.org doesn't host this file: fanboy_notifications_specific_uBO.txt

This doesn't work in uBO: https://easylist-downloads.adblockplus.org/fanboy-notifications.txt

As now this whole list is unloadable in uBO, It's better to leave uBO section out rather than drop the whole list from uBO. If this change is made, the list should work in uBO as before.


krystian3w:
https://github.com/uBlockOrigin/uBlock-issues/issues/2610

PR #15468, opened 2023-04-08
A: https://www.sbstatesman.com/

Nojuuu:
image
image
image


Nojuuu:
That removes only the items that are displayed in the last pic?

PR #15531, opened 2023-04-14
A: https://www.playpilot.com/

Nojuuu:
Hides ads that appear when you search something:
image


Nojuuu:
Could this file also be sorted alphabetically? It seems a bit messy at the moment.

PR #15626, opened 2023-04-21
Do not apply shopify/track rule to github.com

winston-yallow:
This fixes uBlockOrigin/uAssets/issues/17706 by preventing the `/shopify/track` to be applied on `github.com`


winston-yallow:
Oh I should have read the readme first :facepalm: 
> It is preferable to submit issues instead of pull requests, because the EasyList team will need to look through the sites anyway, and pull requests can fall victim to rebase errors.

Just thought "this is an easy fix" and made a PR, but feel free to close this if it doesn't work well with the issue based workflow you have. The issue already exists in the uBlock repo (linked above)

PR #15706, opened 2023-04-28
A: hs-scripts.com

domdomegg:
This domain serves the HubSpot tracking scripts: https://knowledge.hubspot.com/reports/install-the-hubspot-tracking-code

I think some resources may be necessary on it to support the HubSpot admin UI, so have exlcuded it on that site.


Khrin:
https://github.com/easylist/easylist/commit/b2d4cdf566632b3dc838c1519b84ff85e818e46a

PR #15709, opened 2023-04-28
A: data.tm-awx.com

domdomegg:
Appears to be related to https://www.smileweb.net/en/platform, a tracking server used across a number of news websites e.g. https://www.mirror.co.uk/


ryanbr:
Not sure if its a tracker, we have some exisiting rules already for tm-awx.com. Don't want to break anything

PR #15763, opened 2023-05-02
A: https://support.wahoofitness.com

0o5rlubm:


Yuki2718:
Does blocking `https://scripts.swifteq.com/hc_events.js` cause problem?

0o5rlubm:
I'm not enought familiar with this specific site to block the js file, it requires more testing I than what I did I guess

Issue #15775, opened 2023-05-03
jamesallen.com


princesa42:
### List the website(s) you're having issues:

`jamesallen.com` and `bluenile.com`

### What happens?

The Javascript used to track affiliate purchases is blocked, which prevents the affiliate part of the business from performing proper purchase attribution.

If you load either of the below pages with the a_aid parameter set, you will see the request to affiliates.r2net.com fails.
`https://www.jamesallen.com/loose-diamonds/round-cut/1.00-carat-i-color-si1-clarity-very-good-cut-sku-17349217?a_aid=XXX`
`https://www.bluenile.com/diamond-details/LD19911125?a_aid=XXX`

The domain that is blocked is affiliates.r2net.com. R2Net is part of Signet, which owns James Allen and Blue Nile.

`https://www.businesswire.com/news/home/20170824005546/en/Signet-Jewelers-to-Acquire-R2Net-owner-of-JamesAllen.com-to-Accelerate-its-Customer-First-OmniChannel-Strategy`

### List Subscriptions you're using:

The default list that comes with Adblock Plus: ABP filters and EasyList

### Your settings

- OS/version: MacOS Big Sur
- Browser/version: Google Chrome - Version 112.0.5615.137 (Official Build) (x86_64)
- Adblock Extension/version: Adblock Plus - 3.17

### Other details:

First two screenshots show problematic blocking. Second two show requests successful after adding the following rule to "My Filter List" in Ad Block Plus:

`@@||affiliates.r2net.com^$domain=jamesallen.com|bluenile.com`

#### Evidence:
![screenshot_bluenile](https://user-images.githubusercontent.com/132415580/235912324-cf3cee8a-e996-46de-85ac-d84b56072311.jpg)

![screenshot_jamesallen](https://user-images.githubusercontent.com/132415580/235910991-5c821d04-ea74-498c-9cb2-69294e95af29.jpg)

#### Fixed:
![screenshot_bluenile_fixed](https://user-images.githubusercontent.com/132415580/235912440-5415b53f-9244-4fd4-957b-9d8ffb90cdb8.jpg)

![screenshot_jamesallen_fixed](https://user-images.githubusercontent.com/132415580/235912475-530c2af2-d81c-4db1-b656-31e073041f3b.jpg)

ryanbr:
Happy to leave it.

PR #15776, opened 2023-05-03
A: affiliates.r2net.com

princesa42:
Fixes https://github.com/easylist/easylist/issues/15775


ryanbr:
Not needed, we'll block it.

PR #15784, opened 2023-05-03
M: cleanup redundant google entries

peace2000:
Removing entries that I didn't saw in logger when doing google searches.


monzta:
`##.uEierd` still needed for Google with disabled JS (https://github.com/easylist/easylist/commit/cc492c4)

peace2000:
Frankly, I wouldn't believe they could have a separate element for non-JS-version of their site....

Made a new PR, excluding that entry.

https://github.com/easylist/easylist/pull/15878/

@monzta

monzta:
Thank you!

PR #15901, opened 2023-05-12
M: https://www.valostore.fi/

Nojuuu:
update cookie hide


peace2000:
As they change element names, maybe better to form a filter that lasts even if they change element names?

Nojuuu:
Idk, sometimes these can survive long time too, so let's see how long it takes before it changes again.

PR #16033, opened 2023-05-24
Remove lead.im private service domain

psyafter:
We are lead.im private solution.
Our scripts not collect any private information

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_international.txt
+++ b/easyprivacy/easyprivacy_trackingservers_international.txt
@@ -506,7 +506,6 @@
 ||enter-system.com^$third-party
 ||erate.co.il^$third-party
 ||fortvision.com^$third-party
-||lead.im^$third-party
 ||mimgoal.com^$third-party
 ! Hungarian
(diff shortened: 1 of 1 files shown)


ryanbr:
`http://cdn.lead.im/js/lm_url2form.js`  Do explain.

psyafter:
> `http://cdn.lead.im/js/lm_url2form.js` Do explain.

This script save query parameters into html form, so customer will know which pages he  visited in his session.

ps: If it something that not permitted - why block whole domain but not this script only?

ryanbr:
https://github.com/easylist/easylist/commit/db37300a0828074cb16145a9098c6ecce91c785a Not a new filter. Also a company that tracks can't be trusted, easier to change a file name than a domain.

psyafter:
what you mean? I can change file name. I can remove using on original page.
Can this filter to be changed to block single file instead of domain?

ryanbr:
based on another tracking script `https://cdn.lead.im/api/v2/leadim.api.js` we're leaving the block.

Issue #16044, opened 2023-05-25
Indeed.com broken due to thirdparty adblock of CDN


bencripps:
#### List the website(s) you're having issues with:

indeed.com

#### What happens?

All static files will 404, and the page will be broken. 

![image](https://github.com/easylist/easylist/assets/6272507/1f61debe-9726-450e-aad0-12296131d7e3)

#### Filter lists you're using:

EasyList

#### Your settings:

- OS/version: 

OSX 
- Browser/version: 

Chrome

krystian3w:
https://github.com/easylist/easylist/pull/16045

krystian3w:
9777d69

krystian3w:
https://github.com/easylist/easylist/commit/9777d69886b67637a3076a1151ec4b3a08aa10df

PR #16045, opened 2023-05-25
Remove CDN address from 3rd party list

bencripps:
See issue https://github.com/easylist/easylist/issues/16044


AlexeyLop:
upvote

PR #16063, opened 2023-05-26
M: removed too generic GDPR blocking script

peace2000:
This script is part of `cookielaw.org` stuff.

cookielaw.org is already a site specific in Easylist Cookie:
https://github.com/easylist/easylist/blob/66801ddc81f9d0a5c2dc69f3038b8040a0b41ed7/easylist_cookie/easylist_cookie_specific_block.txt#L189

Sample page where it can be seen that script is loading from cookielaw.org: https://www.mtvuutiset.fi/artikkeli/kilpikonna-kiitoradalla-viivytti-viitta-lentoa-japanissa/8248130

kuva

### DIFF
--- a/easylist_cookie/easylist_cookie_general_block.txt
+++ b/easylist_cookie/easylist_cookie_general_block.txt
@@ -892,7 +892,6 @@
 /opnWidget.js
 /opt-in-select/cookie/*
 /optInCookieGdprSSR.
-/otBannerSdk.js
 /oupcookiepolicy_
 /PDCookieConsent_
 /pescookies.


peace2000:
It was added in this commit https://github.com/easylist/easylist/commit/664508d2ee9b6d529a1f21ca13481ecc16676759

You might want to add those domains to site specific `cookielaw.org` rule.

ryanbr:
I've seen instances where "otBannerSdk" being hosted on the own server, which is why its a generic block. Also we're not targeting cookielaw.org specificaly

peace2000:
I see your point. However, if we use `/otBannerSdk.js` as a generic, it will result the same breakages as blocking `||cookielaw.org^` because `otBannerSdk.js` is always present with `cookielaw.org`.

I rewiewed the first half of these cases: https://github.com/easylist/easylist/blob/66801ddc81f9d0a5c2dc69f3038b8040a0b41ed7/easylist_cookie/easylist_cookie_specific_block.txt#L189

And it seems that almost each time if we just block `/otSDKStub.js`, it will result the rest of the `cookielaw.org / onetrust` requests getting blocked as well.

So what I'm suggesting is that we'd change specific `||cookielaw.org^` blocking rule to specific `/otSDKStub.js` blocking rule (and remove the generic `/otBannerSdk.js` as it will cause breakages)

And if there are any remnant requests left (as with `boxrox.com` there is `/tcf.stub.js`) they would be handled with specific filters as well?

ryanbr:
Leaving it.

peace2000:
@ryanbr 

mtvuutiset.fi still needs an exception.

PR #16187, opened 2023-06-03
A: www.natureswaydelivery.com

Yuki2718:


ryanbr:
Does that break the chat app?

Yuki2718:
> Does that break the chat app?

Apparently yes: https://github.com/AdguardTeam/AdguardFilters/issues/152530
@Alex-302 Should we remove the rule?

Alex-302:
@Yuki2718 Removed

PR #16241, opened 2023-06-09
A: https://www.bing.com/

CbumDino:
Go to Bing.com (Edge browser) with the "show news and interests" options enabled. Scroll down to actually get to the news and interest bar (below the search bar). Click on any of the stories. On the search results page that opens, you'll get a refreshed "news and interests" bar (as see in the image capture). In that bar mixed into the stories are ads - you can tell they are ads because it has the word "ad" as a watermark in the upper left-hand corner.

Screenshot 2023-06-09 at 15 49 34


ryanbr:
Are these "ads" ?

Issue #14486, opened 2022-12-23
cdon


Nojuuu:
### List the website(s) you're having issues:

`https://cdon.dk/`
`https://cdon.fi/`
`https://cdon.no/`
`https://cdon.se/`

### What happens?

![image](https://user-images.githubusercontent.com/84513173/209394365-ed0715da-f320-4cb0-8a48-651c97439f77.png)

### List Subscriptions you're using:

EasyPrivacy

### Your settings

- OS/version: Windows 11
- Browser/version: Firefox 108.0.1
- Adblock Extension/version: uBlock Origin 1.45.2

### Other details:

--

0o5rlubm:
could you clarify what is the problem on these pages?

Nojuuu:
Analytics/tracking not blocked.

0o5rlubm:
thanks for the precision.
I well see the load of an analytics.js script file. But there is no data sent back to the website when i'm loading  https://cdon.se .
The only tracking stuff (not related to the one you mentioned) is Akamai bot manager (huge urls like https://cdon.se/lk2EHJ/n/x/ZbYuIKeHbMJm/iGSYQDDp1zr7/IDZSAQ/C2ge/NRYBK1YB), that blocking it may have complex repercussions.

PR #16262, opened 2023-06-11
A: cdon.*

Nojuuu:
Fix https://github.com/easylist/easylist/issues/14486


ryanbr:
`$script` seems to be redundant filter

Nojuuu:
Removed

Issue #15302, opened 2023-03-25
8tm.net


Sonic5435325:
### List the website(s) you're having issues:
https://8tm.net/

### What happens?

An image is being blocked on the site because of the filter `/fbpixel.` in EasyPrivacy which is fixed by adding `@@||8tm.net/static/img/fbpixel.png$image,1p` to My Filters.

Screenshot 1 (Normal, no filters):
![Screenshot from 2023-03-24 20-53-39](https://user-images.githubusercontent.com/89158249/227669979-5778d453-37cd-4fc0-b0b9-83917a2261ec.png)

Screenshot 2 (With an exception filter):
![Screenshot from 2023-03-24 20-53-48](https://user-images.githubusercontent.com/89158249/227670152-7a2376a7-8abf-4d81-9f20-9139ba507079.png)

### List Subscriptions you're using:

EasyPrivacy

### Your settings

- OS/version: 
- Browser/version: 
- Adblock Extension/version:

PR #16279, opened 2023-06-12
Update easyprivacy_general.txt

Sonic5435325:
To address #15302
If not too broad, then `@@||8tm.net/static/img/fbpixel.png`


ryanbr:
Not removing.

Sonic5435325:
Maybe could an exception filter be added in then? @ryanbr

PR #16281, opened 2023-06-12
Update easyprivacy_thirdparty.txt

Sonic5435325:
Add aspiringtoy.com Admiral domain


ryanbr:
Which site? also should be in the Admiral Section in easyprivacy_trackingservers.txt

PR #16291, opened 2023-06-13
Remove visit-prod-us.occa.ocs.oraclecloud.com

leocunhaO:
Hello there,
I'm writing this PR to remove this block, which prevents the user from loading the cart page when using this Oracle Commerce product.

Blocked URL:
https://visit-prod-us.occa.ocs.oraclecloud.com/Visit/js/oracleunifiedvisit.js

Steps to reproduce:
You will not be able to reproduce the cart page issue by yourself because the e-commerce is being tested and is not live yet. It happens on a URL that is only accessible by our client's VPN and other internal credentials. You can test the blocked URL directly in the browser, which is the full path of the request. It will block the request with extensions like uBlock, which uses EasyList as a reference for filters: 

1. Activate uBlock browser extension
2. Try to access https://visit-prod-us.occa.ocs.oraclecloud.com/Visit/js/oracleunifiedvisit.js in your browser

uBlock extension, message blocking the URL:

image

Filter blocking the URL: 

||visit-prod-us.occa.ocs.oraclecloud.com^

The functionality of the script:
It exposes the OracleUnifiedVisit object with various properties and methods related to visitor tracking, including the account ID, visitor ID, visit ID, secondary ID, and functions to set XD (cross-domain) settings, EEID (Experience Edge ID), and host settings. These are the core functions of the cart page, so the user can add items to the cart, go back to the page and still have them in their cart. It is very critical to the functionality of the cart, it won't even show any products and it will prevent the user from continuing to checkout.

Best Regards,

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty_international.txt
+++ b/easyprivacy/easyprivacy_thirdparty_international.txt
@@ -731,7 +731,6 @@
 ||track.noddus.com^
 ||tracker.tolvnow.com^
 ||trrsf.com.br/metrics/
-||visit-prod-us.occa.ocs.oraclecloud.com^
 ||webstats.sapo.pt^
 ||xl.pt/api/stats.ashx?
 ! Romanian


Khrin:
Allowing a known tracker to load isn't a good idea.

Come back when the "site" is live and we'll see.

PR #16317, opened 2023-06-15
A: https://areena.yle.fi/

Nojuuu:
tracking

### DIFF
--- a/easyprivacy/easyprivacy_specific_international.txt
+++ b/easyprivacy/easyprivacy_specific_international.txt
@@ -573,6 +573,7 @@
 ||analytics.sanoma.fi^
 ||api.nettix.fi/counter/$image
 ||data.reactandshare.com^
+||dax.yle.fi/yle/areena/s$image
 ||dax.yle.fi/yya/event
 ||dp.alma.iltalehti.fi/v1/cookie
 ||events.il.fi^


Nojuuu:
@peace2000 

content not loading with these:?

Easyprivacy: `/analytics-sdk.`
AdGuard tracking protection: `||analytics-sdk.yle.fi^`

`https://areena.yle.fi/tv`
`https://areena.yle.fi/tv/ohjelmat/sarjat`

peace2000:
Yeah it seems to.

How about:

`areena.yle.fi##+js(set, window.yleAnalytics, noopFunc)` ?

Nojuuu:
> How about:
>areena.yle.fi##+js(set, window.yleAnalytics, noopFunc) ?


That seems to help

`dax.yle.fi/yle/areena/s` should probably be blocked if `analytics-sdk.yle.fi` gets allowlisted here

peace2000:
I'd say that just let it be. This can be fixed in uAssets and in AG.

ABP doesn't use EP as a default list anyway...

I remember that `areena.yle.fi` was fixed long time ago but probably the fix has been removed as part of cleaning operations I think.

peace2000:
The best would be if EP would contain a different section for uBO to so that scriptlets could be added but this was refused https://github.com/easylist/easylist/pull/8317 (I think it's because policies of ABP.....)

peace2000:
Fixed for uBO https://github.com/uBlockOrigin/uAssets/commit/3550cc1197e9857038f711e725980f004af410f9

peace2000:
@Yuki2718 

Can you add Adguard version of this filter:

`areena.yle.fi##+js(set, window.yleAnalytics, noopFunc)`

To Adguard tracking protection?

Yuki2718:
> 
> Can you add Adguard version of this filter:
> 

A problem is AG has wide range of products some not supporting scriptlet. What should I add for platforms not supporting scriptlet, `@@||analytics-sdk.yle.fi^$domain=areena.yle.fi`?

peace2000:
@Nojuuu 

Too bad EP whitelisted that https://github.com/easylist/easylist/commit/d6e068f960d9ba6771e4b50ef6515e195290716b

Let's block it then on uAssets https://github.com/uBlockOrigin/uAssets/commit/30e49d3d79afcd9b34a28929d0b1a6e87ba35a57

peace2000:
> > Can you add Adguard version of this filter:
> 
> A problem is AG has wide range of products some not supporting scriptlet. What should I add for platforms not supporting scriptlet, `@@||analytics-sdk.yle.fi^$domain=areena.yle.fi`?

Well, if some products on AG won't support set-constant then that needs to be whitelisted. The breakage is not minor so it should be fixed some way.

Nojuuu:
@Khrin @Yuki2718 
Blocking at least `||dax.yle.fi/yle/areena/s` should be fine

image

peace2000:
Ok, that set-constant rule was bad after all.

https://github.com/uBlockOrigin/uAssets/commit/30e49d3d79afcd9b34a28929d0b1a6e87ba35a57#commitcomment-118531536

PR #16342, opened 2023-06-18
A: https://www.gizmodo.jp/2011/03/fox_news.html

koichit806:
This change blocks ads from ad company [Mediagene](https://www.mediagene.co.jp/) displayed at the bottom of Gizmodo articles.

### DIFF
--- a/easylist/easylist_specific_block.txt
+++ b/easylist/easylist_specific_block.txt
@@ -264,6 +264,9 @@
 ||masternodes.online/baseimages/
 ||maxgames.com/img/sponsor_
 ||mbauniverse.com/sites/default/files/shree.png
+||media-assistant.tech^
+||media-platform.com^
+||mediagene.co.jp^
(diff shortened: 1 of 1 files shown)


ryanbr:
Non english request, is there an english site showing these?

koichit806:
It seems that Mediagene is only serving ads on Japanese-language websites. The section immediately after the main content showing price tags under pictures of bags and wallets is the Mediagene ad in question.



This can be removed by blocking the three domains I provided.

The company [makes it clear](https://www.mediagene.co.jp/media/gizmodo) that it is serving ads on Gizmodo, and Whois queries for media-assistant.tech and media-platform.com show that the other two domains are owned by Mediagene.

Yuki2718:
Thank you for informing. `media-assistant.tech` is already blocked, but `media-platform.com` seems to fit into EP. Will evaluate later today and if good will open a PR (anyway needs `$third-party` at least).

PR #16349, opened 2023-06-18
M: theverge GDPR dialog already covered by a generic cosmetic rule

peace2000:
https://www.theverge.com/

This covers it already:
`##.duet--cta--cookie-banner`


ryanbr:
No harm with it, removing the class is quicker and avoid the cookie message "flash"

peace2000:
@ryanbr 

For flickering issues, it's better to use a specific cosmetic filter, instead of scriptlet. That scriptlet causes flickering on my end which is why I paid attention to this issue in the first place. :)

theverge.com##.duet--cta--cookie-banner

That doesn't cause flickering.

in uBO, sometimes generic cosmetics flicker, because uBO doesn't load all tens of thousands generic cosmetics unconditionally on each page, unlike e.g. ABP does. uBo scans first whether a generic needs to be applied or not.

This saves cpu and memory, but as a side effect, a generic cosmetic filter might not be applied fast enough, and it might flicker.

These issues can be resolved by adding a specific cosmetic filter.

peace2000:
> I'm not convinced. we already have .duet--cta--cookie-banner


What you are not convinced about? What I said about uBO's way to operate? It's documented here: https://github.com/gorhill/uBlock/wiki/Does-uBlock-Origin-block-ads-or-just-hide-them%3F#cosmetic-filters

@ryanbr

PR #16395, opened 2023-06-21
A: https://www.lemonde.fr/

0o5rlubm:
Following cat-and-mouse game with this website, that [regularly update it's tracking url to avoid being blocked](https://github.com/easylist/easylist/pulls?q=is%3Apr+lemonde+is%3Aclosed) to continue to track users, I'm afraid we don't have any other solution than blocking images (request type used by tracker) on this host. Fortunately, article images are on subdomain.

The only useful image we'll miss is the small newspaper image on top left, but link still work so the block is still a win I guess.


ryanbr:
Not sure if blocking all images would be a good thing here. For Webcompat not a good idea

PR #16434, opened 2023-06-23
A: http://entywbas.com/ & https://washwasha.org/

smed79:


Khrin:
@smed79 This should be added under: https://github.com/easylist/easylist/blob/master/fanboy-addon/fanboy_social_specific_hide.txt#L764

PR #16458, opened 2023-06-25
M: yle.fi GDPR dialog change to better

peace2000:
https://yle.fi/a/74-20009022

Set social media cookies accepted so that embedded content works. Also hide the dialog.


peace2000:
Solved conflicts.

I'm following the way how they have been set here lately (own section for each filter).

PR #16469, opened 2023-06-26
A: tweakers.net GDPR dialog (uBO)

peace2000:
https://tweakers.net/nieuws/197852/portal-achtige-puzzelgame-the-entropy-centre-komt-in-2022-uit.html


krystian3w:
Compatible with EL Cookies.

PR #16510, opened 2023-06-29
A:`373news.com`

eporubin:
Hi could you please review and add this filter to the list, helps removing lots of ads embedded on the page. 
image
 I know the domain is Japanese but can we please keep it in here for now. 
 Thank you


eporubin:
Banner ads: 
image

easylist/easylist_specific_block.txt
```diff
@@ -5,6 +5,7 @@
 ||1337x.*/images/x2x8.jpg
 ||2merkato.com/images/banners/
 ||2oceansvibe.com/?custom=takeover
+||373news.com/mns_images/*
```
Yuki2718, line 8: Badly braeks the site, please remove.

easylist/easylist_specific_hide.txt
```diff
@@ -1116,6 +1116,7 @@ computerworld.com,networkworld.com,probuildstats.com##.ad-placement
 comedy.com##.ad-placement-wrapper
 beincrypto.com##.ad-post-right-sidebar-1
 beincrypto.com##.ad-post-right-sidebar-2
+373news.com##.ad_response
```
Yuki2718, line 1119: Need to be generic for two reason:
1. We don't add regional specific rule
2. This is not unique to the site, common in all Uzou recommendation

Yuki2718 (changes requested):

Yuki2718:
> Banner ads:

No, you'll see they're linked to `373news.com`. Do not confuse self-promotion with ads which we do not block: https://easylist.to/2011/07/11/the-definition-of-advert-and-link-exchange-policy.html

easylist/easylist_specific_block.txt
eporubin: The filter was working fine for me. I cant see any other breakage.
I removed as you said it's self-promotion.

Issue #16041, opened 2023-05-25
docs.github.com


Sonic5435325:
#### List the website(s) you're having issues with:

`https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections#creating-a-collapsed-section`

#### What happens?

Back to Top button annoyance
`docs.github.com##[aria-label="Scroll to top"]`

#### Screenshot(s) 

![Screenshot from 2023-05-24 20-16-55](https://github.com/easylist/easylist/assets/89158249/3475dfee-d811-46db-9808-48bb3a6de9c3)

#### Filter lists you're using:

Fanboy's Annoyance/Other Annoyances

#### Your settings:

- OS/version: 
- Browser/version: 
- Adblock extension/version:

PR #16528, opened 2023-06-30
Update Fanboy's Annoyance General/specific hide

Sonic5435325:
Possible fix for #16041.


ryanbr:
Make it specific, not generic

Sonic5435325:
I was hoping the generic filter could also help for future cases, but I'll make a new issue anyways.

PR #16549, opened 2023-07-02
A: kotitesti.fi (generic GDPR block)

peace2000:
https://www.kotitesti.fi/

Didn't see any issues, adding items to card works and checkout as well (though didn't actually buy anything).


peace2000:
Not generic enough to be a good generic, I'll make a specific.

PR #16550, opened 2023-07-02
A: kotitesti.fi (GDPR block)

peace2000:
https://www.kotitesti.fi/

Didn't see any issues, adding items to cart works and checkout as well (though didn't actually buy anything).


ryanbr:
Why not `serviceform.com/serviceform-tools/privacy/sf-privacy-partner.js`

peace2000:
Adjusted, is it ok now?

Issue #15926, opened 2023-05-15
ipaddress.com


Sonic5435325:
### List the website(s) you're having issues:

https://www.ipaddress.com/

### What happens?
Social button script
`s.ipaddress.com/shariff/shariff.complete.js`

"To the Top" button
`ipaddress.com##.btt`
![Screenshot from 2023-05-15 18-37-07](https://github.com/easylist/easylist/assets/89158249/a0b0fe4c-fb46-4338-bc28-d503f431d1a1)

Tracker
`wh.ipaddress.com`

Ad scripts
`/detroitchicago/kenai.js`
`/detroitchicago/portland.js`
`/detroitchicago/sidebarwall.js`
`/edmontonalberta/calgary.js`

### List Subscriptions you're using:

EasyList
EasyPrivacy
Fanboy's Annoyance/Social

### Your settings

- OS/version: 
- Browser/version: 
- Adblock Extension/version:

PR #16583, opened 2023-07-05
Update EasyPrivacy, and Fanboy's Social/Annoyance

Sonic5435325:
Hopefully fixes https://github.com/easylist/easylist/issues/15926.
Wasn't sure about the so called ad scripts, so I left them out.
Fanboy's Social change may not be needed, but thought it would be useful since it entirely prevents the social media buttons from ever showing up, even without cosmetic filtering. (And a possible benefit to a bit less bandwidth being used)


ryanbr:
What does each change do?

PR #16586, opened 2023-07-05
M: Remove bootstrap generic rules: bitdegree.org, ebook-searcher.com

airidasbeno:
This PR removes generic bootstrap rules from blacklist. 
pb-1 is used for setting padding bottom to 1rem.
w-100 is used for setting width to 100%.

With these rules website gets broken, and useful content to users is not properly displayed.


ryanbr:
Wholesale removal without a replacement makes this suspect on the motifations. @airidasbeno

airidasbeno:
@ryanbr I'm not sure what was intended to be removed because the targeting is very generic, almost everything on those sites stopped showing. If I would know we would arrange a replacement.

PR #16591, opened 2023-07-05
A: zippingcare.com

socram8888:
These are used by, at least, Simyo (an Spanish internet service provider owned by Orange) to track opened/read emails.

An example is `https://simyo.zippingcare.com/beacon/5ed52b2ae21b840001fd9d35.gif`

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty_international.txt
+++ b/easyprivacy/easyprivacy_thirdparty_international.txt
@@ -910,6 +910,7 @@
 ||tracker.thinkindot.com^
 ||tracking.smartmeapp.com^
 ||webpixel.smartmeapp.com^
+||zippingcare.com/beacon/$third-party,image
 ! Swedish
 ||aftonbladet.se/blogportal/view/statistics?$third-party
 ||collector.schibsted.io^


ryanbr:
How about `.zippingcare.com/beacon/$image`

socram8888:
Should it be `.zippingcare` or `||zippingcare`? Please note I'm not super savvy on the syntax of these rules, so I'm not sure what the implications would be, but I'm just asking because all other rules in the file start with `||`.

socram8888:
Ah, understood, the idea being also blocking proxied requests. My provider (Hostinger) does not proxy them, so I hadn't thought about that.

I'll edit the commit asap.

socram8888:
One minor concern - `.zippingcare.com/beacon` would get URI-encoded as `.zippingcare.com%2Fbeacon`. Should I add both `||zippingcare.com/beacon$image` and `.zippingcare.com%2Fbeacon$image`? The former would be used for RoundCube-based web mails like Hostiner which do not proxy the connections (and indeed leak the user's IP), and the later for proxied connections like Yahoo.

ryanbr:
Most urls aren't encoded. Would be a first out of all the trackers we've added. How do I signup to test this rule?

socram8888:
I don't think you can subscribe to the newsletter if you are not a subscriber of their service. I'm attaching here an exported email which you should be able to use for testing.
[simyo-mail.txt](https://github.com/easylist/easylist/files/11994870/simyo-mail.txt)

Note links also go through zippingcare. Hadn't thought about that.

PR #16612, opened 2023-07-07
A:`aniwatch.to`

eporubin:
Hi, there are popups on this domain anywhere you click, and even if the active at that time adserver it's blocked it is being changed really quickly for another one. It is very intrusive and users complained about it several times, and this seems to be a more long term solution.


ryanbr:
Adding this would break linking to the site.

Issue #16026, opened 2023-05-23
thepiratebay.org


Sonic5435325:
#### List the website(s) you're having issues with:
`https://thepiratebay.org/search.php?q=fsdgsh`

#### What happens?

To reproduce: Look up any keyword
The ad images under `torrindex.net/images/epv/` are being allowed by EasyList. These should get blocked instead, since you can see them if you don't use cosmetic filtering. Hopefully there's a way for this to be blocked in EasyList itself.

#### Filter lists you're using:
EasyList

#### Your settings:

- OS/version: 
- Browser/version: 
- Adblock extension/version:

Daxy7:
This works
`||torrindex.net/images/epv/$important`

PR #16656, opened 2023-07-11
Sonic5435325 patch 7

Sonic5435325:
To address https://github.com/easylist/easylist/issues/16026. I don't think this is needed. 
`||torrindex.net/images/ads/` blocks a "download" button ad image on `https://thepiratebay.org/description.php?id=70263353`. 
Allowing the image/script of the domain causes `||torrindex.net/images/epv/` and `||torrindex.net/static/tinysort.min.js` to not function.


Sonic5435325:
I'll update this real quick.

PR #16756, opened 2023-07-17
M: [Breakage] https://www.clickcease.com/

CbumDino:
Clickcease is being blocked by the following filter ||clickcease.com^  Clickcease is a bot mitigation and click fraud prevent domain


CbumDino:
> > Protect **your ads** and websites with the best-in-class solution
> 
> Does this break functionality on any specific website?

Yes, the domain described above
Screenshot 2023-07-17 at 17 06 53

PR #16781, opened 2023-07-20
A:`olx.bg`

eporubin:
Hi, there is a video advertisement on the [page](https://www.olx.bg/d/ad/top-podarak-digitalna-karikatura-po-snimka-izrabotka-po-vash-dizayn-CID262-ID8NmEs.html ), I know the domain it's Bulgarian, I have contacted the Bulgarian list author  but I did not get any reply over for more than a month. 
Thank you

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -10101,6 +10101,7 @@
 ##.base__header-ad
 ##.base_ad
 ##.baseboard-ad
+##.baxter-placeholder--ad--baxter-right-first 
 ##.bb-ad
 ##.bb-ad-mrec
 ##.bb-adv-160x600


ryanbr:
I'm not seeing other sites using baxter-placeholder--ad--baxter-right-first

ryanbr:
It looks generic, but if it applies to no other sites then is it really generic? Is there other sites using this element?

eporubin:
Hi @ryanbr nope there is no over sites, I would of set it as specific, but the domain is bulgarian and us I have said have tried to contact the FLA of that list without any success

Khrin:
This filter appears to have been added to the Bulgarian list. I'm closing here.

Issue #16835, opened 2023-07-25
[NSFW] allaboutcd.com


Sonic5435325:
#### List the website(s) you're having issues with:

`https://allaboutcd.com/interview-with-gorgeous-tgirl-alice/`

#### What happens?

Back to top button: 
`allaboutcd.com##.atbs-ceris-go-top`
![Screenshot from 2023-07-24 21-30-12](https://github.com/easylist/easylist/assets/89158249/74ae3cb7-fb2f-46a5-a53d-bbbbd9751f41)

Banner:
![Screenshot from 2023-07-24 21-31-31](https://github.com/easylist/easylist/assets/89158249/711e5afa-87ed-4ba5-8245-5261dc46910e)

#### Filter lists you're using:

EasyList
Fanboy's Annoyance

#### Your settings:

- OS/version: 
- Browser/version: 
- Adblock extension/version:

Sonic5435325:
@Khrin You're forgetting the annoyance.

PR #16847, opened 2023-07-25
[NSFW] fix for https://github.com/easylist/easylist/issues/16835

eporubin:
filter to hide the banner on the page


Sonic5435325:
The back to top annoyance wasn't addressed.

PR #16881, opened 2023-07-28
Remove ethyca from easyprivacy_trackingservers.txt

SteveDMurphy:
Also requested via email as well as the forums in https://forums.lanik.us/viewtopic.php?p=166322-ethyca-com#p166322
Copying the text from those here as well
***
Hello,

The following is found on the EasyPrivacy list

`||ethyca.com^$third-party`

Ethyca is the steward of the first (maybe only?) open-source data privacy platform, [Fides](https://github.com/ethyca/fides/) and some of our users recently brought this to our attention. I would like to have it removed as it is impacting users being able to effectively allow their users to control their data privacy choices.

Please let me know if I can answer any questions to help sort this out - thank you!

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers.txt
+++ b/easyprivacy/easyprivacy_trackingservers.txt
@@ -1083,7 +1083,6 @@
 ||etahub.com^$third-party
 ||ethn.io^$third-party
 ||ethnio.com^$third-party
-||ethyca.com^$third-party
 ||etp-prod.com^$third-party
 ||etracker.com^$third-party
 ||etrigue.com^$third-party


ryanbr:
Added here: https://github.com/easylist/easylist/commit/d7780a4ccd06e739a65c0d918c1f3566e5fd1afc

Due to https://cdn.ethyca.com/org/qtZ0SOEWBbViwiEWVzpukkADzQqDKH6i4M5of6Izg%3D/cookie.js (Cleaned up https://pastebin.com/ct2qADB5) 

For privacy, I see no reason to remove this. Fingerprinting, cookie (set/get), referrer, useragent checks.

Khrin:
Discussion went dead, apparently. I'm closing here.

SteveDMurphy:
> Discussion went dead, apparently. I'm closing here.

Hi @Khrin ! Apologies for the delay in responding, just coordinating with our team on next steps. Ok to reopen?

ryanbr:
Sure,, once trackers/analytics/logging are removed. Or seperate the trackers to specific sub-domain only.

cilliankieran:
@ryanbr just to check here - the law requires companies to log when consent is collected, what notice was displayed to the user, an pseudonymous identifier for the user (to prove consent) and ht date.

I'd love your feedback on how you'd solve that problem here - the law requires these functionalities and they are used specifically for this purpose. 

For reference, here's a [link to the Information Commissioner in the UK's specific guidance on requirements for the legal obligation of a data processor to record consent](https://ico.org.uk/for-organisations/uk-gdpr-guidance-and-resources/lawful-basis/consent/how-should-we-obtain-record-and-manage-consent/#:~:text=If%20consent%20was%20given%20online,full%20record%20of%20the%20conversation.).

This block is directly preventing businesses from complying with legal obligations under Article 7 of the GDPR - I don't think that's the intention here but that is the result.

NevilleS:
Hi @ryanbr, also from Ethyca here. I also want to explain what the script is doing in the source code because those browser features (cookies, user agent, etc.) are all done to block trackers and provide support for multiple browsers. I'll reference the [pastebin](https://pastebin.com/ct2qADB5) you shared and some line numbers, etc. but I'll also link to the OSS libraries we use that you're flagging issues with.

> Cookie (set/get)

We do use set a cookie called `cconsent`, e.g. on lines 6060-6064 of the pastebin:
```
                    key: "setCookie",
                    value: function(e, o) {
                        var n = new Date(Date.now() + 31536e6).toUTCString();
                        document.cookie = "cconsent=".concat(JSON.stringify(e), "; expires=").concat(n, "; path=/;"), t.default.createAuditLog(e, !1), window.CookieConsent.onSetCookieHandler && window.CookieConsent.onSetCookieHandler(e), o && o()
                    }
```

That cookie is used to store the user's consent preferences so that if they opt-out of analytics / marketing / etc. that choice is respected on future vists.

> referrer, useragent checks

I believe you're referring to lines like this in the pastebin (line 3253):
```
        const r = e.default.toObjectSet(["age", "authorization", "content-length", "content-type", "etag", "expires", "from", "host", "if-modified-since", "if-unmodified-since", "last-modified", "location", "max-forwards", "proxy-authorization", "referer", "retry-after", "user-agent"]);
```

This is the minified version of the `axios` npm library, you can see the source for that here: https://github.com/axios/axios/blob/a48a63ad823fc20e5a6a705f05f09842ca49f48c/dist/axios.js#L1483
```
  // RawAxiosHeaders whose duplicates are ignored by node
  // c.f. https://nodejs.org/api/http.html#http_message_headers
  var ignoreDuplicateOf = utils.toObjectSet(['age', 'authorization', 'content-length', 'content-type', 'etag', 'expires', 'from', 'host', 'if-modified-since', 'if-unmodified-since', 'last-modified', 'location', 'max-forwards', 'proxy-authorization', 'referer', 'retry-after', 'user-agent']);
```

Axios is used here to provide support for legacy browsers when making XHR requests. You can check out all the source for axios here: https://github.com/axios/axios

> Fingerprinting

Lastly, we don't do any fingerprinting in this script itself but we do generate a fully pseudonymous ID when storing the user's consent record. This is done to _prevent_ the ability to track the user themselves while still remaining compliant with GDPR / etc.

You can see on line 4429 of the pastebin that we use the `uuid` library (https://github.com/uuidjs/uuid) for this:
``` 
var e = require("uuid"),
            t = o(require("axios"));
```

Hopefully that explains what's going on there better, and sorry for not clarifying this earlier 👍

NevilleS:
Hi @ryanbr / @Khrin: any guidance here?

As a compromise if we really do want to block this cookie consent banner we could do this by modifying the block list to be `||cdn.ethyca.com^$third-party` to target this banner and move to one of the `easylist_cookie_*` lists instead?

PR #16897, opened 2023-07-30
P: https://www.autohero.com/fr/search/ (Fix #16854)

### LINKED ISSUE #16854: https://www.autohero.com/fr/search/  (Easylist cookie list)
I use your EasyList Cookie List via AdGuard on iOS, on this site: https://www.autohero.com/fr/search/, I have a problem with the display after the first 3-4 car ads, when I opened a ticket at AdGuard Filters, they gave me the trick to correct the problem, which is to put this rule: autohero.com#@$#body,html { overflow: auto !important; position: initial !important; }

Can you please fix the problem without a manual rule to be supplied by the user and have it corrected directly on the EsayList cookie List.

Nojuuu:
Fix #16854

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_uBO.txt
+++ b/easylist_cookie/easylist_cookie_specific_uBO.txt
@@ -2851,7 +2851,7 @@ storyhouseegmont.dk,storyhouseegmont.no,storyhouseegmont.se,egmont.com,nordiskfi
 ! autohero.com
 autohero.com##.backdrop___is1w-
 autohero.com##.root___3ffa6
-autohero.com##body,html:style(overflow: auto !important; position: initial !important;)
+autohero.com##body:style(overflow: auto !important; position: initial !important;)
 ! tallink.com
 tallink.com###cms-web-consent-root
 tallink.com##.cws-modal-open:style(overflow: auto !important; pointer-events: auto !important;)


Hayastano:
Please don't fix the problem,

Nojuuu:
Why?

Hayastano:
My problem is always like the ticket here: https://github.com/easylist/easylist/issues/16854, the site doesn't work properly I only have access to 4 ads then the site is broken, this problem is only effective with the Easylist Cookie List filter.

Nojuuu:
This should fix that so I don't quite understand why you wouldn't want it fixed?

Hayastano:
I've just updated the filter and the site has the same problem. Have you tried reproducing the problem? Does the site work correctly if you have more than 4 ads?

By adding the user rule on AdGuard iOS it works but I wanted to use my ticket to make all Easylist users benefit from it: autohero.com#@$#body,html { overflow: auto !important; position: initial !important; }

Nojuuu:
Try
```adblock
autohero.com#@$#body,html { overflow: auto !important; position: initial !important; }
autohero.com#$#body { overflow: auto !important; position: initial !important; }
```

Hayastano:
Your rule works just like the rule I posted just above. I've just tried them both and it works, thank you. What I'd like is for it to be integrated without having to add the rule to the user rules, so that all easylist users can benefit from this correction.

By adding the rule to AdGuard, the site works correctly, but if I remove it, it doesn't work correctly any more, although I think you fixed the problem on Easylist? But it doesn't seem to have worked.

Nojuuu:
It's not in the list yet, maintainers need to merge this first and after its merged I think you need to wait some amount of hours before you can see the fix on your end.

Hayastano:
Ok I was in too much of a hurry sorry, thank you very much for your time and sorry if I misspoke, I'll wait.

PR #16942, opened 2023-08-03
A:`asura.gg`

eporubin:
annoying video with lazy load on the bottom of the page:
image
Thank you


ryanbr:
Is that an advert video rather than an annoyance?

ryanbr:
Sounds good

eporubin:
@ryanbr changed as suggested to the Easy List :)

PR #16950, opened 2023-08-04
M: Fix

CbumDino:
Broken QR code at https://bancsabadell.com/cuenta-online/#scan
Screenshot 2023-08-04 at 09 53 05
Screenshot 2023-08-04 at 09 53 19


Khrin:
@CbumDino A better option would probably be something more specific like `@@||cookielaw.org^*/OtAutoBlock.js$domain=bancsabadell.com`.

PR #16953, opened 2023-08-04
A:`electricianforum.co.uk`

eporubin:
3-rd party advertising banners on top of the page:
image


ryanbr:
Why not electricianforum.co.uk##.gb-sponsored ? All 4 images are technically advertising

Daxy7:
`Advertise with us` is like 1p advertising but as an image instead of the footer. (but it's off-center)

eporubin:
I was thinking about blocking all with electricianforum.co.uk##.gb-sponsored , but had the same concern as @Daxy7 it's not redirecting to 3-rd party even if you click on it actually.

ryanbr:
I think its messey doing it this way tbh.

eporubin:
@ryanbr I can change it to `electricianforum.co.uk##.gb-sponsored`.  i agree with you just want to avoid future complains from them about blocking their self advertising.

ryanbr:
Lets change it, if they complain, we can re-address it.

eporubin:
@ryanbr sure, changed as suggested 🙌

PR #16976, opened 2023-08-07
Fix for https://thetvdb.com/

Banou26:
@ryanbr Sorry for [the previous PR](https://github.com/easylist/easylist/pull/16918), it actually broke https://thetvdb.com/, I misunderstood how the rules worked.

This PR actually does what i wanted to do originally, which is to allow 3rd parties to embed images from these sources


Banou26:
@Khrin any reasons why this got closed?
TVDB's images SHOULD be embeddable on 3rd party websites, easylist shouldn't block those requests.

PR #17002, opened 2023-08-10
A:`cnet.com`

eporubin:
Video ads on [article](https://www.cnet.com/tech/services-and-software/how-slacks-new-redesign-revamps-your-work-chats/) page:
image


ryanbr:
Need to re-review this, could cause false positives. Its in Annoyances

eporubin:
Sure didn’t know it’s blocked in Annoyances. That works as well 🤗

Issue #16461, opened 2023-06-25
howcast.com


Sonic5435325:
#### List the website(s) you're having issues with:

`https://www.howcast.com/videos/511186-how-to-do-an-ear-clap-self-defense`

#### What happens?

Ad script
`howcast.com/.bootscripts/webcomponents/adOutstream.min.js`

#### Filter lists you're using:

EasyList

#### Your settings:

- OS/version: 
- Browser/version: 
- Adblock extension/version:

PR #17032, opened 2023-08-13
Update EasyPrivacy Specific, EasyList Specific Block

Sonic5435325:
Fixes https://github.com/easylist/easylist/issues/16461.


Khrin:
is nowhere to be seen

PR #17093, opened 2023-08-18
A:`bollyflix.asia`

eporubin:
Hi, could you please add the filter to the list there is a banner on the domain: 
image
Thanks you


ryanbr:
`data-wpel-link="external"` ?

PR #17106, opened 2023-08-19
A: seutumajakka.fi (generic block)

peace2000:
https://seutumajakka.fi/artikkeli/kolumni-saako-ihminen-kadota/


ryanbr:
Probably deserves to be blocked via regional list.

peace2000:
@ryanbr Wouldn't this be a good generic?

EL already has similar rules:

```
-ad-column-
/column-ad-
/columnads/*
###AdColumn
###adColumn
###adColumn3
```

PR #17134, opened 2023-08-24
A:`fastcompany.com`

eporubin:
Annoying video on mobile mode pops up, not related to the article but produced by the domain
image


ryanbr:
Should be in Easylist @eporubin  invideo advertisiting is still advertising

eporubin:
@ryanbr changed as suggested :) Thanks

PR #17172, opened 2023-08-29
A: s.deepl.com

ZhymabekRoman:
Disable statistic in DeepL


ryanbr:
We are blocking ||deepl.com/web/statistics already, is this really needed?

ZhymabekRoman:
> We are blocking ||deepl.com/web/statistics already, is this really needed?

Yes. The Deepl Chrome extension is sending a number of requests to this domain

Khrin:
Blocking this has been known to cause problems in the past, I doubt that has changed now.

PR #17219, opened 2023-09-02
Allow 4gtv video streaming

pwengineer:
4gtv requires IP address in Taiwan.

For issue that other have reported, please see:
https://github.com/AdguardTeam/AdguardFilters/issues/33755


ryanbr:
Lets just remove it.

PR #17252, opened 2023-09-06
M:Update

eporubin:


FanOfRootBeer:
Is there a method for verifying that these are ad servers?

eporubin:
@FanOfRootBeer they matched the adservers network requests patterns and have been spotted on our checks for mobile, also we manually ensured this are not actual valid websites.

FanOfRootBeer:
thx

Issue #16062, opened 2023-05-26
scribd.com


smed79:
**privacy**

> `https://scribd.com/log/dd`
> `https://scribd.com/documents/*/pingback`

- `A: https://scribd.com/book/219884515/Arch-Linux-Fast-and-Light`

PR #17266, opened 2023-09-07
A: `scribd.com`

partingscientist:
Blocks first-party logging and ping.
A ping-back request can be seen when reading any book preview. 
An example page would be `https://www.scribd.com/read/219884515/Arch-Linux-Fast-and-Light`.
This pull request fixes https://github.com/easylist/easylist/issues/16062.


ryanbr:
Not seen

PR #17291, opened 2023-09-09
A: elisa.rekrytointi.com (GDPR)

peace2000:
Pages requires an account. (Blurred text just in case).
The main domain uses different GDPR dialog than its subdomain.

kuva

kuva


ryanbr:
A "cookie container" could be a potential issue. lets make it specific

peace2000:
@ryanbr Ok, done.

PR #17322, opened 2023-09-14
A:`247media.com.ng`

eporubin:
Hi please add these filters to the list to hide the ads on the domain there are ads on the right side of the domain 
image
and also one add embedded in the content.:
image

Thank you


ryanbr:
```
247media.com.ng##.code-block
247media.com.ng##.widget_media_image
```

How about.

PR #17324, opened 2023-09-14
M:`bailagentnetwork.com`

eporubin:
3rd party advertising banners with redirections: 
image


eporubin:
image
just the first banner from the whole Sponsored section is an add

PR #17407, opened 2023-09-26
A:`accesswdun.com`

eporubin:
Hi please add this filter to the list, there is an ad on the right side of [this page](https://accesswdun.com/article/2023/9/1206027/elberton-woman-charged-with-dui-following-saturday-night-wreck-near-hollywood):
image
thank you


eporubin:
Also on the main page there are 2 banners on top and in the content itself: 
image

ryanbr:
`#a[href*="https://tinyurl.com"]`   Is the wildecard filter needed? is a slower filter

Issue #17402, opened 2023-09-25
flamescans.org


D4niloMR:
#### List the website(s) you're having issues with:
`https://flamescans.org/1695636061-omniscient-readers-viewpoint-chapter-1/`

#### What happens?
Empty space left by ads + social buttons

#### Screenshot(s) 
![image](https://github.com/easylist/easylist/assets/70459964/80c33a8a-336c-41ef-bca1-8263b5ca67b3)
![image](https://github.com/easylist/easylist/assets/70459964/41cd0cc0-9b9e-4557-bd50-2655f2ea9df8)

#### Filter lists you're using:
Easylist annoyances

#### Your settings:

- OS/version: 
- Browser/version: 
- Ad block extension/version:

#### Other details:
```adb
flamescans.org##.code-block
flamescans.org##.socialts
```
The .code-block remove also the leftover by `radioonlinehd.net`

which could be added to EasyList as a generic rule? 
`###radio_content`

I can list a bunch of sites that has this leftover.

D4niloMR:
https://github.com/easylist/easylist/issues/15838

D4niloMR:
@Khrin 

This need to be added:
`flamescans.org##.code-block`

and what about this being added in generic rules? `###radio_content`
https://github.com/easylist/easylist/commit/5a366816447a420d7e14de6ac8dda7682adbe1d0

This commit blocks but not remove the player completely.

Issue #15838, opened 2023-05-08
void-scans.com


D4niloMR:
### List the website(s) you're having issues:
`realmscans.com` `void-scans.com`

### What happens?
Annoyance

### List Subscriptions you're using:
Fanboy's Annoyance

### Your settings

- OS/version: w10
- Browser/version: opera gx
- Adblock Extension/version: uBO 1.49.2

### Other details:
`###radio_content`

D4niloMR:
`realmscans.to##.rev-src`

![image](https://github.com/easylist/easylist/assets/70459964/5e917baa-2d4d-4b78-8172-946c6b4ba8c9)

D4niloMR:
another related (similar to `radioonlinehd.net`):

`cosmicscans.com`:

Could be added as a generic rule too:

```adb
||stream.zeno.fm^$3p
###radio-player-container
```
![image](https://github.com/easylist/easylist/assets/70459964/15ad994b-4f8e-4af3-bf63-e209b8c5785b)

D4niloMR:
> Could be added as a generic rule too:
> 
> ```
> ||stream.zeno.fm^$3p
> ###radio-player-container
> ```

@Khrin 

Would be good if you also add that in fanboy annoyance:
`||stream.zeno.fm^$3p`

With this generic rule to block and hide the player completely: `###radio-player-container`

And
`realmscans.com` `void-scans.com` 
Needs to hide `###radio_content`
(In the other issue I suggested this rule to be generic, I don't see reasons to not, since the radio is already blocked)

PR #17451, opened 2023-10-03
A: cosmicscans.com

D4niloMR:
See: https://github.com/easylist/easylist/issues/15838 and https://github.com/easylist/easylist/issues/17402


Khrin:
Definitely too general and not useful in any case.

D4niloMR:
> Definitely too general

Can be only `###radio_content.abierto` then?

PR #17466, opened 2023-10-05
A: https://www.lemonde.fr

0o5rlubm:
ie https://www.lemonde.fr/economie/article/2023/10/05/taux-d-interet-eleves-le-grand-reequilibrage-economique_6192507_3234.html

other endpoints on this subdomain returns content, so this rule only blocks tracking endpoints


Khrin:
is unclear what really is, better avoid breaking something important

PR #17491, opened 2023-10-09
whitelist zeotap

shubhamzeotap:


ryanbr:
No thanks. We've already removed the main issue filter

PR #17500, opened 2023-10-10
A:`mail.google.com`

eporubin:
Hi 🙌 could you please review and add this allowlisting filter for `gmail.com`, as for some of the users on iOS and macOS the conversion of the EL `.ads` gets applied once the user is being redirected to `mail.goog.com` causing a blank screen when trying to open and read any email.
View above the issue: 
image
Environment: 
- Safari version 17.0.
- MacOs


Yuki2718:
Where did `.ads` come from?

Yuki2718:
Sounds like an ABP issue, why not the compliance filter? Although I'm perfectly okay and positive about adding the rule.

monzta:
The compliance filter list has a different use case and isn't part of ABP for Safari.

PR #17547, opened 2023-10-15
A: github.io

salim-b:
Site to test: https://insightsengineering.github.io/teal/main/index.html

I refrained from adding `##.cookie-consent` to [easylist_cookie_general_hide.txt](https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_general_hide.txt) since I assume it hasn't already been included for a good reason like breaking scrolling on too many pages or the like. If my assumption is true, `##.cookie-consent` should be added to the [*Removed filters* table in the README](https://github.com/easylist/easylist/tree/master/easylist_cookie#removed-filters), I guess.


ryanbr:
Maybe `insightsengineering.github.io` specific, unless there is more examples used on github.io?

salim-b:
I don't have more examples from github.io at this time, so I made the change you suggested :)

PR #17551, opened 2023-10-16
A:Multiple car dealers forums

eporubin:
Right side ads on all 3 domains: 
image
Thank you


eporubin:
Changed the filter so that it wont hide the **LOG IN**  banner on the right as it's functionality of the domains

PR #17566, opened 2023-10-17
M:Update

eporubin:
Hi @ryanbr as discussed previously this list was pretty out of date and required a review, we have looked into the filters and checked if there are hits or not for each of them also updated the filter in some of the cases where it was required.  Also the list is sorted alphabetically by the domain so that at leas some sort of order exists to facilitate future contributions.  Best regards,


ryanbr:
needs a rebase

PR #17574, opened 2023-10-18
update: pushpushgo rules

worotyns:
Hi there,

I am a product owner of pushpushgo services.
There is a pull request that will update our endpoints.

Finally in this PR:
 - blockers will block our popups (js popups)
 - blockers will not block "notifications" for users that want to receive

In some words our domains:
s-eu-1.pushpushgo.com
s-us-1.pushpushgo.com
cdn.pushpushgo.com

Are resoinsible for "showing" js popup <- it will be blocked
api.pushpushgo.com/beacon - is used for tracking <- it will be blocked

all other requests like
api.pushpushgo.com/m/ - get message
api.pushpushgo.com/c/ - get message

will be not blocked and our service will work properly for our clients.

Thanks for approving it,
Regards.


ryanbr:
We're fine leaving it

PR #17652, opened 2023-10-29
formula1.com

iambrianfung:


peace2000:
Hi, unfortunately blocking the GDPR script will break videos on that site.

Sample page: https://www.formula1.com/en/latest/video.qualifying-highlights-2023-united-states-grand-prix.1780316731645839696.html

PR #17808, opened 2023-11-16
A: `mail.ru`

kyosheek:
Hides banner on mail.ru homepage and top and aside ads in mail app at e.mail.ru


ryanbr:
Non english, why not use a Russian filterlist?

kyosheek:
> Non english, why not use a Russian filterlist?

Hey! Didn't know that existed, thanks for pointing out. Will bring this to their list if it's not there, will close this PR then.

PR #17893, opened 2023-11-29
M: https://www.matkahuolto.fi/

Nojuuu:
update cookie hide


peace2000:
As the value seems to change, how about this for uBO?:

`matkahuolto.fi##+js(set-cookie, CookieConsentV2, YES)`

Nojuuu:
I'm ok with that, I can change it when I have time.

Nojuuu:
Done. Not sure if in right place, `set-cookie` section seemed bit messy in that file.

PR #17921, opened 2023-12-03
A: ampparit.com

peace2000:
Useless "We have updated your terms of service" -notification.

kuva

Better to set a cookie instead because `:has` filter for each page load is too slow. Also we don't want to block all notifications, only this specific.


Khrin:
It does not look like it is worth being blocked,

PR #17983, opened 2023-12-11
More precise ad server targeting

AkibaWolf:
Target ad server specifically without affecting all *.afishamedia.net services causing them to fail loading assets.


AkibaWolf:
Resolves #17984.

PR #17986, opened 2023-12-11
[NSFW] kemono.su and coomer.su first-party analytics fix

Linxgos:
Updated previously in #17969 and #17946
Admin keeps changing url ):


Yuki2718:
Is there any problem if we use
```
@@||kemono.su/api/v1/creators.txt
||kemono.su^$xmlhttprequest,~third-party
||coomer.su^$xmlhttprequest,~third-party
```
?

PR #17998, opened 2023-12-12
A: studocu.com

peace2000:
https://www.studocu.com/


peace2000:
@Khrin ABP doesn't need that `:-abp-` prefix nowadays (since 2022). https://blog.adblockplus.org/releases/adblock-plus-313-for-chrome-firefox-microsoft-edge-and-opera

`Added support for :has() alias for :-abp-has() to extended element hiding filters ([core#229](https://gitlab.com/eyeo/adblockplus/abc/adblockpluscore/-/issues/229)).`
`Added support for :has-text() alias for :-abp-contains() to extended element hiding filters ([core#229](https://gitlab.com/eyeo/adblockplus/abc/adblockpluscore/-/issues/229)).`

PR #18009, opened 2023-12-14
M: Update

CbumDino:


Khrin:
`||kaltura.com^` shouldn't be added.

CbumDino:
> `||kaltura.com^` shouldn't be added.
You are right. Removed !

PR #18011, opened 2023-12-14
A: https://www.metsalehti.fi/

Nojuuu:
Newsletter

image


peace2000:
@Nojuuu How about:

`metsalehti.fi##.subscribe-newsletter_wrapper`

Because `gform_wrapper` sounds like something that could be used generically to display various notifications. (Even if they are not doing it at the moment, it could change in the future).

PR #18022, opened 2023-12-16
Correctly identify PostHog

timgl:
app.posthog.com is completely broken with this rule. The only relevant tracking urls are the ones in the PR.


ryanbr:
Its been blocked due to new trackers being used. Why should we trust this.

`https://app.posthog.com/static/recorder-v2.js`

timgl:
Happy to add that script too, but that script will never load if the URLs in my PR are already blocked.

I'm a very happy ublock user, but if you completely break people's apps people will lose trust in ublock 🤷‍♂️

PR #18023, opened 2023-12-16
A: msn.com

peace2000:
Block app banner notification.

kuva

### DIFF
--- a/fanboy-addon/fanboy_notifications_specific_block.txt
+++ b/fanboy-addon/fanboy_notifications_specific_block.txt
@@ -5,6 +5,7 @@
 ||alerts.thedailystar.net^
 ||alibaba.com/sw.js
 ||annahar.com/assets/js/notifications.js
+||assets.msn.com/bundles/v1/homePage/latest/experiences_conditional-banner$script,domain=msn.com
 ||astian.org/?push_notification_sw
 ||bdstatic.com/linksubmit/push.js
 ||bigbangnews.com/sw.js


fanboy-addon/fanboy_notifications_specific_block.txt
```diff
@@ -5,6 +5,7 @@
 ||alerts.thedailystar.net^
 ||alibaba.com/sw.js
 ||annahar.com/assets/js/notifications.js
+||assets.msn.com/bundles/v1/homePage/latest/experiences_conditional-banner$script,domain=msn.com
```
Yuki2718, line 8: `$script,domain=msn.com` not needed. I wonder if `||assets.msn.com/bundles/v1/homePage/latest/libs_ad-service_dist_index_js.` can safely be blocked.

peace2000: Forgot this for a while...

Well, not technically needed but the reason I favor adding them is that the filter should perform better when it's narrowed as much as possible (even if the difference is not noticeable).

Gorhill said to me few years ago that narrowed filters are always better https://www.reddit.com/r/uBlockOrigin/comments/p6d6np/comment/h9flegz/?context=3

kuva

Gorhill isn't at Reddit anymore but he wrote that comment to me.

@Yuki2718

Yuki2718: This comes down to https://github.com/uBlockOrigin/uBlock-discussions/discussions/816 but anyway EL prefers simpler form.

Issue #18063, opened 2023-12-20
paste.fo Filters Destructive


verybigwave:
#### List the website(s) you're having issues with:
paste.fo

#### What happens?
The easylist filter added by https://github.com/easylist/easylist/commit/442c0b64dc9c5ffa9d468810b70c41225c47a96f is destructive by completely removing external links. It for example removes some of the users contacts linked on their profiles, and not just the ads it should remove.

#### Screenshot(s)
The destructive behavior includes breaking these contact fields among other links not really working when being clicked:
<img src="https://i.ibb.co/nkB4sJw/img.png">

#### Filter lists you're using:
UBO with EasyList

#### Your settings:

- OS/version: Windows 11
- Browser/version:  Firefox Latest
- Ad block extension/version:  UBO Latest

#### Other details:
The filter for all target="_blank" anchor tags added in https://github.com/easylist/easylist/commit/442c0b64dc9c5ffa9d468810b70c41225c47a96f

Daxy7:
For the ads:
```
paste.fo##.easylist_is_shit_fuck_huyszn
! invalid
paste.fo##.texta
paste.fo##.wrapper a[href][target="_blank"] 
```

PR #18134, opened 2023-12-29
M: paste.fo

D4niloMR:
Fix https://github.com/easylist/easylist/issues/18063


D4niloMR:
The current filter is hiding the contacts and these two:
image

`https://paste.fo/0486216b70c2`

PR #18257, opened 2024-01-11
M: fix typo

huyszn:
Also, is there an easier way to subscribe to the agegate list on uBO?


Yuki2718:
>  is there an easier way to subscribe to the agegate list on uBO?

@ryanbr

PR #18276, opened 2024-01-12
A: exception for doubleclick redirects from finder.com.au and finder.com

fmrk:
**Adding Exception for Finder Domains**
This pull request proposes an update to the existing `doubleclick.net` filter exception rule to include [Finder](finder.com.au) domains `finder.com.au`, `finder.com` and `creditcard.com.au`.
This change addresses the issue where ad blockers were inadvertently closing user-intentionally opened tabs due to redirects involving `ad.doubleclick.net`.
The issue has been first detected while using uBlock Origin and discussed also here: https://github.com/uBlockOrigin/uAssets/issues/21939

### DIFF
--- a/easylist/easylist_allowlist_popup.txt
+++ b/easylist/easylist_allowlist_popup.txt
@@ -2,7 +2,7 @@
 @@|data:text^$popup,domain=box.com|clker.com|labcorp.com
 @@||accounts.google.com^$popup
 @@||ad.doubleclick.net/clk*&destinationURL=$popup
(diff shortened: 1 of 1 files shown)


mopielka (commented):
I support adding finder.com.au to the list of exceptions. On our site, the redirection process to a partner's site is a deliberate action by the user. It requires the user to click a 'Go to site' button, after which they are presented with a redirect page before being taken to the partner’s site. This process is intentional and cannot occur by accident. Therefore, maintaining this behavior is crucial, as it aligns with the intended user experience on finder.com.au. Implementing a mechanism that interrupts this process would be counterproductive to the site's functionality and user engagement.

PR #18321, opened 2024-01-17
More specific targetting for aria-live

FluidSense:
The current targetting on vy.no, div[aria-live="polite"] also hides critical functionality, like passenger amounts, etc. Targetting all div that has accessibility labeling results in a hard-to-use website for non-screenreaders

| With EasyList | Default |
|---|---|
|image|image|

Did I assume correctly that it is CSS selectors being used? So a [descendant selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) can be more precise targetting? 

The cookie banner that should be targetted is the blue div here:
image

### DIFF
--- a/easylist_cookie/easylist_cookie_international_specific_hide.txt
+++ b/easylist_cookie/easylist_cookie_international_specific_hide.txt
@@ -1671,7 +1671,7 @@ helthjem.no##.schibsted-bar
 nordiskporselen.com##.w3-black
 nitedals.no##.wf-section
 direktesport.no##[aria-label="Informasjonskapsler"]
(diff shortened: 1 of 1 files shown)


ryanbr:
`##div[data-theme="dark"][aria-live="polite"]` how about that

FluidSense:
> `##div[data-theme="dark"][aria-live="polite"]` how about that

I believe that would target everything with a dark-theme. But it has an id now, so that can be targeted directly!

ryanbr:
Assuming it doesn't change the div id, we'll remove it.

Issue #18430, opened 2024-02-01
bot-hosting.net


MathiasDPX:
![ad](https://github.com/easylist/easylist/assets/56231137/76ba87e6-25ec-40f8-97cf-b38f6ff6ca4a)

Ad popup
[https://bot-hosting.net/panel/earn](https://bot-hosting.net/panel/earn)

Solution: bot-hosting.net##div#adModal
I put this in "My filter" in UBO and it work

Yuki2718:
I can't reproduce with logged in.

PR #18463, opened 2024-02-04
A: bot-hosting.net

MathiasDPX:


Yuki2718:
No explanation and no link? Anyway, wrong syntax.

Khrin:
> No explanation and no link? Anyway, wrong syntax.

https://github.com/easylist/easylist/issues/18430 I think

MathiasDPX:
Can I get an explication please,
This is my first filter

Yuki2718:
> Can I get an explication please, This is my first filter

`||` is domain anchor and should not be added to element-hiding rules. Read https://help.eyeo.com/en/adblockplus/how-to-write-filters and https://adblockplus.org/en/filter-cheatsheet

MathiasDPX:
Ooh ok I read everything you send to me,
Si If I remove the domain anchor everything should bé done ?

Yuki2718:
> Si If I remove the domain anchor everything should be done ?

No, as I cannot reproduce the original issue https://github.com/easylist/easylist/issues/18430. We do not add rules just because someone claims the rules are useful. The story could be different if you're one of the long-term contributor, but even if so we prefer having evidences that the rules are needed AND they don't cause trouble. We need at least screenshot(s) of DOM around the ad element which will be shown in devtools.

Issue #8268, opened 2021-07-02
Uber Eats tracker?


c2xusnpq6:
https://www.ubereats.com/ --> https://www.ubereats.com/ramendca/events/ack?seq=0 tracker?
https://www.ubereats.com/ --> https://www.ubereats.com/_events

```
https://www.ubereats.com/ramendca/events/ack?seq=0

||ubereats.com/ramendca/events^
||ubereats.com/*_events^
||uber.com/*_events^
```

ryanbr:
Resolved.

PR #18560, opened 2024-02-16
Update easyprivacy_specific.txt

dev-bun:
Add Medal.tv Amplitude - they have 2, but only one as listed here.


Daxy7:
why is ubereats event trackers being removed? related https://github.com/easylist/easylist/issues/8268

PR #18579, opened 2024-02-18
A: `kompas.com`

partingscientist:
The domain `tracker.oval.id` is used to track page analytics. At the very least, it tracks user id, event id, user agent, time zone, time stamp, user action, and input count.


Yuki2718:
How about `||oval.id/tracker/`?

monzta:
@Yuki2718 I believe blocking it via the sub-domain it easier/better for DNS blocking.

Yuki2718:
Generally agree, but `oval.id/tracker/` is the initiator here.

PR #18582, opened 2024-02-19
A: i.posthog.com

frankh:
PostHog will be adding some new capture endpoints to replace the existing `app.posthog.com` ones that are blocked

We don't want this change to impact adblock users so we want to preemptively add the new domain (anything under `*.i.posthog.com` should be blocked)


frankh:
not sure what the process is for getting PRs merged here, @Khrin @ryanbr could you merge this as we're rolling it out today

PR #18612, opened 2024-02-21
A: https://sadisflix.lat/

smed79:


Yuki2718:
Oh, now ABP supports domain wildcard? But why `#?#`?

PR #18649, opened 2024-02-27
A: https://send.cm/d/wecD

smed79:


Khrin:
There does not seem to be any need for it at the moment.

PR #18745, opened 2024-03-14
A: `upstract.com`

partingscientist:
The endpoint `/tokei/do/ev` is used to receive page-view analytics. At the very least, the endpoint logs page-view events, page referrer, and page href.


Yuki2718:
Maybe better to block parent `https://upstract.com/tokei/l.js`, but does this break something?

PR #18746, opened 2024-03-14
A: `kayak.co.id`

partingscientist:
The endpoint `/log/client/messages` is used to receive error logs. At the very least, the endpoint logs user agent, view port size, adblocker status, and error stack trace. To view the connection, one must wait about a minute after the page has fully loaded.


Yuki2718:
Can you make it generic `/log/client/messages`? We want to minimize use of wildcard in rules.

PR #18781, opened 2024-03-19
Update easyprivacy_trackingservers_thirdparty.txt

philipp-tailor:
Removes birdeatsbug.com, which is neither related to ads, nor tracking users.

Disclaimer: I work for https://birdeatsbug.com.

I'm happy to provide any information needed to convince that the service is unserverdly on an ad blocker list.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
+++ b/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
@@ -312,7 +312,6 @@
 ||bigreal.org^$third-party
 ||bigtracker.com^$third-party
 ||bionicclick.com^$third-party
-||birdeatsbug.com^$third-party
 ||bizible.com^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
There is many tracking bits in `https://sdk.birdeatsbug.com/v2/core.js`  

See what we cover/block: https://github.com/easylist/easylist/blob/master/README.md#easyprivacy  it should cover this script

philipp-tailor:
Thank you for pointing me towards the policy. Is there any specific web technologies that determine that a script is classified as containing "tracking" elements?

For some context: Our script is a bit different than most monitoring solutions in that it does not send any data back to our server automatically. Data is only sent once a user explicitly presses a button. The script currently contains some `XMLHttpRequest` / `fetch`es, which make it impossible for me to proof that no user data is sent back to our servers automatically. Would removing all XHR / fetch requests remove the concerns?

PR #19078, opened 2024-04-29
A: https://l.tapdb.net, https://ad.tapdb.net.

kocomic:
The main domain tapdb.net is not an advertising site, but the two subdomains l.tapdb.net and ad.tapdb.net are. Blocking the main domain will affect the display of static resources (static.tapdb.net) on the data tool website (tapdb.com).

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -1849,6 +1849,7 @@
 ||adzpier.com^
 ||adzpower.com^
 ||adzs.com^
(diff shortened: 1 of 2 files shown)


ryanbr:
```
static.tapdb.net 404 
www.tapdb.net’s server IP address could not be found.
tapdb.net took too long to respond.
```

I'm fine leaving this, 2 rules vs many, and to the outside user it won't matter.

PR #19128, opened 2024-05-06
M: Update

avanikaein:


ryanbr:
Double check these aren't legit domains @avanikaein 

https://github.com/uBlockOrigin/uAssets/issues/23600

avanikaein:
Thanks for the reminder & catching my mistake @ryanbr

PR #19163, opened 2024-05-13
Print the uBO pre-parsing include directive as originally read without adding a space between ! and #

### LINKED ISSUE #19162: Fanboys Social doesn't actually work on uBO
Despite having uBO includes configured, the list isn't properly parsed resulting in 0 out of 0 filters used notice.

this goes back a couple of years https://github.com/uBlockOrigin/uBlock-issues/issues/2191

This is due to an erroneous space in the preparser

`! #include fanboy_social_specific_uBO.txt` should be `!#include fanboy_social_specific_uBO.txt` at line 12954 of fanboy-social.txt

Squall-Leonhart:
This *should* ensure the line is written correctly without adding a space, but i'm not very python literate.

if it behaves as expected, should close #19162

### DIFF
--- a/FOP.py
+++ b/FOP.py
@@ -214,10 +214,6 @@ def fopsort (filename):
                         section = []
                         lineschecked = 1
                         filterlines = elementlines = 0
-                    # Check for uBO specific lines and write them as is
-                    elif line.startswith("! Include ubO specific") and (line + "\n").startswith("!#include"):
-                        outputfile.write(line + "\n")  # Write both lines
-                    else:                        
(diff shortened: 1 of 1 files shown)


Squall-Leonhart:
meh, didn't work, i'll try another approach.

Squall-Leonhart:
pfft, i just realised this project doesn't do actions on PR's, and monetz latest commit triggered one at seemingly the same time i made this confusing me, so now i'm not sure if the original way i tried handling it wasn't working.

Khrin:
not needed

PR #19228, opened 2024-05-25
P: kd.fi (GDPR)

peace2000:
Partially blocked, messes with the text in the page content.

https://www.kd.fi/tapio-puolimatka/

kuva


Yuki2718:
Really? Though the piceker highlights such a wide area, I don't see actual text being affected.

peace2000:
kuva

In this picture, the content text and both the element are highlighted and they overlap clearly.

Anyway, overlap or not, `##.wt-cli-cookie-bar-container` is still visible thus should be hid.

@Yuki2718

Yuki2718:
Oh, I thought you removed an entry as it's `P:`.

PR #19239, opened 2024-05-28
Fix Cookie Clicker addon being blocked by cookie notice list

qtlunya:
https://github.com/CookieMonsterTeam/CookieMonster

https://orteil.dashnet.org/cookieclicker/


ryanbr:
Works fine here.

qtlunya:
Do you mean it isn't being blocked for you in the first place, or my fix works fine?

I'm using uBlock Origin on Firefox and loading Cookie Monster using Tampermonkey, if that makes a difference.

Issue #19252, opened 2024-05-30
Remove houston.advgo.net in easyprivacy_thirdparty_international.txt file


charlyliu:
#### List the website(s) you're having issues with:

||houston.advgo.net^ in https://github.com/easylist/easylist/blob/0d2a0504abc5ea99afb54b0a9da49fe5a47c1e04/easyprivacy/easyprivacy_thirdparty_international.txt#L204

#### What happens?

This line is incorrect since this base url is to download a configuration and not sending any tracking data. Please remove this line.
This is hurting the feature flags used in our sites.

PR #19253, opened 2024-05-30
Remove houston.advgo.net

charlyliu:
This address is to download a feature flags and ab tests configuration, it is not used for any tracking system. For tracking data, if sent, it will be sent to the tracking providers.

That's why I would like to remove this address from the block list.

This is related with the issue https://github.com/easylist/easylist/issues/19252


ryanbr:
https://github.com/easylist/easylist/issues/17926

It could've been a wider block here. Seems fine to block  `it will be sent to the tracking providers.`


https://github.com/easylist/easylist/blob/master/README.md#easyprivacy

PR #19272, opened 2024-06-03
Update easyprivacy/easyprivacy_trackingservers_general.txt

filipealc:
Removes `vturb.com.br` and `vturb.net` which is not related to ads neither tracking users.

Disclaimer: I work at `www.vturb.com`.

I would like to explain a bit for why these endpoints are used and why the website `https://comoconquistarumamulher.com.br/` was using it and our relation to it.

We are a company that provides a platform for users to host their videos and usually use them in their sales pages. 

### About the `vturb.com.br`

The endpoint `vturb.com.br` which at the website was calling a subdomain `api.vturb.com.br` is used to check if the video itself has a valid license from our platform, it's not used to tracking anything related to the user neither do we collect any user information.

### About the `vturb.net`

This endpoint is used to receive analytics data from the video itself, we don't track any user personal information, we store things related to the performance of the video like the velocity the segments are being delivered, if the video A or B was showed, which configuration the video had at the time of the play and other video metrics but not with any user information. We do store at the localstorage things that would allow the user to refresh the page and continue where he left in the video which is useful for clients that provide masterclasses or educational content.

### Our policy for our clients that use our platform

Here is our terms of use in EN: https://vturb.com/en/terms
Here is our privacy in EN: https://vturb.com/en/privacy

Those two are the terms of agreement our clients comply when joining our platform.

We appreciate any pointers on how we can achieve the removal of our domain are we are happy to clarify anything that was not made clear.

The commit that introduced our domains as blocked was this one here: https://github.com/easylist/easylist/commit/daaad642f3825527be1ad21d2c34f9bee325e957 so I will tag @Khrin here. 

Thanks in advance.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_general.txt
+++ b/easyprivacy/easyprivacy_trackingservers_general.txt
@@ -362,8 +362,6 @@
 ||vkanalytics.net^
 ||vpdcp.com^
 ||vstats.me^
-||vturb.com.br^
-||vturb.net^
 ||wct-2.com^
 ||whale3.io^
 ||widgetbe.com^


ryanbr:
```
https://a.vturb.net/y  (ping)
https://a.vturb.net/x (xmlhttp)
```
Seems more than "valid checks".

filipealc:
> ```
> https://a.vturb.net/y  (ping)
> https://a.vturb.net/x (xmlhttp)
> ```
> 
> Seems more than "valid checks".

Hey @ryanbr, thanks for sharing! So, the “valid checks” are at vturb.com.br this is where we validate if a client of ours can have their videos running, they need to have their subscription with us up to date, this is the endpoint that we also use when a client is logging into our platform. 

Regarding the /x and /y at vturb.net we already use a different domain to make it explicit that this is where we are collecting metrics of our videos to our clients. If you look at what’s being received by /X you can see things like, the video was paused, clicked, time is running and thus this provides an analytics for our client about the video performance.
 
 
If this is considered a violation then we would gladly provide more info and/or try to adapt to a better rule where we can still provide video metrics to our clients and still respect user anonymity as we always try to do. 

Thanks in advance, please feel free to ask for any additional info.

ryanbr:
Easyprivacy does block metrics, eventlogs etc, see: https://github.com/easylist/easylist/blob/master/README.md#easyprivacy

Which specific urls are used for only non-metrics/tracking?

filipealc:
@ryanbr the specific url that checks the license if a customer can have the video player is: `https://api.vturb.com.br/vturb/check` , I will complement saying that this domain `api.vturb.com.br` doesn't do any sort of metric neither eventlogs etc.

ryanbr:
In middle of travel, this is the best middle ground for the time being

PR #19388, opened 2024-06-19
A: `arbys.com`

partingscientist:
Hides social icons.


Yuki2718:
@ryanbr I guess we should create `:has` section for FS/FA as well.

Yuki2718:
@partingscientist `#?#.linkBlocks_linkBlock__o6P56:-abp-has(> .linkBlocks_socialLinkBlockList__1foiF)`

Yuki2718:
We usually don't care as seen e.g. in `weather.com##.Footer--socialContainer--3z8CW` but that's fine.

PR #19389, opened 2024-06-19
A: `news.almdrrj.com`

partingscientist:
Hides ad banner leftover.


Khrin:
For me it redirects straight to `www.almdrrj.com`, which seems to be in Arabic only.

@smed79

smed79:
i am redirected `www.almdrrj.com` without seeing any ad.

PR #19439, opened 2024-06-29
A: primevideo.com (GDPR dialog)

peace2000:
https://www.primevideo.com/


Yuki2718:
FOP error, create a comment line for `:has` like easylist_specific_hide.txt and put the rule under that.

PR #19449, opened 2024-07-02
Update cnames

JobcenterTycoon:


Yuki2718:
Those commented out filters should be excluded such as `stats.suenicholls.com` in plausible.txt

JobcenterTycoon:
I did for all other lists already, i only missed the 2 filter in plausible.txt

PR #19478, opened 2024-07-07
M:goodreturns.in

mediumkreation:
found in url: 
https://www.goodreturns.in/classroom/maximize-your-tax-deductions-how-credit-card-expenditures-can-boost-your-itr-filing-this-fiscal-1355957.html

image


Yuki2718:
Typo `goodreturns.in####[class*="oiad"]` and it should not be in taboola section. But is the change really needed? I don't see placeholder or ad newly covered by the change other than `##div[id^="taboola-"]`. Even if needed, it looks `##[class^="oiad"]` is enough.

PR #19487, opened 2024-07-07
P: justwatch.com

peace2000:
These need to be allowed, otherwise you cannot click link to the Disneyplus streaming service.

https://www.justwatch.com/us/tv-show/naruto

kuva


Khrin:
Not needed.

PR #19488, opened 2024-07-08
A: https://www.shoro7atnadir.com/

smed79:


Khrin:
@smed79 I think they both can be safely listed in https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_trackingservers_general.txt

smed79:
https://github.com/easylist/easylist/pull/19489

PR #19513, opened 2024-07-10
A: frequentmiler.com

avanikaein:
Example: https://frequentmiler.com/stop-putting-spend-on-your-delta-card/

body element of page is given 'adhesion' classname after scrolling, which results in a blank page.

Screenshot 2024-07-10 at 12 05 04 PM

### DIFF
--- a/easylist/easylist_general_hide.txt
+++ b/easylist/easylist_general_hide.txt
@@ -7017,9 +7017,9 @@
 ##.adheightpromo
 ##.adheighttall
 ##.adherebox
-##.adhesion
 ##.adhesion-block
 ##.adhesion-header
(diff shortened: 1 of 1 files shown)


Khrin:
Cannot confirm this, there are no hits for this filter (and the page is displayed correctly).

avanikaein:
thanks for taking a look @Khrin 
On my end, the page goes blank after the initial scroll. It loads fine initially and then the classname "adhesion" is added to the body element. Attaching a screen recording.
let me know if there's other info I can provide

https://github.com/easylist/easylist/assets/56893063/fc0e091f-c68b-4748-a7b1-b63019b605d7

Issue #19533, opened 2024-07-12
Audioscrobbler is not an annoyance


XanderStrike:
#### List the website(s) you're having issues with:

The [Web Scrobbler](https://web-scrobbler.com/) extension.

#### What happens?

Scrobbling of songs (aka reporting that they were played) to last.fm is broken.

Related issue: https://github.com/web-scrobbler/web-scrobbler/issues/4783

#### Filter lists you're using:

Fanboy's Annoyances List

#### Your settings:

- OS/version: MacOS Sonoma 14.5
- Browser/version: Chrome 126.0.6478.127 (Official Build) (x86_64)
- Ad block extension/version: uBlock Origin Lite 2024.7.3.674

PR #19534, opened 2024-07-12
Remove audioscrobbler from fanboy annoyances

XanderStrike:
This is not an annoyance, it's blocks desired functionality for people who use last.fm

Fixes https://github.com/easylist/easylist/issues/19533


ryanbr:
Its targeting the embedded widget, happy to modify the block to cover this. But a general removal isn't needed

PR #19545, opened 2024-07-14
A: `speed.cloudflare.com`

partingscientist:
Blocks event logging and hides consent banner.


Khrin:
Better not to block this alert, I think.

PR #19550, opened 2024-07-14
A: https://iptvgratuit.net/

smed79:


Khrin:
It would probably fit better in  `easyprivacy_specific_international.txt`.

PR #19575, opened 2024-07-17
Fix 220.lv website breakage (easylist_cookie_specific_uBO.txt)

erba2061:


Khrin:
Not needed.

erba2061:
Website breaks with current rule. Should I file an issue instead? @Khrin

ryanbr:
Does this help @erba2061

erba2061:
Yes, all websites work as expected. Thanks 👍

PR #19580, opened 2024-07-18
A: https://www.easyschrank.ch

salim-b:


Khrin:
@salim-b Please add it in https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_international_specific_hide.txt

salim-b:
Sorry, I didn't notice the international file. See https://github.com/easylist/easylist/pull/19588.

PR #19626, opened 2024-07-22
reverting tinybird.co

javisantana:
Hello, Tinybird.co founder here.

We are not a tracking company, you can use [our API](https://www.tinybird.co/docs/ingest/events-api) to send events from anywhere (and in this case is being used to get analytics events from a web) but our service is not related to tracking users, ads or anything like that, we are more a database on steroids platform.

Thanks!


Khrin:
@javisantana  Read`https://easylist.to/pages/policy.html`

PR #19692, opened 2024-07-30
fix: remove wildcard end from fingerprint.js filter

jguddas:
This resolves https://github.com/lucide-icons/lucide/issues/1675#issuecomment-2147119821 where the fingerprint icon failed to load.


Khrin:
This would reduce the effectiveness of the filter and is simply not an option.

PR #19693, opened 2024-07-30
Address some tracking

JobcenterTycoon:
Checked and copied from the uBO privacy list


Khrin:
Too much stuff to check, please open a PR for each filter.

PR #19709, opened 2024-08-01
update tarnkappe.info tracking

JobcenterTycoon:
The letters after `mat` are random.


Khrin:
In case you haven't noticed, pings are already blocked

JobcenterTycoon:
Why not block the download and execution of the script?

Khrin:
They would "fix" it quickly. No one has time to waste on this cat-and-mouse game and the end user would see little to no benefit from it.

Feel free to add in the "privacy" list of Ublock Origin.

JobcenterTycoon:
I ask by general because also other pulls like https://github.com/easylist/easylist/pull/19715 got declined.

PR #19710, opened 2024-08-01
address huya.com tracking

JobcenterTycoon:
https://github.com/uBlockOrigin/uAssets/pull/10610

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -427,6 +427,7 @@
 ||dubz.live/analytics/
 ||duckduckgo.com/t/
 ||dumpster.cam4.com^
+||e-stat.huya.com^
 ||e.quizlet.com^
 ||e.used.ca/event
 ||e.zg-api.com/event$domain=zillow.com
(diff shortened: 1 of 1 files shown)


Khrin:
Why do many users in this discussion seem to be suggesting that the rules should actually be badfiltered?

@Yuki2718 @piquark6046 do you both know the site? thoughts on it?

Yuki2718:
I don't, but it looks two user experienced loading issue which OP couldn't reproduce.

JobcenterTycoon:
But only with `||p2p.huya.com^`? The filter is disabled at the moment.

Yuki2718:
Oh, right.

Yuki2718:
But it is classified as resource-abuse, so maybe more suited to easyprivacy_trackingservers_mining.txt though are not exactly miner.

Khrin:
> But it is classified as resource-abuse, so maybe more suited to easyprivacy_trackingservers_mining.txt though are not exactly miner.

Yes, it is definitely out of the scope of the list.

Khrin:
Also, I have never seen an instance of `||e-stat.huya.com^`, so I think it can be closed.

PR #19711, opened 2024-08-01
address https://github.com/uBlockOrigin/uAssets/issues/22676

JobcenterTycoon:
https://github.com/uBlockOrigin/uAssets/issues/22676


Khrin:
An example site to see where it's used?

PR #19716, opened 2024-08-01
address t3n.de tracking

JobcenterTycoon:


JobcenterTycoon:
Wait the request coming from `https://storage.googleapis.com/t3n-de/assets/t3n/2018/scripts/msodrq.js` looks like a tracking only script

PR #19809, opened 2024-08-14
Update easylist_specific_hide.txt for edhrec.com 2024-08 redesign

jnovack:
Add edhrec.com `has` elements for top of site ads and mid-scroll-loading ads on content pages.

Without:
image

With:
image


Khrin:
Not needed

PR #19823, opened 2024-08-16
Remove featuregates.org

phlickey:
FeatureGates.org is a domain used by [statsig](https://www.statsig.com/), a feature flag provider. This domain provides end user functionality, and blocking it can potentially breaks sites that rely on this service for feature flagging.


Khrin:
No broken sites seen so far. Unless someone does it on purpose...

PR #19865, opened 2024-08-22
A: adproxy.tf1.fr

Slowsby:
Add tf1.fr pre-roll and mid-roll ad server.


Khrin:
@Slowsby Please ask here: https://github.com/easylist/listefr

PR #19915, opened 2024-08-30
Allow Snap Inc owned websites to load static assets from snap.com

mctrafik:
For unknown to us reason, `snap.com` is blocked in easylist. This is the main domain for "Snap Inc" and hosts many domains, endpoints, services and static files. 

This rule breaks many of our sites by not loading fonts or allowing us to check if the user should see the cookie consent popup.

As far as we can tell, this rule does not help anyone on the web because we don't serve ads, popups or annoying content from this domain. Let us know why this entry is on easylist, so we can move just that to a different domain. Blocking an entire company's domain feels like overkill.

Here's our previous attempt at unbreaking our sites: https://github.com/easylist/easylist/pull/18537 which didn't get any comments on it.

This PR adds more of Snap Inc's domains to what I presume is the allowlist where this rule doesn't apply.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -38142,7 +38142,7 @@
 ||smowtion.com^$third-party
 ||smpgfx.com^$third-party
 ||snack-media.com^$third-party
-||snap.com^$third-party,domain=~forbusiness.snapchat.com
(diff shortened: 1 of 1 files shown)


ryanbr:
Will be re-added if we get any reports of this domain hosting ads on any sites. The only warning.

mctrafik:
Thank you!

PR #19973, opened 2024-09-07
Hide some cookie banners

qwerty287:
Hello 👋

I fixed some cookie notices that I found during the last days or in some issues.

If I should contribute this in another way, please just comment and tell me what I should do.


ryanbr:
```
berliner-missionswerk.de###cmmodal
siebenbuerger.de###cookiehinweiscontainer
k-mart.ro##.ckPolicyGomag
```

removing the www. bits

PR #20047, opened 2024-09-19
Add rule for 2fa authentication prompt website of facebook

AdilArslannn:
After signing in to an account with 2FA enabled, you will be directed to the 2FA prompt page, where a cookie consent form with different attributes is used.


ryanbr:
Seems insecure to ignore 2fA warnings. more of a personal filter option here.

PR #20048, opened 2024-09-19
Add rule for instagram cookie consent form

AdilArslannn:
Added a new rule for hiding cookie consent form on instagram.com


ryanbr:
too risky too add with a cosmetic

PR #20051, opened 2024-09-19
Fix wsj.com

chrmod:
Screenshot 2024-09-19 at 15 25 48


Yuki2718:
url?

stephenhawk8054:
Related: https://github.com/uBlockOrigin/uAssets/issues/25371 . I can reproduce the issue.

chrmod:
The url is https://www.wsj.com/ - there is an infinite loop caused by two resources re-trying to be downloaded. 

We've already shipped a fixed to Ghostery https://github.com/ghostery/broken-page-reports/commit/9f85ef72171a7d01e2404a3eeda6d4aebac9a197

PR #20067, opened 2024-09-24
Add google.com rule for cookie consent hiding

AdilArslannn:
Hey, I created this rule to match and hide the cookie consent form on google. I tested it and it works as it intended.

`##div#xe7COe[role="dialog"]:-abp-has(a[href^="https://policies.google.com/technologies/cookies"])`

Here is the screenshot of the elements:
[screenshot](https://imgur.com/a/IeIwNVq)


ryanbr:
Hiding this will cause false positives

PR #20069, opened 2024-09-24
Add rule for cookie consent hiding for theguardian.com

AdilArslannn:
Hey, another rule, this one is for theguardian.com . I tested it and works fine.

The rule:
`theguardian.com##div[style*="height: 100%"]:-abp-has(iframe[title="The Guardian consent message"] > div.message-container)`

Screenshot of the element:
[screenshot](https://imgur.com/a/HmaSZhF)


ryanbr:
would break, its addressed in uBO with a trusted click

Issue #19401, opened 2024-06-21
api.sardine.ai: tracking


somehundred:
#### List the website(s) you're having issues with:
api.sardine.ai

#### What happens?
api.sardine.ai requests are found at `https://stockx.com/?nrtv_cid=.nrtv_plchldr.&ar=1845742109423608737&impactSiteId=xRM2Pu24%3AxyKRpuRnMXYg3DRUkHSPCW5hQDo1w0&clickid=xRM2Pu24%3AxyKRpuRnMXYg3DRUkHSPCW5hQDo1w0`

sardine.ai is used for tracking user, recording user's actions. This domain opens websocket in site I linked and keeps it open possibly sending and logging user actions. More info here: `https://sardine.ai`

According to homepage,
**"Capture deep device intelligence and behavior biometrics for every customer touchpoint and interaction on your website or app."**

**"Behavior Biometrics
Typing speed, mouse movement, scrolling and swiping patterns, hesitation and distraction events, stressed behavior, and context switching.**

**Device Intelligence
Device and browser fingerprints, True IP and location, emulators, tampered apps, remote access tools (RATs), Proxy piercing, VPN detection, bots, and device farms."**

#### Filter lists you're using:
Adguard filter lists, except social and widgets filters.

#### Your settings:

- OS/version: windows 11
- Browser/version: 
- Ad block extension/version:

PR #20133, opened 2024-10-08
Isolates sardine.ai tracking script to 3P only

kdenhartog:
This makes it such that the script is available for loading when on a first party site such as https://crypto.sardine.ai/buy where the page breaks if this script doesn't load.

In this case, it's 1P fingerprinting occuring, but in all other contexts this script gets loaded as a 3P tracker and should be blocked.

Fixes https://github.com/easylist/easylist/issues/19401 and follows on from https://github.com/easylist/easylist/commit/ff59ff5f0b6009bfd0881669fa9268ed044388d5 which was meant to fix it, but introduced a webcompat issue on https://crypto.sardine.ai/buy

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -246,7 +246,7 @@
 ||api.collarity.com/cws/*http
 ||api.country.is^$third-party
 ||api.iris.tv/update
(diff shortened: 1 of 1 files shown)


ryanbr:
`^$third-party` -> `$third-party`

kdenhartog:
> `^$third-party` -> `$third-party`

Fixed thanks

PR #20201, opened 2024-10-17
address bolighub.dk dataunlocker

JobcenterTycoon:


Yuki2718:
This site also uses server-side GTM that bypasses current filters.

JobcenterTycoon:
What do you mean with this?

JobcenterTycoon:
But how to they send the data? I see only 3 xhr but with no data send.

Yuki2718:
Hm, you're right. Maybe this site doesn't engage in analytics with GTM.

PR #20217, opened 2024-10-19
Move filters to general hide

Tiagoquix:
Move some `##app-cookie` filters to the `general_hide.txt` file. Remove those from `specific_hide.txt`.

`##app-cookie-popup` is already included in general_hide.


Khrin:
No apparent need for it at this time.

PR #20263, opened 2024-10-28
Update Shield

JobcenterTycoon:


Yuki2718:
@JobcenterTycoon Wrong position

PR #20756, opened 2024-11-29
Add Google Drive popup to notifications list

UsernamesAreNotMyThing:
"Get notifications on your desktop for shared files and important events" message


ryanbr:
Way to generic text naming, also avoid upward()

PR #20797, opened 2024-12-08
A:moranews.com

mediumkreation:
**found in url:** http://moranews.com/ , 
https://moranews.com/serap-aspirasi-masyarakattumpal-pasaribu-gelar-reses/

image
image


monzta:
The language seems to be Indonesian. Please report this to https://github.com/ABPindo/indonesianadblockrules in case it's not blocked there yet.

PR #20875, opened 2024-12-20
A: https://www.nslookup.io/

smed79:


monzta:
@smed79 seems to be a tracking script. I think it should be in EP instead

smed79:
added here https://github.com/easylist/easylist/commit/79c36274eab7384cb046480413d6eddcd7fc3582

Thank you.

PR #20894, opened 2024-12-23
M: https://wolt.com/

user9517:
fixed wolt.com sponsored hiding.

Instead of showing a blank div, it removes the div and shows only non-sponsored divs.

before:
image
after:
image


ryanbr:
https://wolt.com/en/deu/bielefeld
https://wolt.com/en/deu/bielefeld/restaurant/mcdonalds-bielefeld-am-bahnhof

looks fine without the patch here.

PR #20972, opened 2025-01-03
A: https://www.g2.com/products/gedit/reviews

smed79:


easylist/easylist_specific_hide_abp.txt
```diff
@@ -32,6 +32,7 @@ scamwarners.com#?#center:-abp-contains(Advertisement)
 infographicjournal.com#?#.et_pb_widget:-abp-contains(Partners)
 infographicjournal.com#?#.et_pb_module:-abp-contains(Partners)
 infographicjournal.com#?#.et_pb_module:-abp-contains(Partners) + .et_pb_module
+g2.com#?#.full-width:has(a[rel^="sponsored"])
```
monzta, line 35: When using `:has()`, please don't use `#?#` as this is less efficient, use `##` instead.

I wasn't able to verify which element your filter suppos to remove on https://www.g2.com/products/gedit/reviews though.

smed79: Here an update https://github.com/easylist/easylist/pull/21503

PR #20998, opened 2025-01-06
A: https://wolt.com/ 

user9517:
fixed wolt.com sponsored hiding.

Instead of showing a blank div, it removes the div and shows only non-sponsored divs.

before:
image
after:
image

Link for this page: https://wolt.com/en/deu/bielefeld/restaurants


Khrin:
Thank you @user9517

PR #21023, opened 2025-01-09
A: block digitec-Galaxus pageview (including european version)

GeekCornerGH:
Digitec and Galaxus.ch are owned by the same company.
They recently expanded to Europe too, so I included the european sites too.


Khrin:
https://github.com/easylist/easylist/commit/e7a471d10e5df816e7a3996851fcddc600cfa4fd

GeekCornerGH:
@Khrin That's ok, what about Digitec.ch tho?

PR #21025, opened 2025-01-09
A: asuracomic.net

Kladki:
Before:
image
After:
image


Khrin:
This is not an advertisement. it won't be blocked in EasyList.

PR #21152, opened 2025-02-01
A: tfwm.org.uk

Kladki:
Before:
image
After:
image


Khrin:
@Kladki Seems to be already fixed.

PR #21165, opened 2025-02-03
Remove nslookup.io/pl.js

ruurtjan:
This is Plausible analytics, which doesn't track PII, and doesn't use cookies: https://plausible.io/privacy-focused-web-analytics

It's also self-hosted, so it's not sent to a third party.

> EasyPrivacy is an optional supplementary filter list that completely removes all forms of tracking from the internet, including web bugs, tracking scripts and information collectors, thereby protecting your personal data.

So I'd say it doesn't belong on this list?


Yuki2718:
We block analytics: https://github.com/easylist/easylist?tab=readme-ov-file#easyprivacy, we never said we unblock if it does not "track" user.

PR #21242, opened 2025-02-14
A: https://gamesfree.com

furdiburd:
Remove preroll ad on gamesfree.com and fix the game window not showing up due to it being hidden while the ad plays.  Website to test: https://www.gamesfree.com/game/pre_civilization.html

Before: 
easylist_before

After: (ignore ruffle error, it is normal for this site) 

easylist_after

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -8,6 +8,8 @@ seafoodsource.com###Ad1-300x250
 seafoodsource.com###Ad2-300x250
 seafoodsource.com###Ad3-300x250
 boredbro.com###AdBox728
+gamesfree.com##div#prerollad
+gamesfree.com##div#flashgame:remove-attr(style)
 moviekhhd.biz,webcarstory.com###Ads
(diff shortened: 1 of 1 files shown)


furdiburd:
@EasyListHebrew 

Does anyone even maintains easylist these days?

Yuki2718:
With your filters I only see a black screen. Not sure if they're worth adding to other lists.

furdiburd:
That is expected, the site uses flash and if you dont have ruffle extension installed then it displays a black screen.

furdiburd:
Also , i tried a different method here, is this allowed or i should just stop trying fixing that site? https://github.com/easylist/easylist/pull/21415

Yuki2718:
Though a CSS filter can be used, if it leaves the black screen for the rest of user, we won't add the filters. It's not too annoying anyway.

furdiburd:
the black screenis normal if the user don't have ruffle installed otherwise he can not play anything on the site. It is even recommended by the site itself to have a flash eytension installed

PR #21274, opened 2025-02-20
A: (nsfw) https://www.gamesofdesire.com/

rakleed:
https://www.gamesofdesire.com/


Yuki2718:
Intentional. These are more of self-promotion.

god

At first I considered blocking in FA, but some of them are too aggressive to address in optional lists.

PR #21287, opened 2025-02-24
A: (nsfw) https://gamcore.com/games/exiles

rakleed:
https://gamcore.com/games/exiles

### DIFF
--- a/easylist_adult/adult_specific_hide.txt
+++ b/easylist_adult/adult_specific_hide.txt
@@ -254,6 +254,7 @@ boyfriendtv.com##.adblock
 iporntoo.com##.adbox-inner
 ftopx.com##.add-block
 sex3.com##.add-box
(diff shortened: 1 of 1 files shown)


easylist_adult/adult_specific_hide.txt
```diff
@@ -642,6 +644,7 @@ efukt.com##.media_below_container
 lesbianbliss.com,mywebcamsluts.com,transhero.com##.media_spot
 ryuugames.com##.menu-item > a[href^="https://l.erodatalabs.com/s/"]
 camcam.cc##.menu-item-5880
+gamcore.com##.menuArea a[href="https://www.porngames.tv/"]
```
Yuki2718, line 647: @rakleed This and the rule below are not ads, but links to their partner sites (also linked back by the destination). The linking is one of the fundamentals of the Internet and we don't want to break this.

PR #21296, opened 2025-02-25
P: https://decrypt.co/307215/ethereum-bitcoin-fall-bybit-confirms-hack

rakleed:
P: https://decrypt.co/307215/ethereum-bitcoin-fall-bybit-confirms-hack


Yuki2718:
What is broken? The new rule hides an article card

<details>
<summary>Screenshot</summary>

decrypt

</details>

rakleed:
> What is broken? The new rule hides an article card

The old rule left empty space between text. Now, make the new rule more specific so it blocks only ads.

Does `P:` get added when the rule breaks something? I think `P:` is added when an existing rule changes.

Yuki2718:
> Does `P:` get added when the rule breaks something? I think `P:` is added when an existing rule changes.

https://github.com/easylist/easylist/discussions/16755#discussioncomment-6771757

BTW looks `.my-10` is a better selecter here from the site owner's perspective, though this hides "Sponsored Press Releases" on the top page if used alone:

decrypt

PR #21314, opened 2025-02-27
M: https://www.huffpost.com/

rakleed:
https://www.huffpost.com/


Yuki2718:
You can use `A:` in this case. `M` is used if you can't specify url.

PR #21321, opened 2025-02-28
fix tedbaker.com

iambrianfung:
this rule doesn't seem to solve for ios.
any help with that is appreciated.


Yuki2718:
@JobcenterTycoon Is there a problem if cosmetically hidden?

JobcenterTycoon:
Yes scroll is blocked

`tedbaker.com##body:style(overflow: auto !important)`

PR #21415, opened 2025-03-13
A: https://gamesfree.com and A: https://telex.hu

furdiburd:
This removes the pre-roll ad on gamesfree.com when trying to play a game

Changed the method so it does not use :remove-attr() when removing the preroll ad. Flash support or ruffle need to be installed otherwise you get the black screen.

site to test it: https://www.gamesfree.com/game/shop_empire_rampage.html


furdiburd:
Without these: 
before
Using these filters:
after

furdiburd:
The black screen is there because modern browsers dropped flash support, you need to have ruffle or other flash to html extensions installed to play the games on the site

Khrin:
> EasyList policy
> 
> The goal of EasyList is to block ads on English and International sites.

`https://easylist.to/pages/policy.html`

PR #21421, opened 2025-03-14
M: [Breakage] https://www.nkz.cz/

CbumDino:
Styling is broken at https://www.nkz.cz/


Yuki2718:
Shouldn’t you exclude `nkz.cz` and not `ads.burdadigital.cz` in this case?

CbumDino:
Other domains are/can be affected E.g - https://g.cz/ or video breakage at https://www.lifee.cz/
Screenshot 2025-03-14 at 12 26 43
Screenshot 2025-03-14 at 12 27 18

Yuki2718:
`domain=~` doesn't work anyway. In that case what's needed will be `@@`.

Yuki2718:
@CbumDino Feel free to comment or suggest something if you have any.

CbumDino:
@Yuki2718 Fixed.

PR #21491, opened 2025-03-21
P: https://foreignpolicy.com/2025/03/19/how-negotiate-putin-trump-ukr…

rakleed:
…aine/

 The `##.share-dropdown` rule blocks the Save button, which is a site functionality.


Khrin:
There is no need for this.

PR #21492, opened 2025-03-21
M: outlook.live.com

avanikaein:
"upgrade your account" appears at top of outlook inbox with "Ad" label

Screenshot 2025-03-21 at 1 38 41 PM


Yuki2718:
But the rule works only for English interface.

avanikaein:
Thanks for pointing that out @Yuki2718.
Your suggested filter works on my side so I'll go ahead and close this PR in favour of your suggestion.

PR #21531, opened 2025-03-27
Create an AI filter list

nullnominal:
I'll add stuff to this over time as I find AI stuff in sites where it should not exist.


nullnominal:
I feel like this would be useful for most people because companies throw AI stuff in our face often.

PR #21578, opened 2025-04-03
M: Update FOP.py (Python syntax does not require a semicolon at the end of a statement)

smed79:


Khrin:
We've moved to [https://github.com/ryanbr/fop-rs](url).

PR #21579, opened 2025-04-03
M: Update FOP.py (Color output of added and removed lines in commit diff)

smed79:
Updates the commit function to color the output of removed lines in red and added lines in green, improving readability.


Khrin:
Not needed anymore.

PR #21692, opened 2025-04-22
M: Removed

smed79:
The filter `##.cookie-warning` is not in easylist_cookie.


smed79:
https://github.com/easylist/easylist/pull/21691

PR #21730, opened 2025-04-28
A: https://fontawesome.com

salim-b:
Or alternatively remove this line from `easylist_cookie/easylist_cookie_specific_hide.txt` and instead just add `###cookie-consent` to `easylist_cookie/easylist_cookie_general_hide.txt`?


Khrin:
@salim-b It's better to keep it specific for now.

PR #21765, opened 2025-05-02
Fix loading embedded article content on beobachter.ch

salim-b:
Fixes embedded article content like diagrams which are hidden behind an "allow cookies" placeholder. Without this change, clicking "allow cookies" has no effect and the embedded content isn't loaded.

Site to test: https://www.beobachter.ch/magazin/wohnen/so-gibts-platz-fur-alle-777408


Khrin:
The filter has been removed.

Issue #21244, opened 2025-02-14
Request for Whitelisting: convertexperiments.com (False Positive in EasyPrivacy)


mysticaltech:
### **Issue Description**
The domain `convertexperiments.com`, which belongs to [Convert.com](https://www.convert.com) (a leading A/B testing platform), has been listed in EasyPrivacy as a tracking domain. However, this classification is incorrect and negatively impacts website functionality for many businesses that rely on Convert for controlled experiments. 

Convert does not engage in cross-site tracking, advertising, or user profiling beyond the context of the site implementing the A/B test. Blocking `convertexperiments.com` prevents A/B test variations from loading, breaking website functionality and user experience.

---

#### 1. **Convert is Not an Ad Server (EasyList Policy)**
- Convert does **not** serve advertisements, affiliate links, pop-ups, or any third-party promotional content.
- Convert does **not** insert ad elements into a page or act as a clickthrough server.
- Convert’s sole function is **website optimization through controlled experiments** (A/B testing), which is not advertising.
- Competitor services such as **VWO (`dev.visualwebsiteoptimizer.com`) are not blocked**, despite performing the same role.

Blocking Convert under EasyList policies would be a mistake, but it appears to have been added under **EasyPrivacy**, which requires a separate analysis.

---

#### 2. **Convert Does Not Fit the EasyPrivacy Criteria**
EasyPrivacy aims to block user tracking technologies, including:
- **Third-party tracking and fingerprinting**
- **Cross-site tracking & persistent identifiers**
- **Tracking pixels, beacons, telemetry, and resource miners**

Convert does not meet these criteria because:
✅ Convert only operates in a **first-party context** – It does not track users across websites or collect persistent identifiers.
✅ Convert does **not sell or share user data** for advertising or third-party analytics purposes.
✅ Convert is used **only by the website owner** to analyze variations of their own site.
✅ Convert does **not fingerprint users** – it does not rely on browser uniqueness or network signatures to persistently identify users.

In contrast, Convert has been mistakenly classified as a tracking server, despite being used for **site-specific experimentation** rather than cross-site tracking.

---

#### 3. **Blocking Convert Breaks Website Functionality**
Convert enables website owners to **test different layouts, buttons, headlines, and content** to improve user experience. Blocking `convertexperiments.com`:
- Prevents A/B test variations from being displayed.
- Causes blank sections or missing content on websites.
- Negatively impacts user experience by disabling functionality that website owners intend to provide.

[...]

PR #21808, opened 2025-05-13
D: convertexperiments.com

mysticaltech:
Hey @ryanbr,

This addresses https://github.com/easylist/easylist/issues/21244, it is to remove that false positive, it's legit, we are an AB testing service (convert.com) that just run AB tests by our customers on their own websites, for them and by them, we also run permanent personalizations for them, the only metrics we collect are related to the AB tests like Conversion Rates and Revenue per Visitors, we're the most private and conscious in the whole industry, and our worst competitors are not banned, while we are. We are not an analytics service, we are a web experimentation service.

All is explained in the issue. As an open-source maintainer myself, **I wouldn't lose your time tagging you if I didn't have a solid case**. 

Thank you for your consideration.

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
+++ b/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
@@ -509,7 +509,6 @@
 ||conversionruler.com^$third-party
 ||convertagain.net^$third-party
 ||convertcart.com^$third-party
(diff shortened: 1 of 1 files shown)


Khrin:
Someone from a notorious tracking service wants to remove the rule that blocks their tracker.

That's not very nice.

PR #21815, opened 2025-05-16
A: `maps.googleapis.com/maps/api/js/QuotaService.RecordEvent` if domain = geoguessr.com

ziIIas:
[GeoGuessr abuses the block of this](https://silliest.website:3/blog/you-dont-cost-geoguessr-any-money/)


ryanbr:
I don't see we need to "fix" this

PR #21840, opened 2025-05-22
A: internal-t.posthog.com

Wuemeli:


ryanbr:
Where is this from? @Wuemeli

Wuemeli:
> Where is this from? @Wuemeli

Posthogs Dashboard

orian:
@ryanbr Would you consider reverting this? It makes the functional analysis of our own website hard. This domain is used by PostHog to functionally analyze and improve our own product.

PR #21918, opened 2025-06-09
A: anfieldindex.com & eplindex.com

avanikaein:
ads visible on the right sidebar on both domains


Khrin:
@avanikaein The `:has` filters should go in their respective section (under line 6888), while the `:-abp-contains` filter should be in `easylist_specific_hide_abp.txt.`

avanikaein:
Hey @Khrin, thanks for the notes!
Updated to appropriate template files/sections.

PR #21966, opened 2025-06-22
A: https://nodejs.org/

wukko:
this rule removes the sponsored button on nodejs' homepage, but leaves the sponsor's mention on other pages (such as https://nodejs.org/en/about/previous-releases#commercial-support)

before:
image

after:
image


pmnlla:
based asf lgtm

wukko:
not sure why this PR was replaced by a commit that does the same thing, but thanks for merging it either way

Yuki2718:
I just fixed the issue by myself. Yours was almost fine, but we don't add a comment to each rule and also the position to put the rule was wrong. I sometimes ask to correct, other times fix by myself; it's just this time I was addressing this while at the same time discussing with other members in background, so just wanted to quickly fix.

wukko:
thank you!

PR #22019, opened 2025-07-01
M: scripts.stay22.com

wololodev:
With more granularity on what to block: 
scriptpopupblock

### DIFF
--- a/easylist/easylist_adservers_popup.txt
+++ b/easylist/easylist_adservers_popup.txt
@@ -2572,6 +2572,7 @@
 ||statistic-data.com^$popup
 ||statsmobi.com^$popup
 ||stawhoph.com^$popup
+||stay22.com/allez/roam*source=pop*^$popup
 ||steerensout.net^$popup
 ||stemboastfulrattle.com^$popup
 ||stenexeb.xyz^$popup
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -1950,7 +1950,6 @@
 ||scribol.com/traffix/widget_tracker/
 ||script-api.kinja.com/script?
 ||scripts.psyma.com^
-||scripts.stay22.com/letmeallez.js
 ||scripts.swifteq.com/hc_events.js
 ||sd-tagging.azurefd.net^
 ||sdiapi.com/reporter/


ryanbr:
Need more info, site having issues and tests. Not happy with this suggestion.

wololodev:
> Need more info, site having issues and tests. Not happy with this suggestion.

Of course, the premise is that `scripts.stay22.com` is a functional script to help creators build their hotel travel deeplinks. 

One of the features is Nova (popup) which is toggled on by some creators such as: 
`https://www.timetravelturtle.com/kyrgyzstan/soviet-murals/`
`https://www.journalofnomads.com/fergana-valley-uzbekistan/`
`https://backpackmoments.com/fergana-valley-uzbekistan-full-guide/`

so this new rule disables the popup and not the whole functionality which is required for blog to run its swapped links

ryanbr:
Why should we care about hotel travel deeplinks?

wololodev:
> Why should we care about hotel travel deeplinks?

Thnaks Ryan. You're right, we shouldn't, and we shouldn't block them either.
It's just about not breaking the web too much while blocking what's needed. The proposed rule is more granular, blocking the pop product only from it being open as it closes immediately.

Yuki2718:
`$popup` is just an after-care, it doesn't actually prevent popup to happen.

ryanbr:
Lets leave it, gone through the links, nothing broken. Seems good.

wololodev:
We've essentially broken the web a bit for creators that use travel affiliate links to monetize by hard blocking the whole script (and instead, we should block simply the analytics and the pop part) as detailed in this pr

Because of this new rule being too general, the following links are blocked, which usual are normal links:
- [https://travelmexicosolo.com/best-mazunte-hotels/](https://travelmexicosolo.com/best-mazunte-hotels/#:~:text=Outdoor%20swimming%20pool-,%E2%86%B3%20Book%20it,-%232%20PICK)
- [https://lovevenice.net/venice-lagoon-islands/burano/](https://lovevenice.net/venice-lagoon-islands/burano/#:~:text=Giudecca%0AHotels%20%26-,Apartments,-VIEW/BOOK)
- [https://mon-grand-est.fr/lorraine-en-decembre/#:~:text=R%C3%A9servez%20votre%20h%C3%A9bergement%20%C3%A0%20Metz](https://mon-grand-est.fr/lorraine-en-decembre/)

and many others i could share more, from the facebook group community of travel bloggers. 

recording of screen demo:
https://github.com/user-attachments/assets/3f6ec74d-9152-4ff5-b22e-eec5d9b07c33

The follow results in a "whitescreen of death" in the content creator world because they are not registered(in this case, no account with booking.com directly, which was when letmeallez.js script does that for them on the fly)

I am sorry for re-opening an old issue, but there are alternatives to the proposed fix if we want something more general but still non-blocking. Thank you for looking into it again 🙂

Yuki2718:
Not real breakages, these links are just obsolete and redirects to search page of `booking.com` if unblocked.

wololodev:
Thank @Yuki2718  I'm not sure I would agree they are obsolete (they seem to redirect to a blank page to me unless unblocked. If unblocked, it's the hotel page ✅) but I can also understand the priority isn't high on fixing this. if others report it then maybe we can look at it again 👍🏼

Yuki2718:
If not obsolete or expired, then maybe the sites put wrong links. Links on `https://backpackmoments.com/fergana-valley-uzbekistan-full-guide/`, which will be hijacked if the script is unblocked, are working fine with the script blocked.

wololodev:
yes indeed, some links do work on other creators websites but that's not the majority. From the link you shared, it seems that they are working with some advertiser instead of natural hotel link, interesting..

Here's more websites where it doesnt work:
- https://www.trilhaseaventuras.com.br/garanta-uma-viagem-de-offroad-tranquila-e-segura/
- https://www.viajesyrutas.es/2025/03/que-ver-en-sintra.html
- https://notaboutthemiles.com/helsinki-in-one-day/
- https://japlanease.com/where-to-stay-kyoto-first-trip/
- https://thisdarlingworld.com/2-days-in-chicago-hotel-with-a-view-ihg/
- https://mcglobetrotteuse.com/2025/05/19/avis-french-bee-montreal-paris-orly/
- https://planreadygo.com/1-day-in-seattle/

PR #22034, opened 2025-07-09
Remove the Amazon Prime auto-renew disabled nag screen

Ten0:
Amazon is hoping that I'll accidentally click this after not accidentally clicking 3 times on "yes I want to disable auto renew". 🤷

image

Go away, nag screen!


ryanbr:
This could be useful? I don't want to break functionality.

Ten0:
AFAICT that is not so much about functionality as it is about trying to instill doubt about canceling the subscription, which can always be re-taken on next order it necessary anyway.

PR #22062, opened 2025-07-18
M: https://www.eloshapes.com/

brian6932:
Evading #21794, they changed `promos` to `prmos`.


Yuki2718:
We avoid universal selector even if the site in question constantly circumvent our filters.

brian6932:
> We avoid universal selector even if the site in question constantly circumvent our filters.

This will go on forever then. May as well just block the whole `.side-column`, there's nothing useful there anyway. The selector query was actually being relatively conservative.

Yuki2718:
If they're looking, nothing works permanently. They can change to `.side-clmn` etc.

PR #22066, opened 2025-07-20
A: https://www.itpro.com

salim-b:
Hide inline ad

Site to test: https://www.itpro.com/security/cyber-attacks/all-us-forces-must-now-assume-their-networks-are-compromised-after-salt-typhoon-breach


Khrin:
@salim-b i think you can leave out the `div`

salim-b:
@Khrin Ok, done.

PR #22197, opened 2025-08-15
A: samsclub.com

avanikaein:
Carousel images on homepage are broken, only first one appears & others just display an empty container.

Tested with only EasyList enabled

Screenshot 2025-08-15 at 8 38 24 AM


Yuki2718:
On my end `@@||securepubads.g.doubleclick.net/gampad/ads?$domain=samsclub.com` is also needed.

PR #22301, opened 2025-09-04
A: https://gamesfree.com

furdiburd:
Removes preroll ad before loading games on gamesfree.com

Before: 

kép

After (ruffle fails on this site even without this rule, but it instantly loads):

kép


ryanbr:
blocking preroll ad breaks too much

furdiburd:
Could you explain what you mean by 'breaks too much'? It seems to work fine for me.

Without the rule
https://github.com/user-attachments/assets/2e889d33-2647-4401-8b40-b77a42cd401b

With the preroll block rules.


https://github.com/user-attachments/assets/266e3d2b-4e4f-4f73-85e1-e1e4528fdbc4

PR #22362, opened 2025-09-13
A: Adding multiple ad websites

The-Flash-Routine:


Khrin:
The filters are impossible to verify and are listed in the wrong file anyway.

The-Flash-Routine:
> The filters are impossible to verify and are listed in the wrong file anyway.

Hi @Khrin 
Best way to check is to put this on VirusTotal website and look for results
https://www.virustotal.com/gui/home/search

I see that websites keep on creating urls with weird/gibberish names to constantly evade from detection list.
Can you please:
1. Create a dedicated file, where i can keep on adding these URLs, as and when I find.
2. We can include this file into the main fanboy-annoyance ublock-syntax file, that serves as filter to Brave Browser

Khrin:
Fnboy's Annoyances list serves a different purpose and isn't a default list, AFAIK.

Adservers are already listed in their own constantly updated list: https://github.com/easylist/easylist/blob/master/easylist/easylist_adservers.txt

Thank you.

PR #22408, opened 2025-09-22
A: https://www.rewe.de/shop/

allendema:
Example:
 - https://www.rewe.de/shop/c/nudeln/


Khrin:
1.

Khrin:
1. The `$image` tag on the first filter can be removed, as it is not needed.
2. It's better to avoid blocking this element, so please remove the second filter.
3. Please remove the `www.` from the remaining filters, as it is not necessary.

Thanks!

allendema:
Changed now.

PR #22434, opened 2025-09-25
Update fanboy_agegate_specific_uBO.txt

Neatly7244:
Change pornhub.com cookie value from 1 to 2 to avoid blurred out pictures on mobile browsers (specifically Brave and fanboys annoyances + UBO list).


ryanbr:
resolved

PR #22453, opened 2025-09-29
A: https://gamesfree.com

furdiburd:
Removes fake loading screen with a static ad before showing the game content window.

Site to test this:

https://www.gamesfree.com/game/shop_empire_rampage.html

You need to have ruffle extension installed when testing this otherwise you get a black window instead which will look like this, it is due to mainstream browsers no longer supporting flash not because these rules breaking the site.

Képernyőkép 2025-09-29 110740

Without:

485584901-c9acdd56-e547-44b8-b053-5b71d5a3a198

https://github.com/user-attachments/assets/798fa572-12d1-4aa2-a320-de179fa014ec

With these rules:

485585147-124d829c-a36f-4880-b94a-7380b77c8781

https://github.com/user-attachments/assets/d2db5f9f-1e32-41b7-b2cf-1519dda747b0

I could not possibly provide more info that this does not break anything on this site.

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -9,7 +9,7 @@ seafoodsource.com###Ad2-300x250
 seafoodsource.com###Ad3-300x250
 boredbro.com###AdBox728
 gamesfree.com##div#prerollad
-gamesfree.com##div#flashgame:remove-attr(style)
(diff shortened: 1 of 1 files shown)


ryanbr:
Send this too uBO.

furdiburd:
Got it.

PR #22455, opened 2025-09-29
A: EFF Cover Your Tracks

cyyynthia:
Add the EFF marketing stunt to the list. Cover Your Tracks [expects tools to include their URLs](https://coveryourtracks.eff.org/about#:~:text=we%20strive%20to%20have%20our%20test%20domains%20included%20in%20such%20tools%E2%80%99%20lists) in their filtering...

Inclusion to EasyList will stop the extremely misleading "partial protection" result from showing up in all major web browser using an ad-blocker and put an end to misinformation campaigns like this one from everyone's favorite ~~adware~~ browser: https://x.com/brave/status/1914122556375986422

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -482,6 +482,7 @@
 ||discover-metrics.cloud.seek.com.au^
 ||dls-account.di.atlas.samsung.com^
 ||dmtgvn.com/wrapper/js/common-engine.js$domain=rt.com
+||do-not-tracker.org^
 ||docs.github.com/events
 ||dollartree.com/ccstoreui/*/analytics/
 ||domainit.com/scripts/track.js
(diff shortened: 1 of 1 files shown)


ShivanKaul:
This is not really a "gotcha"; adding these testing domains to EL/EP makes sense to me too. EFF says in the [Methodology section](https://coveryourtracks.eff.org/about#methodology) of Cover Your Tracks that they're not trying to hide these testing domains. It would be silly to do that since it would be trivial to figure out which ones are used. These kind of "domain-based" blocking tests are really more effective for heuristics-based blockers like Privacy Badger. I wrote about how these kind of tests are fundamentally flawed (for lists-based blockers) in https://brave.com/blog/adblocker-testing-websites-harm-users/.

ryanbr:
Was added https://github.com/easylist/easylist/commit/aace3632

PR #22529, opened 2025-10-12
A: golem.de

uniquestring:
Not sure if I put this in the correct file and place. I couldn't find any guidelines for this, but I hope this change is fine.

This hides ad 'news' articles that are marked as such. URL for testing: https://www.golem.de/

golem


Khrin:
It's already covered in https://github.com/easylist/easylistgermany

PR #22543, opened 2025-10-16
A: https://search.brave.com/

dumbmoron:
Hides Brave ads on Brave Search front page and SERP

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -1089,6 +1089,9 @@ wordhippo.com###sgwDesktopBottomDiv
 v3rmillion.net###sharingPlace
 fedex.com###shop-runner-banner-link
 search.brave.com###shopping
+search.brave.com##.download-cta
+search.brave.com##.mobile-promo-content-wrapper
+search.brave.com##dialog:has(.promo-content)
 shareinvestor.com###sic_superBannerAdTop
 filedropper.com,lifenews.com###sidebar
 whatsondisneyplus.com###sidebar > .widget_text.amy-widget


Khrin:
It's already blocked in: [https://github.com/easylist/easylist/tree/master/fanboy-addon](https://github.com/easylist/easylist/tree/master/fanboy-addon), list used for annoyance-only elements. This doesn't qualify as an "ads".

dumbmoron:
> It's already blocked in: https://github.com/easylist/easylist/tree/master/fanboy-addon, list used for annoyance-only elements. This doesn't qualify as an "ads".

@Khrin could you import them into there then? It does not look to me like the last two elements are in any lists.

dnduejsdjxudidgv-hash:
googleads.g.doubleclick.net

wukko:
> This doesn't qualify as an "ads".

@Khrin how exactly does it not qualify as an "ad"? it's promoting a product on an page that is not about the product in question.

apricotbucket28:
> > This doesn't qualify as an "ads".
> 
> @Khrin how exactly does it not qualify as an "ad"? it's promoting a product on an page that is not about the product in question.

Not from the EasyList team, but the README says:
> Self-promotion [1](https://github.com/easylist/easylist#user-content-fn-1-a8be073366358aa8f8077f89fdb46279) won't be targeted directly.

and
> Self-promotion is any type of advertising that promotes goods or services that are owned or operated by the domain owner and doesn’t get commercially compensated for by third parties (examples can include new features, new posts, newsletters, subscriptions to printed media products, merchandise etc.)

PR #22665, opened 2025-11-18
Remove ultimedia.com editorial player blocking rule

christopherosko:
Removed blocking rule for ultimedia.com smart.js script.


ryanbr:
Why? Its been in the list since 2021.

christopherosko:
It has been on the list for a long time, but more and more publishers would like their editorial video content to accompany articles.
Furthermore, for users without adblockers, these videos are considered by browsers to be heavy advertisements due to their presence in Easylist and are destroyed after a few seconds of viewing, penalising the user experience.
https://issues.chromium.org/issues/457241501
We have also noticed that other players in a similar situation have been unblocked, for example here: https://forums.lanik.us/viewtopic.php?t=48634-lesnumeriques-com
Thank you for your understanding

christopherosko:
Any update on this ?
Thanks !

ryanbr:
We're fine.

PR #22700, opened 2025-11-30
Fixed a superfluous tilde.

DandelionSprout:
Judging by uBO's logger, the double tilde broke the entry, so I've fixed it now.


Khrin:
@DandelionSprout Thank you!

PR #22706, opened 2025-12-02
P: Removal of adrsbl analytics pixel

markthepixel:
GOAL:
Removal of adrsbl pixel from list

REASON:
The adrsbl pixel is an analytics pixel, much the same as Google Analytics or Amplitude. 

This pixel does not serve ads nor create them. And has nothing to do with pushing advertising directly.


Khrin:
It's listed under "EasyPrivacy", not "EasyList".

`https://easylist.to/pages/policy.html`

PR #22713, opened 2025-12-03
A: Block Adobe helix rum-collector

palano123:
Sample page:  https://www.hersheyland.com/?rum=on

This loads a tracking script from:  https://www.hersheyland.com/.rum/@adobe/helix-rum-js@%5E2/dist/rum-standalone.js
And then sends tracking information to https://www.hersheyland.com/.rum/1


Yuki2718:
Too generic, better something like `/.rum/@adobe/helix-`

palano123:
@Yuki2718 : thanks for the feedback - that indeed is a good recommendation
Updated the PR accordingly

PR #22729, opened 2025-12-11
P: Disable /metrics/rum rule on Datadog websites

Valpertui:
Update the rule so it doesn't block some Datadog API endpoints when navigating Datadog own websites

### DIFF
--- a/easyprivacy/easyprivacy_general.txt
+++ b/easyprivacy/easyprivacy_general.txt
@@ -1498,7 +1498,7 @@
 /metrics/init?
 /metrics/onload
 /metrics/ping?
-/metrics/rum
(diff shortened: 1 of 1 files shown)


ryanbr:
We're happy to leave it

Valpertui:
Hello, 
Could you share more on why?  
This rule is breaking product features our customers are using on Datadog websites (fetching data from our API and showing it on the page).

The disabling of this rule on our domains is not about enabling back anything related to telemetry, analytics or tracking features.

We would prefer for this rule to be updated instead of having to tell ours customers to disable their AdBlockers/Privacy tools relying on EasyPrivacy lists on our domains, which would be objectively a worse situation for everyone.

PR #22812, opened 2026-01-01
Applied HTTPS to 3 in-list licence URLs.

DandelionSprout:
`http://creativecommons.org/licenses/by/3.0/` → `https://creativecommons.org/licenses/by/3.0/`

It probably goes without saying why this would need to be done.


Yuki2718:
Thx

PR #22842, opened 2026-01-06
A: Reddit sidebar AI suggestions

blunket:
On Reddit, the sidebar contains AI suggestions that cannot be removed in settings.


ryanbr:
Which element? screenshot

blunket:
> Which element? screenshot

@ryanbr 

Sorry for not including it before. It is this "Related Answers" section that appears in the sidebar at the top when browsing any thread.

The links there take you to an AI chat.

image

Here's the exact page I've screenshotted above:
https://www.reddit.com/r/symfony/comments/a4v9hv/messenger_when_and_why_use_it/

I'm not sure what the logic is for when it appears, but it doesn't seem to appear for me as often (or maybe at all?) when I'm signed in, so you could try checking in a private window if you're signed in to Reddit.

PR #22892, opened 2026-01-16
Update tracking server entry for swetrix

Blaumaus:
Hi! I'm a maintainer of Swetrix - an [OSS](https://github.com/swetrix/swetrix) web analytics platform.

Replaced swetrix.com with api.swetrix.com in tracking servers list.

Having it as just `swetrix.com` causes issues with assets loading on our websites like `docs.swetrix.com`. The root domain is NOT used for tracking. It's the UI where the users are able to access their stats.

We only use `api.swetrix.com` domain for data ingestion

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
+++ b/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
@@ -2114,7 +2114,7 @@
 ||surveywriter.com^$third-party
 ||survicate.com^$third-party
 ||sweetgum.io^$third-party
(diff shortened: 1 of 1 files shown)


ryanbr:
Won't matter tbh, it's only third-party outside of the main domain

Also: https://lastminuteparis.com/  `https://swetrix.org/swetrix.js`

Blaumaus:
Nevermind, I realised that it was blocking the assets that had `gtm/gtm-*.png` in their name

PR #22924, opened 2026-01-21
A: https://www.glassdoor.com

mister-sir:
An overlay on glassdoor pages for viewing salaries. Pops up after the user tries to make the filters more specific.


Khrin:
It's unlikely to warrant being added to the Annoyances' list.

mister-sir:
@Khrin : I'll take your word for it on what would qualify, looks like you review a ton of these pull requests. Any guidance on what the criteria are? I'm happy to contribute to easy list, but I couldn't find any clear-cut guidance on what does and doesn't make the cut.

PR #22939, opened 2026-01-23
A: https://dl.acm.org/

omentic:
Banner ads. Example: https://dl.acm.org/doi/10.1145/3776662


omentic:
I've added some stuff to remove "Upgrade to Premium" style advertising. It removes the advertising banners, and removes the greyed-out premium-only features that pop up adverts for premium when clicked. This does not remove the premium features when logged into a premium account, the `.premium` class is only used for *disabling* such things. 

I've also added a rule to disable the "Subscribe to Premium" modal that pops up when opening a page in the Digital Library for the first time. This requires removing the `modal-open` class on the body, so I had to make it uBO specific.

I've also added some rules to remove the AI-generated paper summaries button, in the appropriate "AI suggestions" list.

omentic:
Ping @ryanbr @Yuki2718 for review. Let me know if any changes should be made.

omentic:
(cc. @ryanbr and @Yuki2718 for review)

PR #22987, opened 2026-01-29
M: Update Google AI Overview ID

blunket:
On my search results page, the ID Odp5De is nowhere to be found. It seems the ID was changed.

image

(sorry for duplicate PR, made a mistake in the previous one.)

### DIFF
--- a/fanboy-addon/fanboy_ai_suggestions.txt
+++ b/fanboy-addon/fanboy_ai_suggestions.txt
@@ -8,7 +8,6 @@
 ! Generic Cosmetics
 ###kapa-widget-container
 ! Specific Cosmetics
-www.google.com###Odp5De:has-text(AI Overview)
 develop.sentry.dev###ai-list-entry
 reddit.com###answers-suggested-queries-m3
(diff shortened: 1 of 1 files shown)


blunket:
Updated to keep the old ID too... With my above search query, I only get the old ID when I'm signed out. It's nowhere on the page when I'm signed in. It seems like they're both occasionally used?

ryanbr:
Separate lines and use ./fop to sort.

Stevoisiak:
The AI overview layout can change based on the search terms, which can change what filter needs to be used. For example:
- https://www.google.com/search?q=Why+are+we+here
- https://www.google.com/search?q=Celeste+release+date
- https://www.google.com/search?q=unity+compute+dispatch+editor
- https://www.google.com/search?q=ublock+ai+results+not+showing+results


image

image

Stevoisiak:
Does the following set of filters work as an alternative to using `:has-text()`?

```
www.google.com##[data-aquarium], [data-attnms], [data-subtree="mfc"], [data-subtree="mfc,mfl"]
```

PR #22989, opened 2026-01-30
Update linkedin-nopuzzle.txt

iambrianfung:
Add back rule because there seems to be 2 types of layouts. Linkedin could be experimenting.
Screenshot 2026-01-30 at 08 50 55


ryanbr:
Need to ./fop here @iambrianfung

iambrianfung:
@ryanbr fixed

PR #23039, opened 2026-02-05
P: https://ananas.rs/ - Fix broken cookie notice rules (Serbian section)

DC989:
**Website**: https://ananas.rs/
**File**: easylist_cookie/easylist_cookie_international_specific_hide.txt
**Section**: Serbian

**Issue**:
The existing rules for `ananas.rs` target volatile Styled Component class hashes (e.g., `.cOMKpv`) which have changed, causing the cookie wall to reappear.

**Proposed Fix**:
I have removed the obsolete hash-based selectors and replaced them with a procedural cosmetic filter set.
1. Target the banner content using `:-abp-contains`.
2. Target the overlay backdrop using the adjacent sibling combinator (`+`) as it renders immediately after the banner in the DOM.

**Verification**:
Tested locally in Brave (using uBO syntax `:has-text` for verification) and refactored to ABP syntax (`:-abp-contains`) for this submission to ensure strict list compatibility.

### DIFF
--- a/easylist/easylist_adservers.txt
+++ b/easylist/easylist_adservers.txt
@@ -2730,6 +2730,7 @@
 ||alationsulafat.cfd^
 ||alatsbhblpwlp.website^
 ||albeedunged.shop^
+||albinicakasa.cfd^
 ||albinosbergut.cyou^
 ||albonsa.com^
 ||alburnunpetal.help^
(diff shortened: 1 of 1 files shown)


ryanbr:
Moved to trusted rules, rule fixed in uBO

PR #23080, opened 2026-02-09
P: https://locataire.dossierfacile.logement.gouv.fr

0o5rlubm:
This French government website allows users to create a rental file by providing supporting documents, validating them, and sending them to landlords.

It is possible to generate one link per landowner.

The filter hides the link management interface, preventing old links from being deactivated.


Khrin:
Correct file is https://github.com/easylist/easylist/blob/master/fanboy-addon/fanboy_social_allowlist_general_hide.txt

PR #23101, opened 2026-02-11
M: Jigsawplanet

LikeLakers2:
The ID previously targeted no longer exists. This PR changes it to the new ID.

Since I imagine this'll break again over time, I'll have to watch the page layout over time and see if we can't make a filter that's less prone to breakage.


Yuki2718:
That hides the whole jigsaw at `https://www.jigsawplanet.com/?rc=play&pid=020b1be14b80`

LikeLakers2:
@Yuki2718 ~~I'm not sure how. From my own testing while logged in, it only seems to hide the ad and anti-adblock message?~~

EDIT: It seems logged-in and logged-out use different IDs? Which is weird.

Yuki2718:
Confirmed, no breakage if logged in. Anyway, the added rule works for both cases on my end.

PR #23107, opened 2026-02-11
A: https://reddit.com/

Parsa307:
Hello there! this PR would add an filter to AI Suggestions to hide Reddit's Ask in the search bar.


ryanbr:
If you visit that url directly, will that rule break those users? Which element are trying to target

waifucon:
> If you visit that url directly, will that rule break those users? Which element are trying to target

nope, it won't break visiting the url directly:
Screenshot_20260213_090942

it targets the ask button in search bar, i tried poking with inspect element to get a good filter href was unfortunately the one i comed with.
Screenshot_20260213_090715

ryanbr:
`reddit.com##[ask-button-variant="enabled"]` ?

waifucon:
> `reddit.com##[ask-button-variant="enabled"]` ?

weirdly it taked out the whole search bar tho with this filter. (ublock origin)

PR #23114, opened 2026-02-12
A: Block Meta AI widget on Facebook

Rajvira10:
Tested on facebook.com. Filter targets the anchor href pattern rather than 
hardcoded class names or tracking parameters which change frequently.


ryanbr:
Will break too much similar to X/Grok, is there a specific FB AI Nag that can be targeted?

PR #23187, opened 2026-02-17
M: PostHog evades event tracking block

### LINKED ISSUE #20838: posthog.com
#### List the website(s) you're having issues with:

`https://posthog.com/`

#### What happens?

Self-explanatory (check screenshot below)

#### Screenshot(s) 

https://files.catbox.moe/o8mcr0.png
Image

#### Filter lists you're using:

EasyList, EasyPrivacy

#### Your settings:

- OS/version: Linux/6.8.0
- Browser/version: Firefox 133.0.3
- Ad block extension/version: 1.61.2

#### Other details:

N/A

eden881:
As discussed in #20838, PostHog circumvents the current rules blocking tracking event ingestion:
https://github.com/easylist/easylist/blob/852e4e89300dbecda13d7729739af9c47cc55819/easyprivacy/easyprivacy_thirdparty.txt#L331-L334

They rotate the host server, query parameter order, and URL path to evade the current filter.
This change closes those loopholes while keeping their dashboard fully functional.

Closes #20838

Before change:
image

After change:
image

### DIFF
--- a/easyprivacy/easyprivacy_thirdparty.txt
+++ b/easyprivacy/easyprivacy_thirdparty.txt
@@ -1951,7 +1951,6 @@
 ||postageapp.com/receipt/$third-party
 ||posthog.com/e$xmlhttprequest
 ||posthog.com/i/*/e$xmlhttprequest
-||posthog.com/static/array.js
 ||posthog.com/static/recorder-v2.js
 ||poweredbyeden.com/widget/tracker/
 ||ppx.com/tracking/


ghost:
at this point simply block `*.posthog.com` with dns

eden881:
It won't help unfortunately. I still want to use PostHog's product, and blocking them at the DNS level would break their website completely.
The update proposed here should be aggressive enough to block their tracking events but not the rest of their website.

PR #23188, opened 2026-02-17
M: https://gonintendo.com/

apricotbucket28:
`###first-leaderboard` hides the leftover ad container at the top
image

Removed `##.pt-5` because it's not an ad container, but rather an empty container on desktop and a tab bar on mobile.
image
image


Stevoisiak:
It may be more reliable to filter on `###first-leaderboard`

Issue #22533, opened 2025-10-13
genius.com


Daxy7:
ad placeholders
```
https://genius.com/Eminem-everybodys-looking-at-me-lyrics
```
```
genius.com##.sc-8f7a798d-0
genius.com##.sc-a1c52130-0
```
https://github.com/user-attachments/assets/176aa0d4-3f45-43f8-a72e-dfc662cd4ebf

Daxy7:
delayed ad placeholders automatic removal [10-20 seconds] remain
```
https://genius.com/Childish-gambino-v-3005-lyrics
```
https://github.com/user-attachments/assets/3a29b74d-ea37-4f2b-8e00-7367f880655b

apricotbucket28:
> delayed ad placeholders automatic removal [10-20 seconds] remain
> 
> ```
> https://genius.com/Childish-gambino-v-3005-lyrics
> ```
> 
> [github.com/user-attachments/assets/3a29b74d-ea37-4f2b-8e00-7367f880655b](https://github.com/user-attachments/assets/3a29b74d-ea37-4f2b-8e00-7367f880655b)

`genius.com##div[class^="sc-"]` (included in the PR) hides those placeholders.

PR #23190, opened 2026-02-18
M: https://genius.com/

apricotbucket28:
Fixes https://github.com/easylist/easylist/issues/22533

On https://genius.com/Radiohead-creep-lyrics:
| | Before| After |
|--------|--------|--------|
| Desktop | image| image|
| Mobile | image| image| 

Not sure if this [rule](https://github.com/easylist/easylist/blob/073f7a9589bf6722230b43c7a914c890d1b8f3da/fanboy-addon/fanboy_notifications_specific_hide.txt#L596) is still used, left it there just in case.

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -1513,7 +1513,6 @@ semiconductor-today.com##.Horiba-banner
 artsy.net##.IITnS
 sporcle.com##.IMGgi
 yandex.com##.ImagesViewer-SidebarAdv
-genius.com##.InreadContainer-sc-b078f8b1-0
 timesofindia.indiatimes.com##.J7krm
 doodle.com##.JupiterLayout__left-placement-container
 doodle.com##.JupiterLayout__right-placement-container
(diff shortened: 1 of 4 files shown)


apricotbucket28:
Removed `##div[class*="AndroidBannermobile_"]` since I couldn't reproduce it anywhere on multiple devices and pages.

`##.android_banner` blocks the mobile app banner found [here](https://genius.com/artists/Radiohead) for example:
image

`##.android_interstitial:remove()` removes [this](https://genius.com/tags/grunge) other banner:
image

`:remove()` is necessary because the banner layout is broken, and if only hidden with `##.android_interstitial` the close button cannot be pressed, leaving the layout broken without a way to close the banner.

ryanbr:
`genius.com##div[class^="sc-"]`  really? this will create issues.

Trying to address too much in one commit, what are you trying to do specifically?

apricotbucket28:
> `genius.com##div[class^="sc-"]` really? this will create issues.

All elements whose class starts with "sc-" in this case are ad containers. I do get your concern though. But using [autogenerated](https://styled-components.com/) classes to match elements, like `.bIZLkL` or `.ShareButtons__Root-sc-8c8eaea4-0` will break eventually (the [latter](https://github.com/easylist/easylist/commit/6ee76592a975c9e1ec3344ddce42c611f5810748) already does not match anymore).

Without that rule, it seems like the ad containers could be hidden with:
```adblock
! various ad containers, like in the lyrics details
genius.com##div:has(> div[id^="div-gpt-ad-"])
! mobile ad container on lyrics page
genius.com##div[data-exclude-from-selection]:has(> div[class^="sc-"])
! top ad container on lyrics page
genius.com###application > div[class^="sc-"]
! top ad container on home page
genius.com##div[class^="HomeContent-desktop__Container"] > div[class^="sc-"]
```
What do you think?


> Trying to address too much in one commit, what are you trying to do specifically?

I was trying to fix all the issues I managed to find in the site. I can definitely split the social/notifications changes to a separate PR if you want

Khrin:
It looks to be abandoned. Closing here.

PR #23244, opened 2026-02-25
P: Remove join.tlon.io - app deep linking, not tracking

wca4a:
## Summary

Requesting removal of `||join.tlon.io^` from EasyPrivacy (easyprivacy_specific_cname_branch.txt).

## What is Tlon?

[Tlon](https://tlon.io) is a peer-to-peer messaging application built on [Urbit](https://urbit.org), a decentralized identity and hosting network.

## What join.tlon.io does

- Routes invite links to open the Tlon app (e.g., `join.tlon.io/abc123`)
- If the app isn't installed, redirects to the App Store / Play Store
- Passes an invite token so the user lands in the right group after install

## What it doesn't do

- ❌ No cross-site tracking
- ❌ No fingerprinting  
- ❌ No third-party analytics collection
- ❌ No advertising or retargeting

## Why it was flagged

The domain uses Branch.io for deferred deep linking (passing tokens through app install), which is why it appears in the CNAME tracker list. However, Tlon does not use Branch for tracking or analytics—only for the deep link routing.

## Impact of blocking

When `join.tlon.io` is blocked, users with privacy extensions cannot:
- Accept group invitations
- Join communities shared via link
- Complete the onboarding flow when invited by friends

This breaks core functionality for privacy-conscious users—ironically, the users most aligned with a decentralized, privacy-focused messaging app.

## Migration plan

We are building first-party deep linking infrastructure to eliminate the Branch.io dependency entirely. However, we're requesting removal now as the current blocking incorrectly categorizes functional app infrastructure as tracking.

Thank you for maintaining these lists and for your work protecting user privacy.

### DIFF
--- a/easyprivacy/easyprivacy_specific_cname_branch.txt
+++ b/easyprivacy/easyprivacy_specific_cname_branch.txt
@@ -2368,7 +2368,6 @@
 ||join.stuypend.com^
 ||join.talker.network^
 ||join.thekrishi.com^
-||join.tlon.io^
 ||join.travelxp.com^
 ||join.vibely.io^
 ||join.vtail.co^


ryanbr:
```
;; ANSWER SECTION:
join.tlon.io.           300     IN      CNAME   custom.bnc.lt.
```

Yuki2718:
> ## Impact of blocking
> When `join.tlon.io` is blocked, users with privacy extensions cannot:
> 
> * Accept group invitations
> * Join communities shared via link
> * Complete the onboarding flow when invited by friends

We almost always need an actual URL. (People too often think just explaining something without providing real samples is enough. Maybe we should emphasize this on CONTRIBUTING.md?).

ryanbr:
"deep linking" term could be used for any CNAME trackers. Seems to be a poor excuse to not implement any CNAME counters. Just remove the CNAME.

wca4a:
Here's a distinct example:
https://join.tlon.io/0v1oq7g.qtqf4.qg58h.bvcbo.fjpd2

wca4a:
> > ## Impact of blocking
> > When `join.tlon.io` is blocked, users with privacy extensions cannot:
> > 
> > * Accept group invitations
> > * Join communities shared via link
> > * Complete the onboarding flow when invited by friends
> 
> We almost always need an actual URL. (People too often think just explaining something without providing real samples is enough. Maybe we should emphasize this on CONTRIBUTING.md?).

Here's an example: https://join.tlon.io/0v1oq7g.qtqf4.qg58h.bvcbo.fjpd2

Yuki2718:
Looks certainly a tracking link, it first redirects to `sa96e.app.link` and the resulting link has two paramters of `_branch_referrer` and `_branch_match_id`.

wca4a:
Those parameters are Branch's deferred deep linking mechanism, not cross-site tracking.

**What they do:**
- `_branch_match_id` - Links the click to the subsequent app open so the user lands in the correct group after install
- `_branch_referrer` - Carries the invite token through the app store install flow

**What they don't do:**
- No data is sent to third-party advertisers
- No fingerprinting or cross-site profiling
- Tlon doesn't use Branch's analytics dashboard

The alternative without deferred deep linking: user clicks invite → installs app → opens app → has no idea what group they were invited to → has to get the link again. That's a broken UX.

We're building first-party infrastructure to replace Branch entirely. But the current implementation is functional deep linking, not tracking. The data exists solely to complete the invite flow.

jamesacklin:
Jumping in here as the product manager for Tlon—we are using Branch as a means to connect people to the group they intend to join via deferred deep-linking. We do not utilize any of the Branch metrics, nor are we proprietors of user data.

Yuki2718:
Disabled the rule, the link is correctly deep-linked with your mobile app. Note we don't distinguish third-party tracking and first-party, nor if tracking is tied to advertising or not. Deep-link can at any time be used to track user.

jamesacklin:
@Yuki2718 Thank you so much. Like @wca4a said, we're working on our own first-party solution so this should be a moot point.

I personally use products that utilize your list so thank you again for your meaningful contributions to user privacy.

PR #23275, opened 2026-03-03
A: #cookie-bar to cookie general hide list

griffinallen:
Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2214


Khrin:
It's fine to keep it as a specific filter.
https://github.com/easylist/easylist/blob/master/easylist_cookie/easylist_cookie_specific_hide.txt#L3132

PR #23302, opened 2026-03-04
A: (NSFW) hentai2read.com: Ad in manga list

Stevoisiak:
Ads are shown as results in manga list


Khrin:
I can't see where it would be needed. Perhaps it's gone.

Stevoisiak:
> I can't see where it would be needed. Perhaps it's gone.

The ads appear when searching by tag. They are still present.

PR #23308, opened 2026-03-04
A: Klook.com

griffinallen:
Hiding klook.com cookie banner

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/385


griffinallen:
Hey @Khrin, can you explain what you did differently here? I noticed a bunch of my PRs are fixed in a similar way to this one. If there is something I am missing, I'd like to know, so that I can avoid creating extra work for you.

PR #23380, opened 2026-03-12
A: chatgpt.com

Sbrjt:
Hide disclaimer element on chatgpt.com


Khrin:
There isn't an actual need to remove it.

PR #23381, opened 2026-03-12
A: chatgpt.com

Sbrjt:
Hide "Free offer" on chatgpt.com

image


Khrin:
It doesn't seem to be an advertisement.

Sbrjt:
@Khrin 

Will it be considered as an annoyance (fanboy)?

PR #23382, opened 2026-03-12
A: samsungcloud.com

griffinallen:
Hiding cookies from samsungcloud.com

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2047


Khrin:
Not reproducible.

griffinallen:
This is reproducible from the UK
Screenshot 2026-03-16 at 9 29 15 AM

PR #23390, opened 2026-03-12
Add Husky hook for fop sort

Sbrjt:
We need to run `fop sort` manually, so I’ve added a Husky hook to run it on every commit.

Also, after running `fop sort` I see many [diffs caused by line endings](https://web.archive.org/web/20130615060528/http://www.rtuin.nl/2013/02/how-to-make-git-ignore-different-line-endings/) changing from LF to CRLF on Windows. So, I’ve set `eol=lf` in `.gitattributes`. [See](https://stackoverflow.com/questions/29435156).

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,7 @@
-python FOP-nocommit.py
+#!/bin/sh
+
+case "$(uname -s)" in
+  Linux)   ./fop-linux ;;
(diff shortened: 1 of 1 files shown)


ryanbr:
Just use the binaries? FOP py hasn't been updated for a while.

Sbrjt:
@ryanbr is this ok now?

PR #23422, opened 2026-03-13
A: casetify.com

griffinallen:
Hiding cookie banner on casetify.com

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/1927


griffinallen:
@Khrin - I am still getting this cookie banner. Was my fix incorrect?

image

PR #23468, opened 2026-03-18
A: samsungcloud.com

griffinallen:
Hiding cookies from samsungcloud.com -- This is reproducible while browsing from the UK

Fixes brave-experiments/cookiecrumbler-issues#2047


ryanbr:
Unless there is more of these rules, doesn't need it's own section

PR #23469, opened 2026-03-18
A: la7.it

griffinallen:
Blocking cookies from la7.it

This change was suggested from uBlock here: https://github.com/uBlockOrigin/uAssets/pull/32247#issuecomment-4083309645

This fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2230


Khrin:
Not working.

PR #23472, opened 2026-03-18
A: grok.com

YagUber:
Add tracking filter rules for grok.com analytics endpoints, centralising rules currently split across uAssets and AdGuardFilters along with missed telemetry.


Khrin:
@YagUber  Please remove the three separators.

PR #23473, opened 2026-03-18
A: appsflyersdk.com

YagUber:
As of March 10th AppsFlyer PBA Web SDK reports to wa.appsflyersdk.com and is delivered from websdk.appsflyersdk.com.

PR #23474, opened 2026-03-18
A: wa.onelink.me

YagUber:
Add wa.onelink.me AppsFlyer ID sync endpoint, not blocked by existing $image,script scoped onelink.me rules.

### DIFF
--- a/easyprivacy/easyprivacy_specific.txt
+++ b/easyprivacy/easyprivacy_specific.txt
@@ -1687,6 +1687,7 @@
 ||t.airasia.com^
 ||t.allmodern.com^
 ||t.appsflyer.com^
+||wa.onelink.me^
 ||t.av.st^
 ||t.birchlane.com^
 ||t.dailymail.co.uk^


Khrin:
@YagUber Where is it used? Is there a link?

YagUber:
@Khrin `wa.onelink.me` is the hardcoded cross-domain identity sync endpoint in AppsFlyer Web SDK, which includes PBA, Smart Banners, and Smart Script functionality. It has been part of the SDK's default `vendorsConfig` list since at least 2019. Every site following the official integration guide for their [PBA](https://support.appsflyer.com/hc/en-us/articles/360001610038) or [Smart Banners](https://dev.appsflyer.com/hc/docs/dl_smart_banner_v2) will call it.

The SDK calls it during page load, appending a user identifier: `wa.onelink.me/v1/onelink?af_id=<user_id>`. The purpose is cross-site identity matching, correlating a web visitor with a mobile device ID across domains.

This endpoint is never blocked by existing rules because the current `onelink.me` rules are scoped to `$image,script`, while this is an XHR request. Any site embedding the AppFlyer Web SDK (whether via the new CDN at `websdk.appsflyersdk.com` or self-hosted) will make this call unconditionally on every page load.

One of Archive.org's [earliest crawls of the SDK (2019)](https://web.archive.org/web/20190925045036/https://websdk.appsflyer.com/) shows the unminified source with the endpoint hardcoded as `oneLinkUrl` with the comment "request to one link to identify the user," cross-device identity matching has been the explicit purpose since day one.

This is directly related to #23473 and only submitted as a separate PR due to the existing onelink.com rules and potential scope issues.

If blocking attribution endpoints belonging to tracking SDKs that are or are not yet covered by existing rules falls outside the scope of this project, please let me know and I'll close these PRs accordingly.

YagUber:
For example sites using `wa.onelink.me`, please see my comment on the related PR #23473: https://github.com/easylist/easylist/pull/23473#issuecomment-4104996390, any site loading the AppsFlyer Web SDK will call this endpoint, so the site list there applies here as well.

During further investigation I also found `impressions.onelink.me` fires as a $image pixel from the Smart Script, this is already covered by the existing onelink.me rules. However `wa.onelink.me` makes XHR requests outside the current `$image,script` scope, which is what this PR addresses.

It may be worth considering broadening the existing rule to `$image,script,xmlhttprequest,third-party` to cover any similar endpoints going forward, but I'll defer to your preference on whether a specific entry or a rule expansion is cleaner.

Khrin:
I couldn't find it anywhere.

PR #23477, opened 2026-03-19
A: xero.com

griffinallen:
Hiding cookie notice on xero.com. This cookie notice was visible when visiting the site from Belgium.

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/637


ryanbr:
[data-xero-component='nav/cookie-notification'] -> [data-xero-component="nav/cookie-notification"] ?

PR #23502, opened 2026-03-20
A: atthraces.com

griffinallen:
hiding cookie popup for attheraces.com

Fix suggested here: https://github.com/uBlockOrigin/uAssets/pull/32266#issuecomment-4095577475

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2001


ryanbr:
Mangled @griffinallen

PR #23537, opened 2026-03-23
A: gigafile.nu

griffinallen:
These ads are blocked on mobile for Adguard https://github.com/AdguardTeam/AdguardFilters/blob/2ee5d526b535730e1364e189b4d66394cceaab61/MobileFilter/sections/specific_web.txt#L2369

I am blocking them on the mobile and desktop versions with this commit


Khrin:
This site looks to be in Japanese only.

PR #23678, opened 2026-03-25
A: fcbarcelona.com

griffinallen:
Hiding cookie popup on fcbarcelona.com

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/1710


Khrin:
@griffinallen It breaks the site.

PR #23700, opened 2026-03-26
P: weddingbee.com from allowlist

griffinallen:
Removing weddingbee.com from allowlist

@ryanbr - This fixes the issue of the banner flashing on this page.


ryanbr:
Shouldn't be needed, we have a cosmetic fix for this now

PR #23714, opened 2026-03-27
A: centrum24.pl

griffinallen:
Hiding cookie banner on centrum24.pl (this is a brief cookie banner that appears before the trusted click element script runs)

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/391


ryanbr:
`__tealiumGDPRecModal`  has its own section

PR #23734, opened 2026-03-27
M: play.usatoday.com

avanikaein:
on game pages such as `https://play.usatoday.com/crossword`, a video ad appears in the bottom right corner of the page

using ABP + Chrome

07ce533a-3068-4a8f-8672-bf5218e1e419


Yuki2718:
Hmm, I'm not sure whether blocking `gannett-player.ex.co` globally is safe. @Khrin @ryanbr What do you think?

PR #23803, opened 2026-03-31
A: dailydot.com

griffinallen:
Hiding cookie message on dailydot.com

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2006


ryanbr:
`#cc--main` move this to a cc--main section

PR #23806, opened 2026-03-31
A: notion.so: AI

RichardoC:
This should hide both the AI prompt on the left bar, and at the bottom right


ryanbr:
FOP.rs Sort, also remove the www. @RichardoC

RichardoC:
> FOP.rs Sort, also remove the [www](http://www). @RichardoC

I've fixed the linting, but left the www because in my testing the elements I'm filtering only show on www. pages. I wasn't able to find any non www. pages to test on.

PR #23813, opened 2026-04-01
A: (NSFW) nlcosplay.com

griffinallen:
Hiding anti adblock popup on (NSFW) nlcosplay.com

Fix made for webcompat reports on Brave browser


Yuki2718:
Please report anti-adblock walls to uAssets even for NSFW sites.

PR #23838, opened 2026-04-07
A: pinterest.com

griffinallen:
Hiding cookie message on pinterest.com

Fix suggested here: https://github.com/uBlockOrigin/uAssets/pull/32401#issuecomment-4175623328

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2048
Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2250


ryanbr:
Would create a section in the _uBO list since :style()

PR #23867, opened 2026-04-10
Cosmetic fixes for cookie notices

griffinallen:
This is a batch of cosmetic fixes to hide cookie notices from the following pages. These pages had user reports for cookie notices on Brave.

https://tfgm.com
https://sacredeagleimports.co.uk
https://carplus.es
https://karrotmarket.com
https://allianzlife.com
https://speakasap.com
https://hidrive.ionos.com
https://momentumdash.com
https://chochko.com


Yuki2718:
`:has-text()` is not a CSS spec, you have to add `?` between `#` and put in _specific_ABP.txt. Remember that it should be avoided if possible - heavier, less compatibility, and often depends on your locale setting.

griffinallen:
@Khrin thanks for the feedback. I have updated this rule. It should be more appropriate now.

PR #23869, opened 2026-04-10
A: New Webloc/Penlink tracking servers

breadtk:
Citizen Lab found tracking servers for "Webloc" owned by Penlink, a geolocation surveillance company. This commit includes all known servers associated with the company.

Original report: https://citizenlab.ca/research/analysis-of-penlinks-ad-based-geolocation-surveillance-tech/

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_general.txt
+++ b/easyprivacy/easyprivacy_trackingservers_general.txt
@@ -506,3 +506,22 @@
 ||rsras.cdn-fileserver.com^
 ||sra-px.cdn-fileserver.com^
 ||youseasky.com^
+! Webloc/Penlink infrastructure https://citizenlab.ca/research/analysis-of-penlinks-ad-based-geolocation-surveillance-tech/
+||cobwebsapp.com^
+||cwtapp.com^
+||cwsystem.com^
+||13.63.16.113^
+||139.0.5.194^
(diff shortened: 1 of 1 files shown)


ryanbr:
How is blocking these IP's helpful?

breadtk:
The IP are associated with the tracking infrastructure, as detailed in the report. I think it’s fair to imply that adding IPs specifically instead of the sub domains are going to be higher false positive chance in the future as they shift servers.

Should I remove the IPs and resubmit?

Khrin:
It's unclear to me how EasyPrivacy's final user would benefit from these filters. I haven't seen any sites where such domains are used as trackers, which is what we block in EP.

breadtk:
> It's unclear to me how EasyPrivacy's final user would benefit from these filters. I haven't seen any sites where such domains are used as trackers, which is what we block in EP.

That's fair. I haven't been able to find it _directly_ on websites as well. The report mentions that this company may leverage mobile SDKs, so if these domains *do* exist, at most the end users that would benefit from this filter would be DNS sinkhole users. However - as the report also notes - it is unclear whether this infra is also directly receiving telemetry, or if it's being purchased through other data brokers.

Since I can't point to a particular domain pointing to it directly today, I'm happy to rescind this request.

PR #23954, opened 2026-04-20
A: pwc.com

Woojiyeong:
## Site

https://www.pwc.com/

## Issue

The OneTrust cookie consent banner remains visible and is not properly hidden by existing filter rules.

## Analysis

The banner is rendered inside the `#onetrust-consent-sdk` container, which is not currently covered by existing EasyList Cookie rules for this domain.

## Fix

Added a domain-specific cosmetic filter targeting the OneTrust consent container:
`www.pwc.com###onetrust-consent-sdk`

This approach avoids overly broad selectors and ensures that only the intended cookie banner element is affected.

## Testing

* Applied the rule via uBlock Origin "My filters" and verified its effect
* Checked multiple pages (homepage and internal pages)
* Performed A/B testing by toggling the filter on/off to confirm the rule’s impact
* Verified behavior in both normal and private browsing modes
* Confirmed consistent behavior in another browser

No site breakage observed.

## Suggestion

If similar patterns are observed across other PwC regional domains, this rule could potentially be extended or generalized to improve coverage.

## Before
스크린샷 2026-04-21 오후 3 53 56

## After
스크린샷 2026-04-21 오후 3 57 28

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_hide.txt
+++ b/easylist_cookie/easylist_cookie_specific_hide.txt
@@ -574,6 +574,7 @@ novaramedia.com###obligation-bar
 swappa.com###offcanvasCookiesConsent
 postgraduatesearch.com###ol
 estrid.com###onetrust-banner
+www.pwc.com###onetrust-consent-sdk
 openu.ac.il###openu_privacy_protection_policy_alert
 pollylingu.al###opt-form
 atlantic.money###opt-in


Woojiyeong:
@ryanbr This is my first time submitting a pull request, so I’m a bit unsure if everything was done correctly.
If there are any issues, please let me know and I’ll fix them.

PR #23994, opened 2026-04-21
A: euronews.com

Woojiyeong:
## Site
https://www.euronews.com/

## Issue
Didomi cookie consent popup is not hidden on the site.

## Root cause
The cookie consent is implemented using Didomi and injected as a site-specific container (`#didomi-host`) that is not covered by existing EasyList rules.

## Fix
Added a domain-specific cosmetic filter:
`euronews.com###didomi-host`

## Technical notes
Used a site-specific cosmetic filter to precisely target the Didomi container.
A generic rule was avoided to reduce the risk of false positives on other sites.

## Testing
- Verified on homepage and multiple pages
- Confirmed via A/B testing (on/off)
- Tested in normal and private browsing modes
- Checked in another browser
- No breakage observed

## Notes
This fix is intentionally scoped to euronews.com to avoid unintended side effects on other Didomi implementations.

## Before
스크린샷 2026-04-21 오후 3 58 08

## After
스크린샷 2026-04-21 오후 3 58 57

### DIFF
--- a/easylist_cookie/easylist_cookie_specific_hide.txt
+++ b/easylist_cookie/easylist_cookie_specific_hide.txt
@@ -351,6 +351,7 @@ discoveryeducation.com,readingplus.com###de-consent-banner
 universitysupplystore.com###demo-overlay
 filmon.com###design-switcher
 alliancefrancaise.ca###dialogbox_cookiepolicy
+euronews.com###didomi-host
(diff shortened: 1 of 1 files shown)


ryanbr:
This is fixed upstream in uBO cookie list.

PR #24020, opened 2026-04-22
Fix operationsports.com

Alex-302:
Common + domain specific fix.


zloyden (commented):
You can also add the domain to this rule.
```adblock
apnews.com,sidereel.com##+js(acs, document.getElementsByTagName, .pubads)
```

PR #24075, opened 2026-04-25
A: https://facebook.com - block login wall

flexagoon:
Blocks an undismissable Facebook login wall:

image

(One of the classes is named `__fb-light-mode`, but from my testing it seems like Facebook is always in light mode when logged out, so it should be fine without also adding `dark-mode`)


ryanbr:
Will this break if someone actually wants to login though?

flexagoon:
@ryanbr no, there is still a login bar at the top of the screen

image

ryanbr:
I'll leave it

PR #24108, opened 2026-04-26
Add MDN's bottom advertisement to easylist_specific_hide

regedit-sys:
I added a filter that removes the advertisement from the bottom of the MDN's homepage.

Without the added filter rule:
image

With the added filter rule:
image


ryanbr:
Fix commit.

PR #24128, opened 2026-04-27
A: Hide YouTube anti-adblock enforcement popup + cover youtube-nocookie ad_break

jimbo111:
The "Ad blockers violate YouTube's Terms of Service" popup currently isn't covered. uAssets handles it via scriptlets, but for engines that load EasyList without uAssets the popup shows.

`ytd-enforcement-message-view-model` mounts inside either `tp-yt-paper-dialog` or `ytd-popup-container` depending on the session. Hiding only the bare element leaves an empty wrapper covering the page, so all three are needed:

    ! YT anti-adblock enforcement
    www.youtube.com##tp-yt-paper-dialog:has(ytd-enforcement-message-view-model)
    www.youtube.com##ytd-enforcement-message-view-model
    www.youtube.com##ytd-popup-container:has(ytd-enforcement-message-view-model)

Also mirroring `||youtube.com/youtubei/v1/player/ad_break` for `youtube-nocookie.com`. Embedded players using the privacy-enhanced iframe currently slip through.

### DIFF
--- a/easylist/easylist_specific_block.txt
+++ b/easylist/easylist_specific_block.txt
@@ -766,6 +766,7 @@
 ||yimg.com/rq/darla/$domain=yahoo.com
 ||ynet.co.il/gpt/
 ||youswear.com/ads/
+||youtube-nocookie.com/youtubei/v1/player/ad_break
 ||youtube.com/pagead/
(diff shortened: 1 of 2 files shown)


Khrin:
This is not a good idea.

PR #24148, opened 2026-04-29
A: https://brusselsairlines.com

flexagoon:
image


ryanbr:
I suspect this will break other features. should be a trusted rule.

PR #24160, opened 2026-05-04
P: mothership.sg - consolidate imu ad blocks into wildcard rule

### LINKED ISSUE #24144: mothership.sg: ad leftovers
how does this keep evading my clutches

Image

`https://mothership.sg/2026/04/childcare-teacher-stabbed-special-needs-boy-pen-singapore/`

block-imu[NUMBER] seems to count up for every advertisement block, so we need to block it like 
`##[id ^= "block-imu"]:has-text("ADVERTISEMENT")`

yskim308:
This PR closes #24144. 

Ads on mothership.sg are blocked at the network level, but large gray placeholders remain. The current filters use fragile sequential IDs (block-imu1, block-imu2). block-imu3 (and > 3) are thus not hidden. 

Fix: Replaced sequential rules with a single wildcard:
`mothership.sg##section[id^="block-imu"]:has(> .advertisement__container)`

This should handle all current and future block-imu units automatically. The :has() selector ensures we only target ad-specific sections, preventing accidental over-blocking.

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -362,8 +362,6 @@ ancient-origins.net###block-block-49
 insidehighered.com###block-dfptagarticle-sidebar-bottom
 insidehighered.com###block-dfptagarticle-sidebar-top
 spine-health.com###block-dfptagbottomanchorads1
-mothership.sg###block-imu1
-mothership.sg###block-imu2
 kenyans.co.ke###block-keleaderboardmenu
 romania-insider.com###block-nodepagebelowfromourpartners
 romania-insider.com###block-nodepagebelowlatespress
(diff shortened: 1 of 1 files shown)


ryanbr:
Is there an example where the current rules aren't good enough? @yskim308

yskim308:
The example in #24144 links to 
https://mothership.sg/2026/04/childcare-teacher-stabbed-special-needs-boy-pen-singapore/

where at the bottom of the page the advertisement block is not hidden

image

This seems to be happening because current rules block imu-1 and imu-2 but some articles have more than 2 imu sections (imu-3, imu-4, etc) resulting in the css not being hidden.

lunaflight:
https://mothership.sg/2026/07/hougang-abc-cheating-summary/ this link uses `block-outstream9` and such too, instead of "imu"

Issue #20702, opened 2024-11-22
zeotap.com


aniketawati:
Filter affected:
zeotap.com in [easyprivacy_trackingservers_thirdparty.txt](https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_trackingservers_thirdparty.txt#L2462)

1st/3rd-party sites affected:
First party site affect - zeotap.com

How is it broken?
No content on the site is allowed to load

Description of why it should be removed:
This was already removed in #17488  But it got added back in this [commit] (https://github.com/easylist/easylist/commit/7cd9d9fdb75002438375fe64b780c6d65a67cdd7)

zeotap.com is corporate website. Adding the root domain in blocklist is stopping the first party website from loading.

aniketawati:
@ryanbr It would be great if you could review this. Thanks!.

Issue #17488, opened 2023-10-09
zeotap.com


aniket-zeotap:
### Filter affected:
zeotap.com in [easyprivacy_trackingservers.txt](https://github.com/easylist/easylist/blob/7e60e227526f1ebab2a9b79a1367651fb494884f/easyprivacy/easyprivacy_trackingservers.txt#L2932)

### 1st/3rd-party sites affected:
First party site affect - zeotap.com

 
### How is it broken?
No content on the site is allowed to load

### Description of why it should be removed:
zeotap.com is corporate website. The tracking domain for third party data collection is `spl.zeotap.com` So the blocklist should contain `spl.zeotap.com` instead of zeotap.com which blocks the first party website.

PR #24171, opened 2026-05-06
R: Remove ||zeotap.com^$third-party — Zeotap is a CDP, not a tracker

shubhamzeotap:
## Summary

Requesting removal of `||zeotap.com^$third-party` from `easyprivacy/easyprivacy_trackingservers_thirdparty.txt` and `zeotap.com` from `cleaned-domains.txt`.

**Zeotap is a Customer Data Platform (CDP), not a tracking company.** It is enterprise data infrastructure used by 80+ paying customers for first-party data unification, segmentation, and activation. It is not an ad tracker, pixel network, or fingerprinting service.

## What's blocked today

The `||zeotap.com^$third-party` rule blocks **all** `*.zeotap.com` subdomains in third-party context. This includes:

| Subdomain | Purpose | Legitimate? |
|---|---|---|
| `app.zeotap.com` | Primary SaaS application (customer login, dashboards, data management) | Yes — first-party product |
| `auth.zeotap.com` | SSO / OpenID Connect authentication | Yes — auth flows |
| `zeotap.com` | Corporate website | Yes |
| `spl.zeotap.com` | SDK / data collection endpoint | Debatable — this is the only subdomain that could be considered tracking-adjacent |

The rule was added in **August 2017** (commit `0ddad679`) based on a single observation of zeotap.com loading on `news18.com`. Since then, Zeotap has pivoted to a full enterprise CDP platform — it is no longer a third-party data provider or tracker.

## Impact on users

- **Enterprise customers** accessing `app.zeotap.com` through SSO redirect flows, bookmarks from internal portals, or embedded dashboards see broken pages when EasyPrivacy is enabled
- **DNS-level blocklists** (Pi-hole, NextDNS, AdGuard Home) that consume `cleaned-domains.txt` block `zeotap.com` entirely — including direct first-party access to the corporate website and product login

## Prior issues

- **#17488** (Oct 2023) — Zeotap employee reported the unconditional `||zeotap.com^` in `easylist_adservers.txt`. Removed by @ryanbr.
- **#20702** (Nov 2024) — Same unconditional rule re-appeared and was removed again by @ryanbr.

Neither issue addressed the `$third-party` variant in EasyPrivacy, which is what this PR fixes.

## Proposed alternative

If the concern is specifically about `spl.zeotap.com` (the data collection endpoint), we'd be open to a narrower rule:

```
||spl.zeotap.com^$third-party
```

This would block the SDK endpoint while allowing the SaaS application (`app.zeotap.com`), authentication (`auth.zeotap.com`), and corporate website to function normally.

## Changes

- Removed `||zeotap.com^$third-party` from `easyprivacy/easyprivacy_trackingservers_thirdparty.txt`
- Removed `zeotap.com` from `cleaned-domains.txt`

## About Zeotap

- **Website**: https://zeotap.com
- **Type**: Enterprise Customer Data Platform (CDP)
- **Customers**: 80+ enterprise brands
- **Use case**: First-party data unification, identity resolution, audience segmentation, activation
- **Not**: An ad tracker, pixel network, fingerprinting service, or third-party data broker

### DIFF
--- a/cleaned-domains.txt
+++ b/cleaned-domains.txt
@@ -828,7 +828,6 @@ ythd.org
 ytimg.com
 zdassets.com
 zencdn.net
-zeotap.com
 zi-scripts.com
 ziffstatic.com
 ziggogratis.shop
--- a/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
+++ b/easyprivacy/easyprivacy_trackingservers_thirdparty.txt
@@ -2529,7 +2529,6 @@
 ||zdtag.com^$third-party
 ||zedwhyex.com^$third-party
 ||zeerat.com^$third-party
-||zeotap.com^$third-party
 ||zesep.com^$third-party
 ||zeustechnology.com^$third-party
 ||zoomanalytics.co^$third-party


ryanbr:
Outline of tracking.

## Summary

The Zeotap integration on this property (the `ue-*` identity controller plus the
Zeotap mapper tag it feeds) meets the functional definition of a cross-site
tracker. It takes a logged-in user's account identity, converts it into a
persistent, deterministic identifier, and broadcasts that identifier to a
third-party identity-resolution network that syncs it with other ad-tech
vendors.

The position that "this is not a tracker" does not hold up against the code's
actual behavior. I'm requesting removal, or at minimum a documented
justification and a consent gate that demonstrably sits upstream of the
identity broadcast.

## What "not a tracker" would require

For this integration to *not* be a tracker, it would need to be true that the
data it emits is non-identifying, non-persistent, or not shared across sites.
None of those hold:

- **It is identifying.** The value sent is derived from the logged-in user's
  account (`hid` claim from the publisher JWT). It resolves to a specific
  person, not an anonymous session.
- **It is persistent.** The resolved Zeotap ID is stored as a first-party
  cookie with a ~365-day expiry, with localStorage fallback, and the storage
  layer deliberately walks up to the registrable domain so the ID survives
  third-party-cookie blocking.
- **It is shared across sites.** The mapper calls `idplus.zeotap.com` to
  resolve a universal ID and then `spl.zeotap.com/fp` to cookie-sync that ID
  with other vendors. Cross-site/cross-vendor linkage is the entire purpose of
  the endpoint, not a side effect.

A mechanism that produces a durable, person-level identifier and ships it to a
third-party graph is a tracker regardless of what it is labelled internally.

## Evidence from the code

### 1. Identity is extracted from login state

The publisher-side controller maps each hostname to its registration cookie and
reads it to determine login status:

| Property | Cookie |
|---|---|
| elmundo.es | `REGMUNDO` |
| marca.com | `REGMARCA` (+ regional variants) |
| telva.com | `REGTELVA` |
| expansion.com | `expansionID` |

On login it reads a JWT from `localStorage["uejwt"]`, base64url-decodes the
payload client-side (no signature verification), extracts the `hid` claim, and
calls:

```js
window.zeotap.setUserIdentities({ email: hid }, true);
```

The `hid` is the linkage key. If it is a stable hash of the account, the same
user is recognized across all four properties.

### 2. Identity is forwarded to a third-party resolver

The Zeotap mapper tag takes that identity, hashes declared identifiers
(SHA-256), and sends them to:

- `https://idplus.zeotap.com/api/v1/idp` — returns a universal `idpCookie`
- `https://spl.zeotap.com/fp` — cookie-sync pixel + event upload

All requests are made with `withCredentials = true` (credentialed cross-domain).

### 3. The result is stored to persist

- Universal ID written first-party, ~365-day expiry (localStorage fallback).
- Storage walks up to the registrable domain (`__tld_test__` probe) so the ID
  is first-party and resistant to third-party-cookie blocking.

### 4. The sync pixel fans out to other vendors

The `/fp` response is injected into a hidden `div#zeo_mapping` via `innerHTML`
with no sanitization. The injected markup is what performs the actual
cross-vendor sync (partner pixels/iframes). The downstream vendor set is
controlled by the server response and is therefore not visible in the static
code — it can only be enumerated from a live capture.

## Data flow

```
Publisher login (REG* cookie + uejwt JWT)
        │
        ▼
ue identity controller → reads JWT hid → window.zeotap.setUserIdentities({email: hid})
        │
        ▼
Zeotap mapper → SHA-256 declared IDs → idplus.zeotap.com → universal ID
        │
        ▼
ID stored first-party (~365d) → /fp pixel sync → downstream ad-tech vendors
```

## Specific concerns

1. **Consent ordering.** The mapper applies IAB TCF consent gating (vendor 301).
   But the publisher-side controller calls `setUserIdentities` on login state
   alone, with no consent check at that call site. This needs verification:
   does the consent gate provably sit *upstream* of the identity broadcast, or
   can a logged-in user who refused consent still be identified through this
   path?
2. **Permissive default.** In the mapper's consent resolution, when consent
   handling is disabled and opt-out is not set, `track`, `identify`,
   `cookieSync`, and `vendorConsent` all default to `true`. The fail-open state
   is "track."
3. **PII handling.** The mapper additionally accepts email/phone via URL query
   params before hashing — raw PII in URLs leaks via referrer headers and logs.
4. **Unsanitized `innerHTML` sink.** The `/fp` response is injected without
   sanitization; its safety rests entirely on trusting that endpoint.

## How to verify independently

Open DevTools → Network on a logged-in session and watch for calls to
`idplus.zeotap.com` and `spl.zeotap.com/fp`. Inspect the `/fp` response to see
the downstream pixels it injects. Check `localStorage` and cookies for the
persisted `IDP` value and its expiry.

PR #24267, opened 2026-05-21
M: www.google.com: Block Google Search AI Overview independent of language

dd-ast:
The current rule only blocks the AI Overview widget when the results are in English or German.
Since the rule targets the widget's ID, I removed the `:has-text(/AI Overview|Übersicht mit KI/)` check.  I can confirm that the rule now works independent of language (e.g. when the search results are in Romanian).

### DIFF
--- a/fanboy-addon/fanboy_ai_suggestions.txt
+++ b/fanboy-addon/fanboy_ai_suggestions.txt
@@ -113,7 +113,7 @@ mathrubhumi.com##.summary-wrapper
 techedt.com##.teqr-quick-read
 britannica.com##.toc-sticky-header-ai-container
 autonocion.com##.us-box-resumen
-www.google.com#?##Odp5De:has-text(/AI Overview|Übersicht mit KI/)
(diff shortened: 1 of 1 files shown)


Stevoisiak:
The `#?#` can be replaced with `##`. [From the documentation](https://github.com/gorhill/ublock/wiki/Procedural-cosmetic-filters#subjecthasarg):

> Using #?# instead of ## for a procedural cosmetic filter will prevent uBO from trying to convert the filter into a declarative (native) one.

Since this isn't using `has-text()` anymore, the filter can use native CSS.

Also, this may be out of scope for this PR, but AI overviews can appear in region specific domains as well. For example: `https://www.google.com.hk/search?q=Why+are+we+here` and `https://www.google.co.jp/search?q=why+are+we+here`. It may be worth using `www.google.*##` to cover all regions.

dd-ast:
Hi @ryanbr,
Would it be ok if I opened another PR that includes the changes suggested by @Stevoisiak?
Thanks!

PR #24297, opened 2026-05-28
A: mail.google.com

Sbrjt:
Hide "Get professional email" on mail.google.com

image


ryanbr:
Why 3 filters for one element?

Sbrjt:
Initially I wanted to do something like `##div[data*recommendation*]`, but I don't know if easylist syntax supports that. So, I split them into 3 selectors.

Now, I've updated to just one selector.

PR #24305, opened 2026-05-29
A: deliveroo.ie selfpromo

babfozelek:
deliveroo.ie Plus Gold selfpromo filter + scrollbar enable


ryanbr:
Squash the commit, and does it apply to other regional domains?

babfozelek:
@ryanbr,
Squash commits are done: https://github.com/easylist/easylist/pull/24352

No, only applies to Ireland domain.

ryanbr:
Squash commits into 1 commit, and why another PR?

babfozelek:
@ryanbr,
Done.

ryanbr:
Why do we need to enable the scroll? is it disabled by the first commit?

babfozelek:
@ryanbr,

Because the scroll will be hidden until the advertisement is closed.

Deliveroo shows the ad and disables the scroll.
After closed with "X", the scroll will be enabled again.

kép
kép

Issue #24309, opened 2026-05-31
mojeek.com


GreenBeard:
#### List the website(s) you're having issues with:

```txt
https://www.mojeek.com/search?q=chicken
```

#### What happens?

Has unblocked ads in the results.

#### Screenshot(s):

<img width="1236" height="741" alt="Image" src="https://github.com/user-attachments/assets/e243f6d3-b27f-432e-ada5-3ef1ec5b5db0" />

#### Filter lists you're using:

Easylist

#### Your settings:

- OS/version: Linux
- Browser/version: Firefox
- Ad block extension/version: Ublock Origin

#### Other details:

N/A

PR #24310, opened 2026-05-31
M: Fixes https://github.com/easylist/easylist/issues/24309

GreenBeard:


GreenBeard:
I guess mojeek.fr and mojeek.de could also be added which seem to need the same `aria-label="Sponsored products"` match (doesn't seem to be domain or language dependent from my testing). Those two domains are the only other domains listed at https://blog.mojeek.com/2021/10/mojeek-preference-settings-language-and-location.html and from what I see in the language preferences. I'm not sure if that should be done in a separate PR since the other existing rules also seem to be mojeek.com only at the moment and I'm not very experienced at contributing to easylist.

PR #24359, opened 2026-06-05
A: Update revolving adservers

smed79:


ryanbr:
How current are these? These are updated often..

Yuki2718:
Just pick some of them and you'll see many of them are current e.g. `abackswindleconcurrence.com`, `acceptancesuicidegel.com`, etc. We have to merge PRs like this as quickly as possible, but I was too busy last week.
____
I can't merge any longer, as the branch was deleted. I'll try to include living domains in my next commit.

PR #24375, opened 2026-06-07
Add return type annotation to start() function

Lang-Qiu:
### Background
The public `start()` function in `FOP-nocommit.py` lacked a return type annotation. Adding type annotations improves code readability and facilitates static type checking, aligning with modern Python best practices without altering functionality.

### Changes
- Added `-> None` return type annotation to the `start()` function definition to explicitly indicate that it returns no value.

### Verification
- The change is minimal and only affects type hints; the function's behavior remains unchanged.
- Can be verified by running the code normally or using a type checker like `mypy` to ensure no syntax or type errors are introduced.
- No external dependencies are added, maintaining compatibility.


ryanbr:
Note, use FOP rs. FOP.py is old/dead/redundant.

PR #24451, opened 2026-06-18
A: Added cky-overlay class

balazsnasz:
Added .cky-overlay blocking. This fixes overlay stuck on all websites using CookieYes.

CookieYes is one of the most popular cookie consent plugins for WordPress.
There is a general block that already blocks the consent popup showing, but the overlay stays. It makes websites using CookieYes unusable.

Adding ##.cky-overlay fixes this problem.


ryanbr:
This is bound to create issues, not comfortable with this

balazsnasz:
@ryanbr Without this rule, lots of websites using CookieYes don't work properly. The overlay remains visible, making the site unusable.
There are already several CookieYes-specific entries in easylist_cookie_specific_hide.txt. If this issue affects all websites using CookieYes, why not handle it with a general rule instead of adding site-specific exceptions?

Issue #15960, opened 2023-05-18
||non.li^ blocks l.francebleu.fr, l.bfmtv.com and a bunch of other domains used on social media


moisemorard:
### List the website(s) you're having issues:

`https://www.facebook.com`
`https://www.twitter.com`
`https://www.linkedin.com`
`https://www.instagram.com` (Link in bio)

### What happens?

Subdomains delegated to us by our clients via CNAME are being blocked by EasyList / Adguard (DNS Filter) when users click on our clients' social media posts shared on platforms such as Twitter, Facebook, Instagram, and LinkedIn, despite our commitment to user privacy and assurance that our domains are not used for track individual users, displaying ads or retargeting users.

### List Subscriptions you're using:

- Adguard (DNS Filter)
- It seems to be this [PR](https://github.com/easylist/easylist/pull/15312) managed by @Yuki2718 @0o5rlubm @ryanbr 

### Your settings

- OS/version: Mac OS, Android
- Browser/version: Up to date
- Adblock Extension/version: Adguard 2.10.0.1272

### Other details:

We follow strict guidelines to ensure appropriate anonymization of user data. Our approach includes rounding up click-through rates on our graphs to the nearest ten, with a granularity lower than ten minutes, ensuring individual users cannot be tracked or identified. We are also compliant with [CNIL](https://www.cnil.fr/fr/cookies-solutions-pour-les-outils-de-mesure-daudience) guidelines and have received an exemption from consent requirements.

We kindly request that EasyList considers removing the restriction on the CNAME non.li and whitelisting our subdomains. Additionally, if there is any specific reason for maintaining the current block, we would appreciate your guidance on the necessary adaptations we can make on our end to ensure compliance with your requirements.

Examples of affected social media posts shared by our clients:

- `https://www.facebook.com/36550584062_10161795094749063`
- `https://www.facebook.com/153632746935_10160510007091936`
- `https://www.twitter.com/BFMTV/status/1659127779038773248`
- `https://twitter.com/34867057/status/1659129228174979074`

Users can disable click measurements by visiting the following links:

- `https://l.francebleu.fr/cookie/consent`
- `https://l.leparisien.fr/cookie/consent`
- `https://l.bfmtv.com/cookie/consent`

Best regards,

Moïse
CEO, Nonli

moisemorard:
> As far as AdGuard Tracking Protection is concerned, I didn't add `||non.li^`.

[...]

PR #24503, opened 2026-06-25
Remove non.li third-party filter from EasyPrivacy

moisemorard:
## Summary

Remove the broad `||non.li^$third-party` rule from EasyPrivacy.

## Context

This follows the same false-positive cleanup that was accepted in Liste FR:

- PR: https://github.com/easylist/listefr/pull/294
- Maintainer commit: https://github.com/easylist/listefr/commit/4389c23e83fe2bc628c4e11af40f9e6264338e06

The EasyPrivacy rule appears to come from the 2023 BFMTV report where `l.bfmtv.com` was observed as an alias for `non.li`:

- Original PR: https://github.com/easylist/easylist/pull/15312
- Follow-up issue: https://github.com/easylist/easylist/issues/15960

The current EasyPrivacy rule should be removed because the current Nonli SDK behavior does not match this domain-level tracker classification.

The SDK / analytics tag behavior is documented here:

https://www.nonli.com/en/faq/nonli-sdk-analytics-tag-q-a

That document explains the current production behavior:

- the analytics tag is served from a subdomain of the publisher client's main domain;
- traffic analysis happens on Nonli servers anonymously;
- the tag drops no cookie in the visitor's browser and reads no cookie either;
- no individual visitor journey is reconstructed;
- collected data is anonymized and never consolidated across different domains;
- each brand gets its own isolated editorial measurement;
- the SDK is focused on article traffic analysis and social-publishing decisions, not advertising targeting or user retargeting.

This means the `non.li` domain-level EasyPrivacy rule no longer has a technical reason to exist for the current Nonli service.

The broad rule also creates false positives outside the SDK context. `non.li` is used for legitimate publisher shortlink / redirect infrastructure, so blocking or classifying the whole domain causes downstream products that consume EasyPrivacy-derived domain rules to flag normal publisher links before they even reach the website.

Examples of affected legitimate uses include:

- publisher shortlink / redirect domains such as `l.numerama.com`, which resolve to Nonli redirect infrastructure;
- direct Nonli short links used to redirect readers to publisher content;
- publisher infrastructure where the short URL identifies a public content item, not a person.

The shortener behavior is documented here:

https://www.nonli.com/en/shortener-data-protection

That document explains the current production behavior of Nonli short links:

- a short-link redirect is a plain HTTP redirect;
- the response sets no cookie and serves no script, pixel, or executable content;
- no local storage, fingerprinting, visitor identifier, recipient identifier, session identifier, or user identifier is used;
- the shortener has set no cookie at all since January 22, 2025, so older observations no longer reflect the current service;
- click data is limited to aggregated editorial counters on behalf of the publisher;
- no advertising targeting, profiling, personalized advertising, or individual profile building is performed by Nonli or on its behalf.

Because of that, this EasyPrivacy rule is not an accurate representation of the current service. It also breaks products that flatten ABP-style rules into DNS/app-level tracker lists, where `$third-party` and path context may no longer protect normal publisher SDK or shortlink usage.

With the current production behavior, we do not see a technical reason for keeping `non.li` in EasyPrivacy.

## Verification

- Removed only `||non.li^$third-party` from `easyprivacy/easyprivacy_trackingservers_international.txt`
- Ran `git diff --check`
- Verified there are no remaining `non.li` rules under `easyprivacy/`

### DIFF
--- a/easyprivacy/easyprivacy_trackingservers_international.txt
+++ b/easyprivacy/easyprivacy_trackingservers_international.txt
@@ -308,7 +308,6 @@
 ||mmtro.com^$third-party
 ||netquattro.com/stats/
 ||netvigie.com^$third-party
-||non.li^$third-party
 ||pa-cd.com^$third-party
 ||phywi.org^$third-party
 ||pingclock.net^$third-party


ryanbr:
You're not showing any proof of actual breakage,  we don't trust you. If you can't actually follow through with any examples, it'll be closed and ignored.

moisemorard:
@ryanbr Thanks for the feedback. Here is a concrete breakage example.

A normal user opening `https://non.li/L0w` in Chrome with NordVPN Threat Protection enabled is stopped before reaching the link destination. NordVPN shows a full-page warning saying:

> This site will track your activity
> If you continue, trackers will run.

Screenshot, redacted to remove browser bookmarks/extensions:

NordVPN warning for non.li shortlink

This is not an ad-slot or cosmetic issue. It blocks normal navigation to a publisher shortlink and makes the user think the link itself is unsafe/tracking-heavy.

I understand EasyPrivacy is an ABP-style list and this rule is `$third-party`. The actual breakage we are seeing comes from downstream products that consume tracker classification/filter data in contexts where the browser-extension `$third-party` semantics no longer protect normal first-party navigation. The exact vendor pipeline is opaque, but EasyPrivacy still publicly contains the broad `||non.li^$third-party` rule, while Liste FR and AdGuard public filters no longer contain `non.li`.

Current public checks:

- `easyprivacy.txt` version `202606301623` still contains `||non.li^$third-party`
- the EasyPrivacy source still contains it in `easyprivacy/easyprivacy_trackingservers_international.txt`
- Liste FR no longer contains `non.li` after https://github.com/easylist/listefr/pull/294
- AdGuard public DNS / Tracking / French filters no longer contain `non.li`
- DuckDuckGo Tracker Radar still contains `non.li`; we opened https://github.com/duckduckgo/tracker-radar/issues/388

On the technical side, the current Nonli SDK does not match a tracker classification:

- no cookie is dropped
- no cookie is read
- no individual visitor journey is reconstructed
- data is anonymized and not consolidated across domains
- no advertising targeting, profiling, retargeting, or personalized advertising

SDK documentation:
https://www.nonli.com/en/faq/nonli-sdk-analytics-tag-q-a

Also, for the shortener path specifically, Nonli has not set any cookie at all since January 22, 2025:
https://www.nonli.com/en/shortener-data-protection

Could you please reopen and reconsider the removal?

PR #24546, opened 2026-06-29
A: dailykos.com

avanikaein:
bottom sticky banner ad appears after the soft Admiral wall is dismissed, takes a few seconds to load.

Tested on Chrome with ABP, using EasyList & ABP Filters

appears on homepage or subpage, for example `https://www.dailykos.com/stories/2026/6/29/800062732/community/pwb-peeps-emergency-diary/`

dailykos


ryanbr:
Subscribe to Easyprivacy.

PR #24649, opened 2026-07-13
A: https://dl.acm.org/doi/10.1145/3485447.3512218

smed79:


Yuki2718:
~~PubMetrics is not an analytics service, it's a tool for clinical researchers to help meta-analysis.~~ Whoops, it looks the script is unrelated to PubMetric, I missed the last `s`. However, it's still not an analytics, just a helper script for paper metrics that are important to researchers.

PR #24724, opened 2026-07-23
discord quests and promoted games

### LINKED ISSUE #24657: discord quests & promoted games
i've maintained my own personal filter lists for a few years now, which have primarily focused on discord quests & promoted games, for both the web and desktop clients: https://github.com/DetachHead/filter-lists/blob/466e87b03cdd0a4159857bbc349d227eca509874/list.txt#L79-L103

but now i'm looking to upstream as many of these as i can into a widely used filter list so that more users can benefit from them, and so my repo can be repurposed to focus on the [css-based filters for the desktop client instead](https://github.com/DetachHead/filter-lists/blob/master/discord.css)

i'm happy to make a PR if this is something that's considered in-scope for easylist. my definition of ads include:

- quests (ads for specific games that give "rewards")
- orbs (can only be obtained by clicking on ads)
- promoted activities

it does not block nitro promotions/upsells, etc. as they are considered self promotion.

DetachHead:
fixes #24657 

this is a cut down version of what i had in my own list, with only the rules i could find current examples of.

some rules are duplicated with slightly different class names because for some users the format of the class names is different.

i'll add a thread for each rule with a screenshot

### DIFF
--- a/easylist/easylist_specific_hide.txt
+++ b/easylist/easylist_specific_hide.txt
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
 checkfelix.com,kayak.*,swoodoo.com##div[class*="-adInner"]
 checkfelix.com,kayak.*,swoodoo.com##div[data-resultid]:has(a.IZSg-adlink)
 checkfelix.com,kayak.*,swoodoo.com##div[id^="inline-"]
+! discord
+! quests tab in DMs:
+discord.com##nav[class$='-privateChannels'] li:has(a[href='/quest-home'])
+discord.com##nav[class^='privateChannels_'] li:has(a[href='/quest-home'])
+! quests tab in Discover section:
+discord.com##nav>div:has-text(/^Quests$/)
+! quests on user profiles:
+discord.com##div[id^='popout_']:has(h2:has-text(/^Quest Available!/))
+! promoted games in Discover section:
+discord.com##section[class$='-contentSection'] div:has(>h3:has-text(/^Promoted Games$/))
+discord.com##section[class^='contentSection_'] div:has(>h3:has-text(/^Promoted Games$/))
+! promoted activities:
+discord.com##div[class$='-drawerSizingWrapper'] div:has(>div:has-text(/^Promoted$/))
+discord.com##div[class^='drawerSizingWrapper_'] div:has(>div:has-text(/^Promoted$/))


easylist/easylist_specific_hide.txt
```diff
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
 checkfelix.com,kayak.*,swoodoo.com##div[class*="-adInner"]
 checkfelix.com,kayak.*,swoodoo.com##div[data-resultid]:has(a.IZSg-adlink)
 checkfelix.com,kayak.*,swoodoo.com##div[id^="inline-"]
+! discord
+! quests tab in DMs:
+discord.com##nav[class$='-privateChannels'] li:has(a[href='/quest-home'])
+discord.com##nav[class^='privateChannels_'] li:has(a[href='/quest-home'])
```
DetachHead, line 8732: Image

easylist/easylist_specific_hide.txt
```diff
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
...
+! quests tab in Discover section:
+discord.com##nav>div:has-text(/^Quests$/)
```
DetachHead, line 8734: Image

easylist/easylist_specific_hide.txt
DetachHead: i can imagine this might stretch the definition of an ad, but my reasoning is that this tab literally only contains ads, so i figured it makes more sense to just hide the quests page entirely instead of keeping it and blocking everything on the page:

Image

though i'm happy to do that instead if you prefer

easylist/easylist_specific_hide.txt
```diff
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
...
+! quests on user profiles:
+discord.com##div[id^='popout_']:has(h2:has-text(/^Quest Available!/))
```
DetachHead, line 8736: Image

easylist/easylist_specific_hide.txt
```diff
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
...
+! promoted games in Discover section:
+discord.com##section[class$='-contentSection'] div:has(>h3:has-text(/^Promoted Games$/))
+discord.com##section[class^='contentSection_'] div:has(>h3:has-text(/^Promoted Games$/))
```
DetachHead, line 8739: Image

easylist/easylist_specific_hide.txt
```diff
@@ -8726,3 +8726,17 @@ checkfelix.com,kayak.*,swoodoo.com##div[class*="-ad-card"]
...
+! promoted activities:
+discord.com##div[class$='-drawerSizingWrapper'] div:has(>div:has-text(/^Promoted$/))
+discord.com##div[class^='drawerSizingWrapper_'] div:has(>div:has-text(/^Promoted$/))
```
DetachHead, line 8742: Image

Stevoisiak:
This might fit better in [the annoyance list](https://github.com/easylist/easylist/blob/master/fanboy-addon/fanboy_annoyance_specific_hide.txt) as opposed to the main Easylist.

I imagine there are people who use Quests that would be confused to see the menu option disappear suddenly without warning. Especially since the main Easylist is enabled by default in many adblockers, while the annoyance list is opt-in.

Khrin:
@DetachHead Sorry, but it doesn't look like an advertisement.

DetachHead:
What about the promoted games? Those are pretty obvious advertisements

PR #24735, opened 2026-07-26
A: pinterest.com

zphrs:
Blocks the "more like this" footer and the in-grid "related searches" and "related interests" boxes. 

Hidden elements:
image

image

image


zphrs:
Moved to #24736 because I wanted to rename the branch

PR #24765, opened 2026-07-27
A: androidworld.nl

griffinallen:
Hiding cookie notice on https://androidworld.nl/nieuws/google-pixel-7-kopen-nederland

Fix is in response to user reports on Brave


griffinallen:
Good catch, thanks for closing

Issue #24925, opened 2026-08-16
top.gg


smed79:
self-promo (annoyance)

- https://i.ibb.co/N2xRdXr3/i.png

```adblock
top.gg##.h-\[250px\]
```

A: `https://top.gg/`

PR #25037, opened 2026-08-24
A: (#24925) https://top.gg/

smed79:


velddev:
This would cause a lot of breakage since you're using a generic width property - can you use the ID attached to the ad instead of breaking my website. 🥲

PR #25141, opened 2026-08-26
A: wowscale.com

josh-collins1:
`wowscale.com` and subdomains collect overzealous browsing/device information used for cross-site tracking - most likely for marketing/ads


josh-collins1:
@ryanbr

Khrin:
On which site this tracker is used? We need examples to see where it's used.

PR #25170, opened 2026-08-27
A: mashable.com: Social footer

Stevoisiak:


ryanbr:
Ugly, and probably break over time. I'll leave

Issue #24473, opened 2026-06-23
youtube.com - custom list anti-autodub audio


ghost:
https://github.com/easylist/easylist/blob/master/custom-lists/youtube-autodubbed.txt

Stevoisiak:
Auto-dubbed audio can play as the default on a video even if the options for them are hidden. There's [a uBlock Origin filter to prevent this](https://www.reddit.com/r/uBlockOrigin/comments/1npyxsc/comment/ng4kmje).

```
youtube.com##+js(json-edit, ..adaptiveFormats.*[?.audioTrack.isAutoDubbed])
youtube.com##+js(json-edit-xhr-response, ..adaptiveFormats.*[?.audioTrack.isAutoDubbed], propsToMatch, /youtubei)
youtube.com##+js(json-edit-fetch-response, ..adaptiveFormats.*[?.audioTrack.isAutoDubbed], propsToMatch, /youtubei)
youtube.com##+js(rpnt, script, (function serverContract(), if("object"===typeof window.ytInitialPlayerResponse){window.ytInitialPlayerResponse=JSON.parse(JSON.stringify(window.ytInitialPlayerResponse));}(function serverContract(), sedCount, 1)
```

The rpnt filter makes sure the filter works when loading a video directly by URL. Can result in a brief visible "refresh" on the video.

ghost:
That is true, but it takes a while to load unlike the instant working https://chromewebstore.google.com/detail/youtube-anti-translate/ndpmhjnlfkgfalaieeneneenijondgag extension.  The filter also does not work with mv3 in my experience like with Adguard MV3 on chrome: https://greasyfork.org/en/scripts/555043-anti-youtube-auto-dub#:~:text=Notice%3A,Works%20best%20with%3A,

Stevoisiak:
A different filter is needed for AdGuard.

```
youtube.com#%#//scriptlet('json-prune', '$..adaptiveFormats.*[?.audioTrack.isAutoDubbed]')
```

ryanbr:
Sure, we can test this.

ghost:
@Stevoisiak I added your adguard filter. It does have the same effect of needing to wait longer before watchign the video. But in Adguard MV2 in Firefox it does not switch to the original audio.

Is it possible to have a filter that loads instantly like the extension does or are the filters limited. If so how?

Stevoisiak:
You’d have to ask the person who wrote the filters originally

griffinallen:
[...]

PR #25288, opened 2026-09-03
M: Re-add YouTube autodub JSON filters

Stevoisiak:
Re-adds JSON filters for YouTube autodubbing. This prevents autodubbed audio tracks from playing as the default option on a video and prevents the "Audio track" menu from appearing in the video settings if all alternate tracks are autodubbed. These filters were originally created by  Originally created by paintboth1234 (https://www.reddit.com/r/uBlockOrigin/comments/1npyxsc/comment/ng4kmje/)

This filter was previously removed due to a now-resolved issue in Brave https://github.com/brave/brave-core/pull/37658 that caused ads appear on YouTube when `json-edit` was used.

Addresses #24473

### DIFF
--- a/custom-lists/youtube-autodubbed.txt
+++ b/custom-lists/youtube-autodubbed.txt
@@ -7,3 +7,8 @@
 !
 youtube.com#?#yt-lockup-view-model.ytd-item-section-renderer:has-text(Auto-dubbed)
 youtube.com#?#ytd-rich-item-renderer.ytd-rich-grid-renderer.style-scope:has-text(Auto-dubbed)
+! https://github.com/easylist/easylist/issues/24473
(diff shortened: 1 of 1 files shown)


ryanbr:
Will the break again? how safe is it

Stevoisiak:
As mentioned in the PR, the previous breakage was due to a bug in Brave. I've had the json-edit filters in [my AI blocklist](https://github.com/Stevoisiak/Stevos-AI-Blocklist) for the past 5 months and the Brave breakage was the only issue I ran into.

I'm admittedly less certain about the `rpnt` filter. Without it, the `json-edit` filters don't work when a YouTube video is loaded directly via URL. (As opposed to clicking on a video link while already on YouTube). My understanding is it works by reloading the page when loaded directly via URL. It may be safer to leave that one out.

PR #25296, opened 2026-09-04
A: open.online

griffinallen:
Hiding cookie notice on https://www.open.online/2026/06/30/bastoni-indagato-prostituzione-minorile-agenzia-escort-cinisello

Fix is in response to user reports on Brave


ryanbr:
Not needed

PR #25443, opened 2026-09-13
A: shellshock.io

VillainsRule:
part of the filters discussed in https://github.com/uBlockOrigin/uAssets/issues/34422

the hostname matching portion was committed in other filters to uAssets; the site has ~80 domains as seen [here](https://shellshock.io/unblocked/json/proxy-list.json)

<details><summary>the ads</summary>
image

image
</details>


Khrin:
Unable to find these elements, and in any case, from the names all seem to be related to self-promotion, which we don't block in EasyList.

PR #25446, opened 2026-09-13
A: drive.google.com

Sbrjt:
Hide "Get professional email and tools at 50% off"


Sbrjt:
Is it possible to do something like:

```
google.com##*:matches-attr(/^data.*recommendation.*/)
```

Issue #25458, opened 2026-09-14
bug(workflow): python-adb derived flrender incorrectly handles !#if directives


YagUber:
I wanted to get some easyprivacy filters integrated for a popular website, to assure the best effort was taken to preserve user privacy this would require making use of the [!#!if](https://github.com/gorhill/ublock/wiki/static-filter-syntax#if-condition) directive to handles differences across blocking extensions. This conditional directive is used in a few easylist filters already, but my specific use case was for [easyprivacy_specific_uBO.txt](https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_specific_uBO.txt) and [easyprivacy_specific_abp.txt](https://github.com/easylist/easylist/blob/master/easyprivacy/easyprivacy_specific_abp.txt).

As it turns out, the production [easyprivacy.txt](https://easylist.to/easylist/easyprivacy.txt) turns all directives into properly formatted comments unconditionally (`! #if`), leading to any use being effectively inert if consumed upstream without post-processing.

This issue is from [flrender](https://github.com/adblockplus/python-abp/blob/4b0aec2c91f97446e6f20c908a28ab449cb626cc/abp/filters/parser.py#L349-L355) by abp, which is used in the [build workflow](https://github.com/easylist/easylist/blob/aae8c5c7fd4e3d3bf8ddc9dad1e8173446c5c840/.github/workflows/build.yml#L76) in this repo.

While uAssets just [clones](https://github.com/uBlockOrigin/uAssets/blob/a79d1cb07cbf3286d8b869fb096cc8c196147246/tools/make-easylist.sh#L8) this entire repo and only removes [empty](https://github.com/uBlockOrigin/uAssets/blob/a79d1cb07cbf3286d8b869fb096cc8c196147246/tools/make-easylist.mjs#L130-L137) comments. AdGuard falls victim to the mangling in its [production](https://filters.adtidy.org/extension/chromium/filters/118.txt) sets since they include the raw easylist.to filter as their [source](https://github.com/AdguardTeam/FiltersRegistry/blob/90b4d400acc959e7520e85b65e483abf6cfd726e/filters/ThirdParty/filter_118_EasyPrivacy/template.txt#L2).

Ideally I can (and will) make a PR on python-abp, but I am not sure if anyone looks at that repo anymore. If that ever goes through hopefully the pinned commit can be bumped in the workflow.

<hr>

Example:
```py
from abp.filters.parser import parse_line

[...]

Issue #19841, opened 2024-08-19
Recent change in #19825 breaks ChatGPT


leebyron:
I'd like to ask for the change in #19825 to be reverted, which breaks the ChatGPT product.

The blocked URLs are used to serve feature gating rules which enables features for all users as they are gradually released. Anyone using a blocker based on easylist will now find features within ChatGPT they previously relied on have now disappeared.

For any ChatGPT user who reports this kind of problem, I do not want to advocate disabling those blockers or whitelisting any URLs which is insecure and error prone.

Disclaimer, I work on ChatGPT. If there are better ways to structure these URLs to allow for more clear application of user intent of privacy, I'd be happy to hear them, but I'd also ask for this change to be reverted in the meantime since its impact is too coarse and has resulted in many reports of broken behavior.

Yuki2718:
Adjuste the rules to allow feature gating. The report is valid, you send telemetry by them and EP blocks telemetry.

lee-openai:
Much appreciated! I'll test once this has propagated, but the change makes sense and as far as I understand should resolve the broken product concern

leebyron:
Much appreciated! I'll test once this has propagated, but the change makes sense and as far as I understand should resolve the broken product concern

Issue #19825, opened 2024-08-16
https://chatgpt.com/ces/v1/t and https://ab.chatgpt.com/v1/rgstr


ayebrian:
#### List the website(s) you're having issues with:

`https://chatgpt.com/ces/v1/t`
`https://ab.chatgpt.com/v1/rgstr`

#### What happens?

These paths are used for sending telemetry that logs everything from button clicks to the creation of new chats. This includes collecting data such as userID and other usage-related information.

#### Screenshot(s) 

![image](https://github.com/user-attachments/assets/665094ab-71c0-4caf-b954-892734e819d7)

![image](https://github.com/user-attachments/assets/d53f6a57-bca3-4c16-ba0a-77a3d21eb7b1)

#### Filter lists you're using:

EasyList, EasyPrivacy

#### Your settings:

- OS/version: Windows 10 22H2 19045
- Browser/version: Chrome 127
- Ad block extension/version:  uBlock Origin 1.59.0

#### Other details:
`||chatgpt.com/ces/v1/t`
`||ab.chatgpt.com/v1/rgstr`

PR #25459, opened 2026-09-14
A: chatgpt.com

YagUber:
Partial chatgpt.com tracking protections based on my own investigations. Remaining endpoints pending #25458 since they need synthetic responses from `no-fetch-if` to not endlessly retry. 769b89b4dd10284a007adcd0c497f9a2232ae544 was the removal of existing rules for what I believe was the same retry issue I came across. Additionally, existing rules in other lists (AdGuard...) throw CSP rejections in Firefox before their `noopjson` embed ever even hits JS, leading again to the aforementioned issue.

Original CES/Statsig coverage added from  #19825, OAI staff requested A/B blocking rules to be narrowed in #19841.

Funny enough, partially added again today in 3a719226875a7502d5921ac3708ce6d6d1c9a8cf.
This PR changes that to an explicit rule for their Datadog RUM proxy endpoint, explicitly blocking both the fetch and flush on exit with `sendBeacon`, additionally blocks their other proxy at `/awe/api/v2/rum`.

The additional rules included are pretty simple and their originating callers fail gracefully as well.

Statsig JS client SDK [exception proxy](https://github.com/statsig-io/js-client-monorepo/blob/b1d30176d9143b72fb3c338a509ae0d5b1a6aa82/packages/client-core/src/StatsigOptionsCommon.ts#L69-L74) endpoint.
```adblock
||api.oaistatsig.com/v1/sdk_exception^$xmlhttprequest,domain=chatgpt.com
```

When downloading codex from `/codex`, only fired if user is logged in. Endpoint 404's but the request is still made.
```adblock
||chatgpt.com/backend-api/codex/app-download-signal^$xmlhttprequest
```

The rest of the included `/unauth-mweb/` endpoints are hit while either unauthenticated or on mobile routes such as `m.chatgpt.com`.


ryanbr:
I have my doubts here, do we really need all of these, expanding to many more rules for one site...

YagUber:
Haha yeah, we could almost certainly consolidate the `unauth-mweb/events/` rules into a single broader rule with no issue.

I kept them separate because I only wanted to block the endpoints I had verified as telemetry, rather than something that might break on such a popular website.

Happy to consolidate them if you prefer the broader rule.

ryanbr:
`||chatgpt.com/unauth-mweb/events/$xmlhttprequest`

YagUber:
Consolidated as requested.

I retained `$ping` because `client-fatal-error` and `statsc/flush` use `navigator.sendBeacon()` with a `fetch()` fallback, so both `$ping` and `$xmlhttprequest` are needed.

